PR #8, opened 2025-12-21
Fix Optional<T> as field

### LINKED ISSUE #5: lots of Optional<T> in instance classes and possible hot execution paths
i assume thats part of a bit too naive port. 

```
 public final class Table implements StatefulWidget<TableState> {

    private final List<Row> rows;
    private final List<Constraint> widths;
    private final Optional<Row> header;
    private final Optional<Row> footer;
    private final Optional<Block> block;
    private final Style style;
```

similar some methods returns Optional which I could imagine being quite inefficient
in highperf updates:

```
     public Optional<GridPoint> getPoint(double x, double y) {
```

should probably review these and move to nullable fields and make  use of jspecify where helpful

melix:
The `Optional<T>` are preserved on return types though.

Fixes #5

### DIFF
--- a/demos/table-demo/src/main/java/ink/glimt/demo/TableDemo.java
+++ b/demos/table-demo/src/main/java/ink/glimt/demo/TableDemo.java
@@ -28,6 +28,7 @@ import org.jline.terminal.Terminal.Signal;
 import org.jline.utils.NonBlockingReader;
 
 import java.io.IOException;
+import java.util.ArrayList;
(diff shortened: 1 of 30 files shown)


maxandersen:
This is definitely better - we'll have to see how it behaves since if you have to create an optional for every cells foreground/background when something animates/move this could be expensive. but again - might not be so lets merge this and leave it for future perf issues that might never happen ;)

PR #9, opened 2025-12-21
Decouple JLine

### LINKED ISSUE #6: jline terminal leaky abstraction
from demos:
```
Terminal<> terminal = new Terminal<>(backend);

            // Handle resize
            backend.jlineTerminal().handle(Signal.WINCH, signal -> {
                try {
                    terminal.draw(this::ui);
                } catch (IOException e) {
                    // Ignore
                }
            });

            NonBlockingReader reader = backend.jlineTerminal().reader();
```

signalHandler and NonBlockingReader are JLine specific types which if all demos does this - then everything will be coupled to jline terminals which is unfortunate given jline3 and jline4 are both relevants (and possible other)

similarly toolkit.TUIRunner have a hard hardcoded JLineTerminal dependency because of this.

melix:
- Add onResize(), read(), peek() methods to Backend interface
  - Create BackendProvider SPI and BackendFactory with ServiceLoader discovery
  - Update all demos to use BackendFactory instead of JLineBackend
  - Change glimt-jline to runtimeOnly dependency

This allows swapping terminal backends transparently since the backend will be loaded via service loading.

Demos have been updated accordingly

Fixes #6

### DIFF
--- a/glimt-jline/src/main/java/ink/glimt/backend/jline/JLineBackend.java
+++ b/glimt-jline/src/main/java/ink/glimt/backend/jline/JLineBackend.java
@@ -15,8 +15,10 @@ import ink.glimt.style.Style;
 import ink.glimt.terminal.Backend;
 import org.jline.terminal.Attributes;
 import org.jline.terminal.Terminal;
+import org.jline.terminal.Terminal.Signal;
 import org.jline.terminal.TerminalBuilder;
 import org.jline.utils.InfoCmp;
+import org.jline.utils.NonBlockingReader;
 
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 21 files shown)


maxandersen:
nice - definitely solves the specific problem of coupling.

I actually found the same coupling interesting enough exists in ratatui - low level code uses crossterm directly. Which i found interesting - but yeah; still think it should be something at least possible to handle without direct coupling.

Not sure if it should all be flattened on the backend interface directly or should be more an event queue - i.e.  onResize() is an event, and the peek/read are used to figure out what keys are pressed. I'll try it out a bit.

glimt-jline/src/main/java/ink/glimt/backend/jline/JLineBackend.java
```diff
@@ -42,6 +45,7 @@ public JLineBackend() throws IOException {
             .jansi(true)
             .build();
         this.writer = terminal.writer();
+        this.reader = terminal.reader();
```
maxandersen, line 48: I think *always* opening this is not good.

It can have OS specific funny things happen/leak and you don't use/need it for all uses of terminal.

melix:
Ok I'll let you play with this, I only did the minimal thing to decouple.

PR #12, opened 2025-12-22
Fix demos not runnable from directory

### LINKED ISSUE #11: demo build issue
not sure if this is expected or me being dumb.

should it not be possible to build a demo project by just going into the project?

```
 ../../gradlew build

FAILURE: Build failed with an exception.

* What went wrong:
Project directory '/Users/manderse/code/glimtink/glimt/demos/file-browser-demo' is not part of the build defined by settings file '/Users/manderse/code/glimtink/glimt/settings.gradle.kts'. If this is an unrelated build, it must have its own settings file.
```

This is the same error I get on any attempt running/building them from IDE's so i think we should find a way to fix this
or consider a different demo structure on disk.

cc @melix

melix:
Fixes #11

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -9,7 +9,7 @@ include(
     "glimt-toolkit"
 )
 
-File("demos").listFiles()?.forEach {
+File(settingsDir, "demos").listFiles()?.forEach {
     if (it.isDirectory) {
         include("demos:${it.name}")
     }


settings.gradle.kts
```diff
@@ -9,7 +9,7 @@ include(
     "glimt-toolkit"
 )
 
-File("demos").listFiles()?.forEach {
+File(settingsDir, "demos").listFiles()?.forEach {
```
maxandersen, line 12: cool - but what does that actually do differently?

melix: If you run gradle from a demo directory, the current directory will be the demo directory, which will not include any "demos" subdirectory, so no demos are included in the build :)

maxandersen: gotcha! so its to ensure they have the same fixed relative root. makes sense.

PR #14, opened 2025-12-23
feat: jbang enable all the demos and readme

maxandersen:
I got tired of fighting gradle and Intellij over clean and simple running the demos :)

This PR enables to run or open in IDE the demos directly - no gradle/maven etc. needed.

For now this uses `//DEPS dev.tamboui:...:LATEST` so it will use latest snapshot you have installed.

Probably want it to use :RELEASE or similar once its published.


maxandersen:
merging these as I need the scripts work reliably cross-platform for vhs screen recording :)

PR #17, opened 2025-12-24
tamboui-tfx a tachyonfx port

maxandersen:
a port of tachyonfx - draft as i'm not sure if this should be in the main repo or separate on its own?

wdyt?

tachyonfx lets you get animations by modifying on the cellbuffer directly which means it works "everywhere" but also means its not as easy to have it apply a widget level...but its quite nice anyways.
 
- **feat: tfx - initial port of tachyonfx**
- **fix: fix that styles should be patched rather than cleared**
- **fix: fixing various discrepencies between rust and java tfx demos**

### DIFF
--- a/demos/listelement-demo/src/main/java/dev/tamboui/demo/ListElementDemo.java
+++ b/demos/listelement-demo/src/main/java/dev/tamboui/demo/ListElementDemo.java
@@ -1,3 +1,6 @@
+//DEPS dev.tamboui:tamboui-widgets:LATEST
+//DEPS dev.tamboui:tamboui-toolkit:LATEST
+//DEPS dev.tamboui:tamboui-jline:LATEST
 /*
  * Copyright (c) 2025 TamboUI Contributors
  * SPDX-License-Identifier: MIT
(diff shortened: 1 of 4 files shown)


maxandersen:
I've rebased this - its not 100% complete but as separate addon that works independently + the fixes in core I'm interested in getting input/merged in.

tamboui-core/src/main/java/dev/tamboui/buffer/Buffer.java
```diff
@@ -191,7 +195,9 @@ public int setString(int x, int y, String string, Style style) {
             String symbol = new String(Character.toChars(codePoint));
 
             if (col >= area.left()) {
-                set(col, y, new Cell(symbol, style));
+                Cell existing = get(col, y);
```
maxandersen, line 199: without this styles can't be layered/blended. 

tfx is reliant on that behavior (its what ratatui does too afaics). 

but it also makes background style and basic manipulation much saner and simpler in widgets - without this text in some places does not get written on top of existing background.

melix: Maybe this should be an independent PR then? Does it break some demos?

melix:
@maxandersen I have fixed the demos which didn't launch, as well as failing tests. The animations seem a bit sloppy though.

maxandersen:
> @maxandersen I have fixed the demos which didn't launch, as well as failing tests. The animations seem a bit sloppy though.

Yeah they aren't polished. When I pushed draft i realized they work as generated ..will take a look to make them nicer. 

It does makes massive diff what terminal you are in for some of them though.

melix:
I added a FPS counter on one of the demos, I saw 60fps as expected, so the "sloppiness" may just be how it is given terminal capabilities.

tamboui-core/src/main/java/dev/tamboui/buffer/Buffer.java
maxandersen: i'll check - it breaks tfx as it assumes it can patch style - by setting just forground

Issue #18, opened 2025-12-25
cleaner separation in repo/docs between core immediate library and "stuff on top"


maxandersen:
currently examples and readme and repo in general blends immediate render with toolkit/tui higher level. 

be good to communicate to users but also just to easier verify we don't leak between the layers.

I think it would be good if we try make it clear in docs and in code/repo what fits where.

Either by separating out in separate repos (more work) or
at least within the repo has examples and docs co-located with what they show.

PR #21, opened 2025-12-26
reorg demos per module

maxandersen:
a way to get closer to #18 by having demos per module rather than one big one.

this also adds a `jbang run-demo` instead of the bash only runner that will run on all three major platforms including windows. `run-demo.sh`. is still there as shortcut.


melix (changes requested):
LGTM but I think you forgot to delete the top level demos.

maxandersen:
Cool. I've added more demos since and just makes me more convinced it would be good to have the demos more "local". 

I'll refresh this to include the new ones and clean it up.

melix:
The drawback of this approach is that it's less obvious to find all demos.

maxandersen:
jbang alias list . Lists them all

And the run-demo.sh will too.

melix:
I meant when browsing the repo to search for samples.

PR #31, opened 2025-12-29
Add missing flex support

### LINKED ISSUE #26: Flex layout is just an enum, no implementation
noticed Flex.* is not doing anything when porting demos.

So thats a todo :)

melix:
Fixes #26

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/layout/Flex.java
+++ b/tamboui-core/src/main/java/dev/tamboui/layout/Flex.java
@@ -36,5 +36,10 @@ public enum Flex {
     /**
      * Distribute space around elements.
      */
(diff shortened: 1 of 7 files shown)


maxandersen:
this isn't implementing the flex layout is it ? the ratatui implementations uses kasuari which is a implementation of cassowary constraint resolver  - this seems to just be some basic spacing ?

maxandersen:
i started testing this by implementing/porting the flex demo and it revealed a bunch of smaller other bugs in tabs and boarders ...so taking a bit longer. will open prs for those - as for flex itself it feels to me we are missing that constraint resolver to cover it?

melix:
No it's a simple flex support, no constraint resolver.

PR #32, opened 2025-12-29
Fix border rendering

### LINKED ISSUE #23: check borders of blocks are right
spotted that Quadrant has some dodgy corners:

Image

melix:
Fixes #23

### DIFF
--- a/tamboui-widgets/src/test/java/dev/tamboui/widgets/block/BlockTest.java
+++ b/tamboui-widgets/src/test/java/dev/tamboui/widgets/block/BlockTest.java
@@ -173,10 +173,10 @@ class BlockTest {
             .build()
             .render(buffer.area(), buffer);
         Buffer expected = Buffer.withLines(
-            "▖▄▄▄▄▄▗",
+            "▗▄▄▄▄▄▖",
             "▐     ▌",
             "▐     ▌",
-            "▘▀▀▀▀▀▝"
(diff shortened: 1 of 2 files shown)


tamboui-widgets/src/test/java/dev/tamboui/widgets/block/BlockTest.java
```diff
@@ -163,6 +163,90 @@ void titleInheritsBorderStyle() {
...
+        Block block = Block.builder()
+            .borders(Borders.ALL)
+            .borderType(BorderType.QUADRANT_INSIDE)
+            .build();
+        Buffer buffer = Buffer.empty(area);
+
+        block.render(area, buffer);
+
+        // Corners
+        assertThat(buffer.get(0, 0).symbol()).isEqualTo("▖");  // top-left
```
maxandersen: here using assertThat(buffer).isEqualTo(expected) would be nice ;)

melix: Fixed ;)

maxandersen:
running this give same weird corners - the characters in the test are wrong hence why it passes. working on fix.

maxandersen:
with my latest push I get:

image

which matches what ratatui blocks demo has if enables quardrant inside style.

melix:
Thanks!

PR #33, opened 2025-12-30
Add Figlet-style text rendering widget and demonstration

maxandersen:
Marking as draft as had to cut out tests as assertj for buffers was removed as module in https://github.com/tamboui/tamboui/commit/6ad0961bb821baa3065e7a632a58b91e731b00c3#r173657826

@melix how are modules going to use assertThat(buffer).equalsTo(expected) now?

Also not sure if we should put this in core/widgets. feel like something that is just as fine externally. 

- Introduced a new `Figlet` widget for rendering text banners using FIGlet fonts.
- Added `FigletFont` class to parse and render FIGlet fonts (.flf), supporting kerning, custom spacing, and trim options.
- Included several bundled FIGlet fonts (standard, slant, big, small, mini) as resources.
- Added `BundledFigletFont` helper to manage built-in fonts.
- Created a demonstration app (`FigletDemo`) and supporting Gradle project, showing cycling through fonts and editing text.
- Added documentation and short video tape (tape script) for the demo.
- Added tests for FIGlet font rendering.

Co-Assisted-by: cursor


melix:
> @melix how are modules going to use assertThat(buffer).equalsTo(expected) now?

I already answered on your other comment, the widgets are already using the test fixtures, there is no change, you can use them.

melix:
For me everything that we want to "support" should be in this repo. It also makes it easier for users.

maxandersen:
i rebased this and i noticed some fonts aren't rendering as nice which hints the figlet implementaiton is not fully complete  - not something I would necessarily find important to "support" in core but good enough as a first example how a widget contibution projbect would look like.

I suggest I take this and make a tamboui/tamboui-figlet repo to publish seperately and be used as example of how 3rd party widgets could work?

melix:
Right, I'm fine if it's in an external contrib repo.

maxandersen:
added https://github.com/tamboui/tamboui-widget-figlet - its not complete but pushed whats there so we can make PRs and iterate on it.

added issues for what i struggled with - but all files open for debate :)

PR #34, opened 2025-12-30
Refactor: Use OSHI for system metrics

maxandersen:
feat: jtopdemo made portable with oshi

current jtopdemo only works on linux.

this PR make use of OSHI library to have something that can get cpu utiliziation across all three major platforms.

In addition I played a bit with adding some dififerent cpu/memory visualizations.

jtop-demo

### DIFF
--- a/demos/jtop-demo/build.gradle.kts
+++ b/demos/jtop-demo/build.gradle.kts
@@ -6,7 +6,7 @@ description = "JTop - System monitor demo using the DSL module"
 
 dependencies {
     implementation(projects.tambouiToolkit)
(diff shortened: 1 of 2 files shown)


demos/jtop-demo/build.gradle.kts
```diff
@@ -6,6 +6,7 @@ description = "JTop - System monitor demo using the DSL module"
 
 dependencies {
     implementation(projects.tambouiToolkit)
+    implementation("com.github.oshi:oshi-core:6.9.2")
```
melix: You should put this in the libs.versions.toml file and use `libs.oshi.core` as a dependency notation.

PR #35, opened 2026-01-01
Add demo using MVC model

### LINKED ISSUE #16: TUI doesn't tick by default
The current implementation of the TUI abstraction is a bit confusing: the UI will never get refreshed by default because no tick events are sent. The documentation doesn't tell how a user is supposed to handle this.

Most likely the default should be to tick every x ms, as this is the most predictable pattern. The distinction with polling is not clear either.

melix:
This commit introduces a new "file manager" demo to illustrate how to build an application using an MVC model. In the process I discovered a few usability, API quirks, which this PR fixes:

- there is no longer need to call `registerElement` in a custom `render` method. This was an internal implementation detail that shouldn't have been exposed to users.
- truncation/ellipsis wasn't consistent in the widget module. There is now a Overflow enum which can be used to define how text which overflows should be handled
- fixes #16, which is to use apply a default tick period of 100ms

Added a tutorial for toolkit apps.

Happy New Year!

### DIFF
--- a/demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerDemo.java
+++ b/demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerDemo.java
@@ -17,8 +17,7 @@ import java.time.Duration;
  * <ul>
  *   <li>{@link FileManagerController} - Application state and file operations</li>
  *   <li>{@link DirectoryBrowserController} - Single directory browsing state</li>
- *   <li>{@link FileManagerView} - Main UI composition</li>
- *   <li>{@link DirectoryBrowserView} - Single browser panel rendering</li>
+ *   <li>{@link FileManagerView} - Main UI composition and browser panel rendering</li>
  *   <li>{@link FileManagerKeyHandler} - Event routing to controllers</li>
  * </ul>
  *


maxandersen:
just looking and i can the demo but it feels "broken"

image

notice that it does not fill out the full list - when scrolling down its as if it assumes a hard fixed height?

maxandersen:
also [Enter] on a file does not seem to do anything?

melix:
I'll take a look at the fixed height, that's not intended. `[enter]` allows entering a directory, it does nothing on files.

tamboui-widgets/src/main/java/dev/tamboui/widgets/text/Overflow.java
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.widgets.paragraph;
+
+/**
+ * Text overflow handling mode for paragraphs.
+ * Determines what happens when text doesn't fit in the available space.
+ */
+public enum Overflow {
```
maxandersen, line 11: why this change to Overflow instead of Wrap which is what ratatui uses ? 

Also, just noticed this is inside widgets.paragraph package - where I think this actually should be in widgets or possibly widgets.block as its not unique to paragraph.

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+### Essential Imports
+
+```java
+import dev.tamboui.toolkit.app.ToolkitRunner;
+import dev.tamboui.toolkit.element.Element;
+import dev.tamboui.toolkit.event.EventResult;
+import dev.tamboui.tui.Keys;
+import dev.tamboui.tui.event.KeyCode;
+import dev.tamboui.tui.event.KeyEvent;
+import dev.tamboui.widgets.paragraph.Overflow;  // For text overflow handling
```
maxandersen: paragraph.Overlflow isn't essential import is it?

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+
+        try (var runner = ToolkitRunner.create()) {
+            runner.run(() ->
+                view.render()
+                    .onKeyEvent(event -> handleEvent(event, controller))
+            );
+        }
+    }
+
+    private static EventResult handleEvent(KeyEvent event, TodoController ctrl) {
```
maxandersen, line 212: is it intentional to suggest to use all static methods for the main class?

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+        }
+        // Delegate text editing to the utility (handles chars, backspace, delete, arrows, home/end)
+        if (handleTextInputKey(ctrl.inputState(), event)) {
+            return EventResult.HANDLED;
+        }
+        return EventResult.UNHANDLED;
+    }
+
+    private static EventResult handleNormalMode(KeyEvent event, TodoController ctrl) {
+        if (Keys.isChar(event, 'a')) { ctrl.startInput(); return EventResult.HANDLED; }
```
maxandersen, line 236: given KeyEvent is a tui level construct it is a bit surprising can't just do `event.isChar('a')` ?

similar with the checks below.

melix: yup, but I think it doesn't belong to that PR.

tamboui-widgets/src/main/java/dev/tamboui/widgets/text/Overflow.java
melix: "Wrap" felt a bit restrictive as a name to me. It's really not just wrapping, it's about what to do with text overflow.

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+
+### Key Utilities
+
+The `Keys` class provides convenient matchers. Some methods include vim-style alternatives for ergonomic keyboard navigation:
+
+| Method | Keys Matched |
+|--------|--------------|
+| `Keys.isQuit(event)` | q, Q, Ctrl+C |
+| `Keys.isUp(event)` | Up arrow, k, K |
+| `Keys.isDown(event)` | Down arrow, j, J |
```
maxandersen, line 747: does that make sense ? that Keys will treat j/k as up/down always? and is shift+j really the same as j in vim? shift+j (J) is join lines not down.  and shift+k (K) is show documentation.

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+
+The flow is:
+1. User presses a key
+2. Your handler updates the controller state
+3. You return `EventResult.HANDLED`
+4. The framework calls your render function and redraws the UI
+
+This is efficient because the UI only redraws when something actually changes, rather than polling every 100ms.
+
+**Note:** For most applications, the default ticking behavior is sufficient. Only disable ticking if you have a specific reason to do so (e.g., minimizing CPU usage for simple, static UIs).
```
maxandersen, line 903: shouldn't this be flipped that notick() is the default and tick() should be enabled?

its only if you have live data you need to spend cpu time...and since it still ticks for key press and mouse events things will still update for majority of apps wont it?

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+        }
+    }
+}
+```
+
+### Running the Example
+
+```bash
+# If using JBang
+jbang TodoApp.java
```
maxandersen, line 1480: cool, but nowhere in example does it show how/where to put //DEPS for this to work?

docs/toolkit-tutorial.md
```diff
@@ -0,0 +1,1519 @@
...
+
+# Or with Gradle/Maven after adding dependencies
+./gradlew run
```
maxandersen: this doesn't work afaik and example does not show how maven setup is

tamboui-widgets/src/main/java/dev/tamboui/widgets/block/Block.java
```diff
@@ -12,6 +12,7 @@
 import dev.tamboui.text.Line;
 import dev.tamboui.text.Span;
 import dev.tamboui.widgets.Widget;
+import dev.tamboui.widgets.paragraph.Overflow;
```
maxandersen: code smell block.Block has dependency on higherlevel Paragraph package?

tamboui-widgets/src/main/java/dev/tamboui/widgets/block/Title.java
```diff
@@ -7,34 +7,37 @@
 import dev.tamboui.layout.Alignment;
 import dev.tamboui.text.Line;
 import dev.tamboui.text.Span;
+import dev.tamboui.widgets.paragraph.Overflow;
```
maxandersen: same smell. Overflow or rather Wrap should move up/out of paragraph.

docs/toolkit-tutorial.md
melix: Well, what caused this change is that I got hit several times by UI not refreshing, and it's difficult to find out why. I've been used to fixed rate rendering for a while, so having a default which works the other way around, with the risk of "lagging" is not my preferred choice.

docs/toolkit-tutorial.md
melix: This documents the _current_ behavior, not what we aim for. I think ideally we'll want to switch this with a keymap so that users can pick vi style, emacs style, ... But that is not the goal of that PR.

docs/toolkit-tutorial.md
melix: ^@#!!

docs/toolkit-tutorial.md
melix: Since the example is self contained and that state shouldn't leak into the App class and that the controller is not the one which should handle key events...

melix:
@maxandersen I fixed the most important issues. I'm keen on merging this to move forward and will work on fixing the other comments in different PRs, since they are not directly related to this one.

docs/toolkit-tutorial.md
maxandersen: yeah i'm also not super set on it being the other way - mainly just that when looking at performance and seeing java tuis pulling 100% cpu because of busy looping and others didnt' exhibit similar load felt bad.

but it can be separate PR and probably related to other things than just this 'tick'

tamboui-widgets/src/main/java/dev/tamboui/widgets/util/Overflow.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.widgets.util;
```
maxandersen: why not block  or even text ? paragraph,list etc. uses block /text.

Actually next to text near Line.java, Span.java, Text.java seems natural?

Other alternative is in style package if Overflow is considered part of style for more than just text?

tamboui-widgets/src/main/java/dev/tamboui/widgets/text/Overflow.java
maxandersen: ok, just to make it easier to port over examples ;)  but sure as long as its consistent and i think your reasons are valid and more reasons why it fits in .text package.

melix:
@maxandersen Moved `Overflow` to the `text` package :+1:

PR #36, opened 2026-01-02
fix: do full build in PRs

maxandersen:


maxandersen:
i assume this shouldn't hurt :) at least it would have caught some of the prs aren't actaully building.

PR #38, opened 2026-01-02
Add pseudo-CSS styling module

melix:
This commit introduces a new mechanism for styling UIs. While it can work with core and widget APIs, it is significantly easier to use with the toolkit module. Therefore, there are 2 demos which demonstrate how to use with and without toolkit, and an additional demo which shows how to use with custom components.

The CSS mechanism is inspired from Textual and supports selectors, background color, font weight, pseudo-classes, padding, border style and alignment.

### DIFF
--- a/tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
+++ b/tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
@@ -4,6 +4,9 @@
  */
 package dev.tamboui.css;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
@@ -56,4 +59,33 @@ public interface Styleable {
      * @return the parent element, or empty if this is a root element
      */
     Optional<Styleable> cssParent();
+
+    /**
+     * Builds the list of style types for an element, ordered by precedence.
+     * <p>
+     * The list contains the element's type and all parent types in the class
+     * hierarchy that implement {@link Styleable}. Types are ordered with lower
+     * precedence first (parent types) and higher precedence last (most specific type).
+     * <p>
+     * For example, if {@code MyPanel extends Panel} and both implement Styleable,
+     * calling {@code styleTypes(myPanel)} returns {@code ["Panel", "MyPanel"]}.
+     *
+     * @param element the element to get style types for
+     * @return an ordered list of style type names
+     */
+    static List<String> styleTypesOf(Styleable element) {
+        List<String> types = new ArrayList<>();
+
+        // Walk the superclass hierarchy for inherited types
+        Class<?> clazz = element.getClass().getSuperclass();
+        while (clazz != null && Styleable.class.isAssignableFrom(clazz)) {
+            types.add(0, clazz.getSimpleName());
+            clazz = clazz.getSuperclass();
+        }
+
+        // Add the element's declared type last (highest precedence)
+        types.add(element.styleType());
+
+        return Collections.unmodifiableList(types);
+    }
 }
--- a/tamboui-css/src/main/java/dev/tamboui/css/selector/TypeSelector.java
+++ b/tamboui-css/src/main/java/dev/tamboui/css/selector/TypeSelector.java
@@ -14,6 +14,9 @@ import java.util.Objects;
  * A type selector that matches elements by their type name.
  * <p>
  * Example: {@code Panel { ... }} matches all Panel elements.
+ * <p>
+ * Type selectors also match subclasses. For example, if {@code MyPanel extends Panel},
+ * the selector {@code Panel { ... }} will match both Panel and MyPanel elements.
  */
 public final class TypeSelector implements Selector {
 
@@ -34,7 +37,7 @@ public final class TypeSelector implements Selector {
 
     @Override
     public boolean matches(Styleable element, PseudoClassState state, List<Styleable> ancestors) {
-        return typeName.equals(element.styleType());
+        return Styleable.styleTypesOf(element).contains(typeName);
     }
 
     @Override
--- a/tamboui-css/src/test/java/dev/tamboui/css/selector/SelectorTest.java
+++ b/tamboui-css/src/test/java/dev/tamboui/css/selector/SelectorTest.java
@@ -31,6 +31,42 @@ class SelectorTest {
     }
 
     @Test
+    void typeSelectorMatchesSubclass() {
+        TypeSelector panelSelector = new TypeSelector("BasePanel");
+        MyPanel myPanel = new MyPanel();
+
+        // MyPanel extends BasePanel, so Panel selector should match MyPanel
+        assertThat(panelSelector.matches(myPanel, PseudoClassState.NONE, Collections.<Styleable>emptyList())).isTrue();
+    }
+
+    @Test
+    void typeSelectorMatchesExactTypeForSubclass() {
+        TypeSelector myPanelSelector = new TypeSelector("MyPanel");
+        MyPanel myPanel = new MyPanel();
+
+        // MyPanel selector should also match MyPanel
+        assertThat(myPanelSelector.matches(myPanel, PseudoClassState.NONE, Collections.<Styleable>emptyList())).isTrue();
+    }
+
+    @Test
+    void styleTypesReturnsCorrectOrder() {
+        MyPanel myPanel = new MyPanel();
+        List<String> types = Styleable.styleTypesOf(myPanel);
+
+        // Order should be parent first (lower precedence), child last (higher precedence)
+        assertThat(types).containsExactly("BasePanel", "MyPanel");
+    }
+
+    @Test
+    void styleTypesForSimpleElementReturnsOnlyOneType() {
+        Styleable element = createStyleable("Panel", null, Collections.<String>emptySet());
+        List<String> types = Styleable.styleTypesOf(element);
+
+        // TestStyleable has no Styleable parent classes
+        assertThat(types).containsExactly("Panel");
+    }
+
+    @Test
     void typeSelectorSpecificity() {
         TypeSelector selector = new TypeSelector("Panel");
         assertThat(selector.specificity()).isEqualTo(1);
@@ -278,5 +314,34 @@ class SelectorTest {
         public Optional<Styleable> cssParent() {
             return Optional.empty();
         }
+    }
+
+    /**
+     * Base panel class for testing type hierarchy matching.
+     */
+    private static class BasePanel implements Styleable {
+        @Override
+        public Optional<String> cssId() {
+            return Optional.empty();
+        }
+
+        @Override
+        public Set<String> cssClasses() {
+            return Collections.emptySet();
+        }
+
+        @Override
+        public Optional<Styleable> cssParent() {
+            return Optional.empty();
+        }
+    }
+
+    /**
+     * Subclass of BasePanel for testing type hierarchy matching.
+     * CSS selectors for both "BasePanel" and "MyPanel" should match instances of this class.
+     */
+    private static class MyPanel extends BasePanel {
+        // Inherits all methods from BasePanel
+        // styleType() defaults to "MyPanel" (getClass().getSimpleName())
     }
 }


tamboui-core/src/main/java/dev/tamboui/style/Color.java
```diff
@@ -211,11 +211,11 @@ public String toString() {
     Color BLUE = new Ansi(AnsiColor.BLUE);
     Color MAGENTA = new Ansi(AnsiColor.MAGENTA);
     Color CYAN = new Ansi(AnsiColor.CYAN);
-    Color WHITE = new Ansi(AnsiColor.WHITE);
-    Color GRAY = new Ansi(AnsiColor.BRIGHT_BLACK);
+    Color WHITE = new Ansi(AnsiColor.BRIGHT_WHITE);
+    Color GRAY = new Ansi(AnsiColor.WHITE);  // ANSI WHITE (7) renders as light gray
```
maxandersen, line 215: doesnt that depend on your terminal ? seems counter intuitive to use different definitions here?

tamboui-css/src/main/java/dev/tamboui/css/cascade/CascadeResolver.java
```diff
@@ -0,0 +1,181 @@
...
+import dev.tamboui.css.model.PropertyValue;
+import dev.tamboui.css.model.Rule;
+import dev.tamboui.css.property.PropertyRegistry;
+import dev.tamboui.style.Color;
+import dev.tamboui.style.Modifier;
+
+import java.util.*;
+
+/**
+ * Resolves CSS cascade and specificity to produce final computed styles.
```
maxandersen, line 16: css are tricky - is this following a specific pattern or inspired by existing stuff? or mainly claude coded?

either is fine - just to grok what to look for ;)

tamboui-css/src/main/java/dev/tamboui/css/engine/StyleEngine.java
```diff
@@ -0,0 +1,398 @@
...
+            }
+        }
+
+        return variables;
+    }
+
+    private String readClasspathResource(String resource) throws IOException {
+        InputStream is = getClass().getResourceAsStream(resource);
+        if (is == null) {
+            // Try without leading slash
```
maxandersen, line 391: is that normal/needed?

tamboui-core/src/main/java/dev/tamboui/style/Color.java
melix: I think it's a leftover of trial to fix a bug...

tamboui-css/src/main/java/dev/tamboui/css/engine/StyleEngine.java
melix: Just a convenience

demos/css-demo/src/main/resources/themes/dark.tcss
```diff
@@ -0,0 +1,105 @@
...
+$highlight-bg: #333333;
+
+/* Base styles */
+* {
+    color: $fg-primary;
+    background: $bg-primary;
+}
+
+/* Panel styling */
+Panel {
```
maxandersen, line 17: Question: I believe textualize lets Panel apply to any instances of Panel or subclasses hereof. Are tambotui css applying same logic ?

demos/css-no-toolkit-demo/src/main/java/dev/tamboui/demo/CssNoToolkitDemo.java
```diff
@@ -0,0 +1,483 @@
...
+            .split(area);
+
+        renderHeader(frame, layout.get(0));
+        renderMainContent(frame, layout.get(1));
+        renderFooter(frame, layout.get(2));
+    }
+
+    private void renderHeader(Frame frame, Rect area) {
+        // Resolve style for header using CSS
+        SimpleStyleable headerStyleable = new SimpleStyleable("Panel", null, Set.of("status"));
```
maxandersen, line 196: I feel these lines could be greatly simplified somehow. 

Not sure what yet :)

tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+package dev.tamboui.css;
+
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Interface for elements that can be styled via CSS.
+ * <p>
+ * Elements implementing this interface can be targeted by CSS selectors
```
maxandersen, line 16: Is this limited to Elements ?

tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Returns the style type name used for type selectors.
+     * <p>
+     * For example, a Panel element would return "Panel" to match
+     * the CSS selector {@code Panel { ... }}.
+     *
+     * @return the style type name (never null)
+     */
+    String styleType();
```
maxandersen: Could default implement this as this.getClass().getsimpleName() ?

tamboui-css/src/main/java/dev/tamboui/css/cascade/CascadeResolver.java
maxandersen: Inspired by textual implemented by Claude code iterations.

tamboui-css/src/main/java/dev/tamboui/css/cascade/CascadeResolver.java
```diff
@@ -0,0 +1,181 @@
...
+
+        return builder.build();
+    }
+
+    private String resolveVariables(String value, Map<String, String> variables) {
+        if (value == null || !value.contains("$")) {
+            return value;
+        }
+        String result = value;
+        for (Map.Entry<String, String> entry : variables.entrySet()) {
```
maxandersen: This isn't a deterministic replacement as it will depend on the sequence of variables. 

Should be driven by extracting property expressions in the string.

tamboui-css/src/main/java/dev/tamboui/css/cascade/ResolvedStyle.java
```diff
@@ -0,0 +1,236 @@
...
+import dev.tamboui.widgets.block.Padding;
+
+import java.util.*;
+
+/**
+ * Represents the computed style for an element after cascade resolution.
+ * <p>
+ * Contains all resolved CSS properties that can be applied to an element.
+ */
+public final class ResolvedStyle {
```
maxandersen, line 22: Question: I was wondering if css styles should allow set character as it's used as "style" for ie. Hatch in scrollbars or certain characters in certain elements ?

tamboui-css/src/main/java/dev/tamboui/css/engine/StyleEngine.java
```diff
@@ -0,0 +1,398 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * Main entry point for CSS styling.
+ * <p>
+ * The StyleEngine manages stylesheets, resolves styles for elements,
+ * and supports live stylesheet switching for theming.
+ * <p>
+ * Usage:
+ * <pre>
```
maxandersen, line 55: What's the semantics of names stylesheets and multiple inline stylesheets ?

tamboui-css/src/main/java/dev/tamboui/css/engine/StyleEngine.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+
+        String css;
+        if (entry.classpathResource != null) {
+            css = readClasspathResource(entry.classpathResource);
+        } else if (entry.filePath != null) {
+            css = new String(Files.readAllBytes(entry.filePath), StandardCharsets.UTF_8);
+        } else {
+            // Inline stylesheet - cannot reload
+            return;
```
maxandersen, line 265: Javadoc states it would throw exception? Javadoc wrong I guess and it should say inlined are ignored ?

tamboui-css/src/main/java/dev/tamboui/css/engine/StyleEngine.java
```diff
@@ -0,0 +1,398 @@
...
+         */
+        void onStyleChange();
+    }
+
+    /**
+     * Internal entry for named stylesheets.
+     */
+    private static final class StylesheetEntry {
+        final Stylesheet stylesheet;
+        final String classpathResource;
```
maxandersen: Should these just be a functional provider ?

maxandersen:
is it intentional that the list items/title has black background or bug in rendering/css?



image


in no toolkit demo it renders more "sane":

image

maxandersen:
the custom demo fails hard with this exception if you type `rgb(2,304,4);`

```
Running custom-component-demo...
Exception in thread "main" java.lang.IllegalArgumentException: RGB values must be 0-255: (2, 340, 5)
        at dev.tamboui.style.Color$Rgb.<init>(Color.java:141)
        at dev.tamboui.style.Color.rgb(Color.java:251)
        at dev.tamboui.css.property.ColorConverter.convert(ColorConverter.java:95)
        at dev.tamboui.css.property.PropertyRegistry.convertColor(PropertyRegistry.java:69)
        at dev.tamboui.css.cascade.CascadeResolver.buildResolvedStyle(CascadeResolver.java:116)
        at dev.tamboui.css.cascade.CascadeResolver.resolve(CascadeResolver.java:98)
        at dev.tamboui.css.engine.StyleEngine.resolve(StyleEngine.java:256)
```

 I wonder if should consider having a TUIException or similar to be easier to guard these ?

demos/css-demo/src/main/resources/themes/dark.tcss
```diff
@@ -0,0 +1,105 @@
...
+    background: black;
+    border-color: $border-color;
```
maxandersen, line 19: i notice textualize allows:

`border: rounded $border-color`

should we have similar shorthand?

demos/css-no-toolkit-demo/src/main/java/dev/tamboui/demo/CssNoToolkitDemo.java
melix: By using `toolkit` ;)

tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
melix: Good idea

tamboui-css/src/main/java/dev/tamboui/css/cascade/ResolvedStyle.java
melix: I thought about it too, but I need to think about it more. Typically some borders require 6 elements, others 8. So probably we need something that depends on the border type.

demos/css-no-toolkit-demo/src/main/java/dev/tamboui/demo/CssNoToolkitDemo.java
maxandersen: well - I was thinking more like some form of shorthand like: `WidgetStyle.of(Panel.class,"#status").style()` should be feasible.

tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
maxandersen: having read the current resolver i'm not sure its a good idea as we dont traverse the class hiearchy so a subclass would of Panel would no longer get Panel applied to it if default implmeent it getSimpleName()?

so need some consideration on how this should work to get decent consistency without too much repetition.

tamboui-css/src/main/java/dev/tamboui/css/cascade/ResolvedStyle.java
maxandersen: would be intereesting to see if textual allow restyling of those parts. I reckon they just have a very specific selector  they lookup with to allow override but haven't verified.

melix:
> is it intentional that the list items/title has black background or bug in rendering/css?

That was a bug. I'm investigating how by re-architecturing we could avoid these.

tamboui-css/src/main/java/dev/tamboui/css/engine/StyleEngine.java
melix: Javadocs clarified

demos/css-demo/src/main/resources/themes/dark.tcss
melix: Let's keep this in mind for future improvements. Create a ticket.

tamboui-css/src/main/java/dev/tamboui/css/Styleable.java
melix: Only `Element` would use it in practice

maxandersen:
this is awesome - lets merge and go play :)

PR #41, opened 2026-01-04
Introduce key bindings

melix:
This commit adds the concept of KeyMap, which removes the hardcoded `isUp`/`isLeft` methods, which were bound to `vi`. Now there is a proper keymap concept, with a simple version which simply uses arrows, and then optional keymaps for vi, emacs, ...

The KeyEvent class delegates to KeyMap to make the API nicer to use.

### DIFF
--- a/docs/toolkit-tutorial.md
+++ b/docs/toolkit-tutorial.md
@@ -739,7 +739,7 @@ public EventResult handleKeyEvent(KeyEvent event, boolean focused) {
 
 ### Key Utilities
 
-KeyEvent provides convenience methods that delegate to the configured `KeyMap`. The default keymap (`standard`) uses only arrow keys:
+KeyEvent provides convenience methods that delegate to the configured `Bindings`. The default bindings (`standard`) uses only arrow keys:
 
 | Method | Keys Matched (Standard) |
 |--------|-------------------------|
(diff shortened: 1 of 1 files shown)


tamboui-tui/src/main/java/dev/tamboui/tui/keymap/Action.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.tui.keymap;
+
+/**
+ * Semantic actions that can be triggered by key events.
```
maxandersen: how about mouse ? it can do up/scroll etc.

would bindings be a better term than keymaps ?

tamboui-tui/src/main/java/dev/tamboui/tui/keymap/Action.java
```diff
@@ -0,0 +1,99 @@
...
+ * <p>
+ * Actions represent the intent behind a key press, allowing the same
+ * action to be bound to different keys depending on the keymap.
+ */
+public enum Action {
```
maxandersen: is it really a fixed enum? how does one extend this list of actions?

tamboui-tui/src/main/java/dev/tamboui/tui/keymap/Action.java
melix: Mouse events are now handled

tamboui-tui/src/main/java/dev/tamboui/tui/keymap/Action.java
melix: Made it open

PR #44, opened 2026-01-04
Add action binding

melix:
This PR builds on top of #41 and introduces action binding. In addition, it defines an `@OnAction` annotation, which can be used by components in place of `onKeyEvent`/`onMouseEvent`. There are 2 ways to use the `@OnAction` annotation. The preferred way is to use the annotation processor, but there is a graceful fallback using reflection.

### DIFF
--- a/tamboui-annotations/src/main/java/dev/tamboui/annotations/bindings/OnAction.java
+++ b/tamboui-annotations/src/main/java/dev/tamboui/annotations/bindings/OnAction.java
@@ -2,7 +2,7 @@
  * Copyright (c) 2025 TamboUI Contributors
  * SPDX-License-Identifier: MIT
  */
-package dev.tamboui.annotations;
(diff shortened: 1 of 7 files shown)


tamboui-annotations/src/main/java/dev/tamboui/annotations/OnAction.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.annotations;
```
maxandersen: is this intended to be generic and usable beyond the tui/toolkit?

melix: I guess not. I created a single annotations package for simplicity. We could move that type to `dev.tamboui.annotations.binding`.

PR #46, opened 2026-01-05
Add support for Java modules

### LINKED ISSUE #42: Add explicit Java module descriptors
Modern Java projects should make use of modern Java features whenever possible.

Considering the project targets Java 8 as a minimum this would require adding the module descriptor in the versioned space i.e, `/META-INF/versions/11/`.

FWIW the moditect Gradle plugin is no longer maintained and likely incompatible with Gradle configuration cache.

melix:
This configures the build so that all libraries are declared as proper Java modules. The modules are found in a Java 11 source set, because our baseline is Java 8. This makes the build setup more complicated because IntelliJ will not understand the model properly, but it works both in the IDE and command line, which is what we expect.

In the process, a task to verify that we don't create split packages has been added. It actually discovered one, so the `widgets` package in the core module has been renamed to `widget`.

A specific demo has been implemented showing the configuration.

Fixes #42

### DIFF
--- a/buildSrc/src/main/java/dev/tamboui/build/SplitPackageCheckTask.java
+++ b/buildSrc/src/main/java/dev/tamboui/build/SplitPackageCheckTask.java
@@ -1,0 +1,217 @@
+package dev.tamboui.build;
+
+import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
+import org.gradle.api.file.FileTree;
+import org.gradle.api.file.RegularFileProperty;
+import org.gradle.api.provider.ListProperty;
+import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.Nested;
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.PathSensitive;
+import org.gradle.api.tasks.PathSensitivity;
+import org.gradle.api.tasks.TaskAction;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.stream.Stream;
+
+/**
+ * A task that checks for split packages across source sets.
+ * <p>
+ * Split packages occur when the same Java package exists in multiple modules,
+ * which causes problems with the Java Platform Module System. This task
+ * analyzes source directories and reports any packages that appear in more than one module.
+ */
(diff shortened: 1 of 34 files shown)


buildSrc/src/main/java/dev/tamboui/build/SplitPackageCheckTask.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.stream.Stream;
+
+/**
+ * A task that checks for split packages across source sets.
+ * <p>
+ * Split packages occur when the same Java package exists in multiple modules,
+ * which causes problems with the Java Platform Module System. This task
+ * analyzes source directories and reports any packages that appear in more than one module.
+ */
+@CacheableTask
```
aalmiray, line 36: Ideally this should be provided by an external plugin, right? 
Of course it's OK to have it here if no suitable solution is available in the GPP

buildSrc/src/main/kotlin/dev.tamboui.parent.gradle.kts
```diff
@@ -0,0 +1,27 @@
+import dev.tamboui.build.SplitPackageCheckTask
+
+plugins {
+    base
+    id("dev.tamboui.maven-central-publishing")
+}
+
+val splitPackageCheck = tasks.register<SplitPackageCheckTask>("splitPackageCheck") {
```
aalmiray, line 8: FWIW split packages should be check also against dependencies.
I guess it's OK to start with project dependencies

buildSrc/src/main/java/dev/tamboui/build/SplitPackageCheckTask.java
melix: Ideally this should be built in Gradle :) For now it's just a poor man's solution which will work well for us.

buildSrc/src/main/kotlin/dev.tamboui.parent.gradle.kts
melix: It would be a mistake for us to use a package from an external dependency in any case.

buildSrc/src/main/java/dev/tamboui/build/SplitPackageCheckTask.java
aalmiray: I know. Didn’t want to open that can of worms 😉

buildSrc/src/main/kotlin/dev.tamboui.parent.gradle.kts
aalmiray: True. But how would you enforce it?

Additionally, consumed dependencies may have split packages between them, even if your codebase does not. Yes, those dependencies should be fixed when that’s the case. 

Now, given that the library has a minimum set of external dependencies at the moment this may not be an issue for now. Just making sure to flag this as a potential problem.

PR #49, opened 2026-01-06
Add style property system for style-aware widgets

melix:
This commit introduces a clean abstraction for widgets to resolve style properties from external sources (like CSS) while supporting programmatic overrides.

It is designed around a few core concepts:
  - PropertyKey: Identifies a style property by name and knows how to convert string values to typed values (e.g., "color" → Color) by associating it with a `PropertyConverter`.
  - PropertyConverter: Transforms strings into typed values (e.g., "#FF0000" → Color.rgb(255, 0, 0))
  - PropertyResolver: Interface that resolves property values by key, allowing widgets to be agnostic of the styling source
  - StyledProperty: Holds a property value with resolution order: programmatic value → resolver → default. This is the key to agnostic rendering, providing consistent resolution of properties across all widgets.

Widgets can now accept a styleResolver and automatically resolve properties like background, foreground, border-color, and border-type. This keeps widgets decoupled from any specific styling system while enabling rich style integration.

This considerably simplifies code, since `Element`s don't have to deal with styles anymore, and that when not using `toolkit`, styles don't have to be handled unless they are CSS specific.

### DIFF
--- a/docs/widget-development.md
+++ b/docs/widget-development.md
@@ -119,7 +119,7 @@ public interface PropertyResolver {
 }
 ```
 
-The CSS module provides `CssStyleResolver` which implements this interface. Widgets don't need to know about CSS - they just accept a `PropertyResolver`.
+The CSS module provides `CssStyleResolver` which implements this interface. Widgets don't need to know about CSS - they or more often their Builder just accept a `PropertyResolver` which is used at creation time to apply styles.
 
 ### StyledProperty
 


docs/widget-development.md
```diff
@@ -0,0 +1,553 @@
...
+public interface PropertyResolver {
+    <T> Optional<T> get(PropertyKey<T> key);
+
+    static PropertyResolver empty() {
+        return key -> Optional.empty();
+    }
+}
+```
+
+The CSS module provides `CssStyleResolver` which implements this interface. Widgets don't need to know about CSS - they just accept a `PropertyResolver`.
```
maxandersen:
```suggestion
The CSS module provides `CssStyleResolver` which implements this interface. Widgets don't need to know about CSS - they or more often their Builder just accept a `PropertyResolver` which is used at creation time to apply styles.
```
 
Are widgets expected/able to hold on to the resolver and use during rendering? or possilby listen to changes to avoid reconstruction ? 

not sure needed -just trying to grok where the intended/expected limits of the resolvers lifecycle is.

tamboui-core/src/main/java/dev/tamboui/style/PropertyResolver.java
```diff
@@ -0,0 +1,63 @@
...
+ * <h2>Usage in widgets:</h2>
+ * <pre>{@code
+ * public void render(Rect area, Buffer buffer, PropertyResolver resolver) {
+ *     Color borderColor = resolver.get(BORDER_COLOR).orElse(Color.WHITE);
+ *     // render with borderColor...
+ * }
+ * }</pre>
+ */
+@FunctionalInterface
+public interface PropertyResolver {
```
maxandersen: should this be a StylePropertyResolver or rather StyleResolver that take PropertyKey's ? or could this be something used for more than just Styles (then it shouldn't be in style ?

tamboui-core/src/main/java/dev/tamboui/style/PropertyKey.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.style;
```
maxandersen, line 5: thinking out loud - should the core notion of properties be separate from style? i.e. bindings and others would have similar mechanis.

maxandersen:
just a few clarifications - and thoughts; but beyond the clarification of what actually gets resolver the other parts can probably be iterated over as long as we dont plan on doing GA as the first release :)

docs/widget-development.md
melix: The resolver is only used at build time, it's intentionally simple (no change listening).

tamboui-core/src/main/java/dev/tamboui/style/PropertyResolver.java
melix: I think renaming to `StylePropertyResolver` makes sense.

tamboui-core/src/main/java/dev/tamboui/style/PropertyKey.java
melix: well it is bound to property converters, which are style specific now. I don't really foresee using this somewhere else now, but we can refactor if needed.

PR #56, opened 2026-01-07
User experience fixes + coding assistant demo

melix:
This commit introduces a new "AI Coding assistant" like demo. In the process, several pain points were found and addressed:

- added BorderType.NONE and Panel.borderless()
- added list(ListItem...) factory methods to Toolkit
- added displayOnly() to ListContainer for cases where there shouldn't be any highlighted element (useful for logs, messages, ...)
- added scrollToEnd() to ListState and ListContainer
- added scheduler API to ToolkitRunner for periodic operations
- added a WaveText widget and demo for progress updates
- now all styled elements have a `focusable()` method. Without this some components cannot receive key events, which was painful.

### DIFF
--- a/demos/fakodex-demo/src/main/java/dev/tamboui/demo/coding/CodingAssistantDemo.java
+++ b/demos/fakodex-demo/src/main/java/dev/tamboui/demo/coding/CodingAssistantDemo.java
@@ -1,3 +1,7 @@
+//DEPS dev.tamboui:tamboui-toolkit:LATEST
+//DEPS dev.tamboui:tamboui-jline:LATEST
+//FILES styles/codex.tcss=../../../../../resources/styles/codex.tcss
+
 /*
  * Copyright (c) 2025 TamboUI Contributors
  * SPDX-License-Identifier: MIT


davsclaus:
Nice work, the display only for logs is something I am going to need. Glad to see that this project innovates with demos that shows the features and also help validate their usage and its excellent for others to learn and "copy from".

demos/fakodex-demo/src/main/java/dev/tamboui/demo/coding/CodingAssistantDemo.java
```diff
@@ -0,0 +1,710 @@
+/*
```
maxandersen, line 5:
```suggestion
//DEPS dev.tamboui:tamboui-toolkit:LATEST
//DEPS dev.tamboui:tamboui-jline:LATEST
//FILES styles/codex.tcss=../../../../../resources/styles/codex.tcss

/*
```

melix: mmm, that's unfortunate. Does this mean we have to redeclare all resources? JBang can't be provided with the "base directories" where to look for?

melix:
> Glad to see that this project innovates with demos that shows the features and also help validate their usage and its excellent for others to learn and "copy from".

I am convinced that it's by trying out our APIs that we will see what we miss, what is cumbersome to use and what can be improved :+1:

demos/fakodex-demo/src/main/java/dev/tamboui/demo/coding/CodingAssistantDemo.java
maxandersen: @melix not at the moment. But I have considered at least expose a property i.e. ${source-root} or similar that would be derived based on package statement if present. 

Then it would be:

```
//FILES styles/codex.tcss=${source-root}/../resources/styles/codex.tcss
```

note, if there is more than one resource we can also do:

```
//FILES /=../../../../../resources/**/*
```

and just mount everything.

Downside of that though is then you can't run it without cloning the repo locally.


But as of now above is the best option.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
```diff
@@ -58,12 +63,18 @@ public final class ToolkitRunner implements AutoCloseable {
...
     private final EventRouter eventRouter;
     private final DefaultRenderContext renderContext;
+    private final ScheduledExecutorService scheduler;
 
     private ToolkitRunner(TuiRunner tuiRunner) {
         this.tuiRunner = tuiRunner;
         this.focusManager = new FocusManager();
         this.eventRouter = new EventRouter(focusManager);
         this.renderContext = new DefaultRenderContext(focusManager, eventRouter);
+        this.scheduler = new ScheduledThreadPoolExecutor(1, r -> {
```
maxandersen, line 73: nice and simple - won't start a  thread until first schedule, but it will keep thread around after first task run - but thats probably fine in most cases.

I was  thinking if virtual threads executor service would be something - but dont think there is one that will do scheduling nor that its relevant to do multiple concurrent virtual threads so probably sometihng for a non-scheduled bound tasks and not tied to toolkit?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/WaveTextElement.java
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * An element that displays text with an animated wave brightness effect.
+ * <p>
+ * By default, a dark "shadow" moves through otherwise bright text.
+ * This can be inverted to have a bright peak moving through dim text.
+ *
+ * <pre>{@code
+ * // Dark shadow on bright text (default)
+ * waveText("Loading...", Color.CYAN)
```
maxandersen, line 23: If I grok this right then an alternative to this is how 
https://github.com/tamboui/tamboui/pull/17 works.

It allows you to apply such "animations"/styles to any regions/areas/widgets. Not just Text.

Also reminded me that PR has some important fixes for proper background rendering. i should get it rebased :)

maxandersen (commented):
I thinks this all look good except I'm not convinced the WaveText* is right approach to have dedicated component to that when can be applied with something like the effects library.

If you want to merge with wavetext in now and we can decided later if to take it out or leave it out of this PR and we can bring it back i'll let be up to you @melix 

Let me go rebase that styling PR :)

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
melix: We could change that in the future using service loader probably.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/WaveTextElement.java
melix: That's interesting. Not sure how this would be aware of which area to apply the style to but I like the idea. Let's merge this as is, and we can replace it once the tfx PR is ready and shows how to do the same.

PR #58, opened 2026-01-08
fix: allow selecting one of many backend

maxandersen:
this isn't really relevant before we have panama or some other backend provider -but its nice to use to test panama branch.

with this you get more help when multiple backends present:

```
jbang colors-rgb-demo
Exception in thread "main" java.lang.IllegalStateException: Multiple backend providers found on classpath.
Include only one backend or specify which provider to use by setting the tamboui.backend system property or TAMBOUI_BACKEND environment variable to one of:
jline (dev.tamboui.backend.jline.JLineBackendProvider)
panama (dev.tamboui.backend.panama.PanamaBackendProvider)
        at dev.tamboui.terminal.BackendFactory.create(BackendFactory.java:112)
        at dev.tamboui.demo.ColorsRgbDemo.run(ColorsRgbDemo.java:48)
        at dev.tamboui.demo.ColorsRgbDemo.main(ColorsRgbDemo.java:44)
```

or if you use bad name:
```
jbang -Dtamboui.backend=jlinek colors-rgb-demo
Exception in thread "main" java.lang.IllegalStateException: BackendProvider 'jlinek' not found. Available providers:
jline (dev.tamboui.backend.jline.JLineBackendProvider)
panama (dev.tamboui.backend.panama.PanamaBackendProvider)
        at dev.tamboui.terminal.BackendFactory.create(BackendFactory.java:101)
        at dev.tamboui.demo.ColorsRgbDemo.run(ColorsRgbDemo.java:48)
        at dev.tamboui.demo.ColorsRgbDemo.main(ColorsRgbDemo.java:44)
```

or just do it right:

`jbang -Dtamboui.backend=jline colors-rgb-demo`

or 

`TAMBOUI_BACKEND=jline jbang colors-rgb-demo`

...you can also use full classname with a dot in the name which skips creating any providers.

### DIFF
--- a/tamboui-widgets/demos/colors-rgb-demo/src/main/java/dev/tamboui/demo/ColorsRgbDemo.java
+++ b/tamboui-widgets/demos/colors-rgb-demo/src/main/java/dev/tamboui/demo/ColorsRgbDemo.java
@@ -1,7 +1,6 @@
 ///usr/bin/env jbang "$0" "$@" ; exit $?
 //DEPS dev.tamboui:tamboui-widgets:LATEST
 //DEPS dev.tamboui:tamboui-jline:LATEST
-//DEPS dev.tamboui:tamboui-panama-backend:LATEST
 /*
  * Copyright (c) 2025 TamboUI Contributors
  * SPDX-License-Identifier: MIT


tamboui-widgets/demos/colors-rgb-demo/src/main/java/dev/tamboui/demo/ColorsRgbDemo.java
```diff
@@ -1,7 +1,7 @@
 ///usr/bin/env jbang "$0" "$@" ; exit $?
 //DEPS dev.tamboui:tamboui-widgets:LATEST
 //DEPS dev.tamboui:tamboui-jline:LATEST
-
+//DEPS dev.tamboui:tamboui-panama-backend:LATEST
```
melix: :x: This needs to be removed, because it will break on `main` branch.

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -53,59 +55,41 @@ private BackendFactory() {
...
+            userSelectedProvider = System.getenv("TAMBOUI_BACKEND");
         }
+        String providerSpec = userSelectedProvider;
+        boolean isClassName = providerSpec != null && providerSpec.contains(".");
 
         // Load all available providers (needed for simple name matching or auto-discovery)
         ServiceLoader<BackendProvider> loader = ServiceLoader.load(BackendProvider.class);
-        List<BackendProvider> providers = new ArrayList<>();
-        loader.forEach(providers::add);
+        List<BackendProvider> providers = StreamSupport.stream(loader.spliterator(), false)
```
maxandersen, line 67: this change means BackendProviders *MUST* be serviceloader provided + we do load all the serviceloader provided classes even if we explicitly set a classname and thus don't need it and its classes loaded.

is that necessary?

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -53,59 +55,41 @@ private BackendFactory() {
...
+            if (isClassName) {
+                details = " for class name '" + providerSpec + "'";
+            } else if (providerSpec != null && !providerSpec.isEmpty()) {
+                details = " for provider name '" + providerSpec + "'";
+            }
             throw new IllegalStateException(
-                "No BackendProvider found on classpath. " +
-                "Add a backend dependency such as tamboui-jline, or set the tamboui.backend system property " +
-                "or TAMBOUI_BACKEND environment variable."
+                "No BackendProvider found on classpath" + details + ".\n" +
```
maxandersen, line 88: you dont want the details provided to the user in the exception?

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
melix: To work with Java modules, yes. Note that it's the provider that we load, not the backend itself, so IMO perfectly fine, especially knowing that in practice it's extremely unlikely a user would have more than one.

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
melix: There are details provided, the string is computed depending on the actual content of the filtered collection.

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
maxandersen: I dont think it would be rare to have multiple. 

Ie. In JVM mode of jbang i plan to have jline and Panama present to choose which makes sense based on what JVM I'm running on. 

Similar even when running on higher JVM the os might not have the matching library Panama uses.

melix: IMO it's cumbersome if the user has to choose or expose an env variable. That's fine for us, but I doubt a user would like to do this.

> Similar even when running on higher JVM the os might not have the matching library Panama uses.

Can you clarify this one? AFAIK Panama bindings don't change based on the VM, and if the os doesn't have the library, then you're in bigger trouble.

maxandersen: > IMO it's cumbersome if the user has to choose or expose an env variable. That's fine for us, but I doubt a user would like to do this.

That is for the case of being able to explicitly request a specific provider.

I expect we'll (once we have more than one provider :) add either near this code or higher above a "dependent on env/java/lib availability auto-detect possible backends" ...kinda similar to what jline has.

Thats why I was after avoiding iterating them all.
  
> > Similar even when running on higher JVM the os might not have the matching library Panama uses.
> 
> Can you clarify this one? AFAIK Panama bindings don't change based on the VM, and if the os doesn't have the library, then you're in bigger trouble.

jbang runs everywhere jvm are available also on OS's with funky alternative libc implementations...alpine, solaris, ibm as400 and more fun - I haven't tested if the termios bindings cause issues but it definitely do on macos (unti we fix it) and I'll be after being able to make a JVM based cli that doesn't fall-over just because the panama implementation isn't working on a weird OS, and be able to provide a fallback (jline for now which in itself have additional fallbacks)

maxandersen:
merging this as its good enough - we can revisit it when we have more experience with multiple backends and if panama prevents graceful fallbcks :)

PR #60, opened 2026-01-08
Add inline display mode

melix:
This commit introduces an "inline display" mode, which allows reusing widgets, but without being in full screen mode. There are several utilities. The first one is a way to get an ANSI string from a `Buffer`, which serves simple use cases.

The second one, more useful, is an `InlineDisplay` utility, which allows "reserving" space, e.g a few lines, perform rendering on these, then either clear the result when done, or continue.

### DIFF
--- a/demos/inline-progress-demo/src/main/java/dev/tamboui/demo/InlineProgressDemo.java
+++ b/demos/inline-progress-demo/src/main/java/dev/tamboui/demo/InlineProgressDemo.java
@@ -14,7 +14,9 @@ import dev.tamboui.style.Style;
 import dev.tamboui.text.Line;
 import dev.tamboui.text.Span;
 import dev.tamboui.text.Text;
-import dev.tamboui.tui.InlineDisplay;
+import dev.tamboui.inline.InlineDisplay;
+import dev.tamboui.terminal.Backend;
+import dev.tamboui.terminal.BackendFactory;
 import dev.tamboui.widgets.gauge.Gauge;
 import dev.tamboui.widgets.paragraph.Paragraph;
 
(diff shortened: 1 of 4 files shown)


tamboui-tui/src/main/java/dev/tamboui/tui/InlineDisplay.java
```diff
@@ -0,0 +1,357 @@
...
+ *         });
+ *         display.println("Step " + i + " completed");
+ *         Thread.sleep(100);
+ *     }
+ * }
+ * }</pre>
+ *
+ * @see Buffer#toAnsiString()
+ */
+public final class InlineDisplay implements Closeable {
```
maxandersen: does this need to be in tui? wouldn't widget or event core suffice?

melix: Doesn't have to, but it follows a similar design pattern, which is that it uses classes from core with some lifecycle management. Whether we move it to core or not is really whether we consider this higher level or not (the ANSI conversion lives in core though).

melix: @maxandersen do you want me to move it into core? It's really a judgement call IMO.

maxandersen: i feel inline is quite core/basic feature so having the ability there and the fwk build on top of that rather than one MUST use tui to render these.

so +1 for putting in core.

melix: Ok will do!

melix: @maxandersen done

PR #62, opened 2026-01-10
filemanager with images

maxandersen:
wanted to add image viewer to filemanager - and it works - but only in xterm ...ghosty and iterm2 with kitty/iterm2 gets stuck. so this is just a draft for now.

### DIFF
--- a/demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerView.java
+++ b/demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerView.java
@@ -304,15 +304,15 @@ public class FileManagerView implements Element {
         }
 
         String fileName = file.getFileName().toString();
-        boolean isPng = fileName.toLowerCase().endsWith(".png");
+        boolean isImage = isImageFile(fileName);
 
         // Calculate dialog size (use most of the screen)
         int dialogWidth = Math.min(area.width() - 4, 120);
         int dialogHeight = Math.min(area.height() - 4, 40);
 
-        // For PNG images, create the image widget first to get protocol name
+        // For image files, create the image widget first to get protocol name
         String dialogTitle = fileName;
-        if (isPng) {
+        if (isImage) {
             try {
                 ImageData imageData = ImageData.fromPath(file);
                 Image image = Image.builder()
@@ -330,7 +330,7 @@ public class FileManagerView implements Element {
         Element viewerElement = new Element() {
             @Override
             public void render(Frame frame, Rect renderArea, RenderContext ctx) {
-                if (isPng) {
+                if (isImage) {
                     renderImage(frame, renderArea, file);
                 } else {
                     renderTextFile(frame, renderArea, file);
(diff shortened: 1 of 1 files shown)


maxandersen:
missed gradle dep - now it builds but something broken with image display.

maxandersen:
example of the issue - ignore the horrible coloring in xterm - those are my fault :) 



those doing pixel rendering fails... i reckon it might have to do with how blocks are rendered and thus don't get fully rendered?

maxandersen:
ok so the #64 fixed the rendering within the popup here so I rebased this on main. 

There are a few glitches left though:

1) when pressing Esc in the viewer the image is not cleared. 


2) when running with java 8 (jbang --java 8 filemanager-demo) then things work...but:
ESC stops working - loading images are sluggish and I got this funky error (haven't been able to replicate it - most other cases things just seem to be stuck)


 With respect to java 8 - that commit is separate and I can remove it again; the main intent was to just have one demo that can incorporate as much of tamboui as possible and have it as a litmus test to use for older JVM running. 

I'm inclined we merge this as I think the above are not bugs in this demo but just revealing bugs we need to fix anyway.

wdyt?
yes/no to use filemanager as java 8 to <20  litmus test?
yes to merge?

demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerView.java
```diff
@@ -0,0 +1,456 @@
...
+            } catch (IOException e) {
+                // If we can't load the image, just use the filename
+            }
+        }
+
+        // Create a custom element for the viewer
+        Element viewerElement = new Element() {
+            @Override
+            public void render(Frame frame, Rect renderArea, RenderContext ctx) {
+                if (isPng) {
```
melix: This should handle other file formats (e.g jpg). I was trying the demo and got surprised that it showed me my JPG as text instead of image.

melix (commented):
I confirm that:

- the demo works fine with Java 21 (including ESC)
- it doesn't work with Java 8 (ESC fails closing)

I'm not seeing any slowness with Java 8 though. I have added a note because currently the demo only handles PNG files.

While doing this I figured out having a filter to the file view would be helpful.

I'm fine if we move that demo to Java 8, but please, let's keep a single one. We have to remember that Java 8 is EOL, I don't want to spend too much time debugging things that are outdated and make things too verbose.

maxandersen:
Yes. I don't want java 8 to be spreading. Nor having artificial demos that's annoying to maintain. File manager was my best guess of something simple but broad enough to use as test bed. 

Any idea why java 8 wouldn't close esc ?

demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerView.java
maxandersen: Which are supported ? I assume we just dump base64 encoded to the stream so Is it up to the terminal what it supports ?

maxandersen: added support for jpg/png

maxandersen:
merging as fixed issues and we can iterate on rest.

melix:
> Any idea why java 8 wouldn't close esc ?

No. I suspect this could be an incompatibility with JLine, but at this stage I have no idea.

Issue #61, opened 2026-01-10
image bugs


maxandersen:
run the image demo, scale/zoom/resize and you get stuff like this eventually:

iterm2:

<img width="1510" height="895" alt="Image" src="https://github.com/user-attachments/assets/990f7cb5-6b91-4002-afb2-ad407ed3ab60" />

ghostty:

<img width="1343" height="860" alt="Image" src="https://github.com/user-attachments/assets/97ca2f40-d981-42e9-ad19-01c4c3faedfd" />

melix:
mmm, I'm not seeing this behavior... I suspect it's a platform dependent bug.

maxandersen:
I can no longer reproduce this after #64 merged.

PR #64, opened 2026-01-10
Image fixes

melix:
This PR fixes a couple issues:

1. inconsistent scaling, which was exercised in the demo by the fact that the scaling methods was working differently based on the protocol. This is now consistently using the rescale method of `Image`
2. most important part, the rendering of images was done in the demo, instead of the widget. This was done because there was no way to have access to the raw output in a widget. Since this capability is very special, the PR exposes a `RawOutputCapable` interface which has to be opted-in, and a `FrameInternal` utility class, so that the raw output is not exposed to public API of `Frame` and `Widget`.

### DIFF
--- a/tamboui-image/demos/image-demo/src/main/java/dev/tamboui/demo/ImageDemo.java
+++ b/tamboui-image/demos/image-demo/src/main/java/dev/tamboui/demo/ImageDemo.java
@@ -70,7 +70,6 @@ public class ImageDemo {
     private final String imageTitle;
     private ImageProtocol currentProtocol;
     private ImageScaling currentScaling = ImageScaling.FIT;
-    private boolean needsFullRedraw;
     private boolean forceProtocol;
 
     public static void main(String[] args) throws Exception {
@@ -100,23 +99,8 @@ public class ImageDemo {
 
             var terminal = new Terminal<>(backend);
 
-            // Handle resize
-            backend.onResize(() -> {
-                try {
-                    terminal.draw(this::ui);
-                } catch (IOException e) {
-                    // Ignore
-                }
-            });
-
             // Event loop
             while (running) {
-                // Force full redraw if needed (e.g., after protocol switch)
-                if (needsFullRedraw) {
-                    terminal.clear();
-                    needsFullRedraw = false;
-                }
-
                 terminal.draw(this::ui);
 
                 var c = backend.read(100);
@@ -175,15 +159,13 @@ public class ImageDemo {
                 break;
             case 6: // Ctrl+F - force protocol
                 forceProtocol = true;
-                needsFullRedraw = true;
                 break;
             default:
                 break;
         }
 
-        // Force full redraw and reset force flag when switching protocols
+        // Reset force flag when switching protocols
         if (previousProtocol != currentProtocol) {
-            needsFullRedraw = true;
             forceProtocol = false;
         }
     }


tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
```diff
@@ -79,12 +83,19 @@ public int height() {
...
+     * <p>
+     * If the widget implements {@link RawOutputCapable}, the raw output stream
+     * will be passed to enable native terminal protocol rendering.
      *
      * @param widget the widget to render
      * @param area the area to render within
      */
     public void renderWidget(Widget widget, Rect area) {
-        widget.render(area, buffer);
+        if (widget instanceof RawOutputCapable) {
```
davsclaus, line 94: is this code compiled for older Java as you can do that local variable with instanceof so you dont need type cast in the following code line

davsclaus:
Yeah I played a bit more with that demo last night, and when resizing then you can see the raw ascii codes, and my terminal was misbehavior when I exited the run-demo app.

tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
melix: Yes, to my great despair core/widget libraries are stuck with Java 8.

melix:
I think the misbehavior could be related to using unsupported protocols. I have added an explicit force mode in that PR, aka "use at your own risk"

tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
maxandersen, line 94: took me a bit to grok why the need of additonal interface and it does feel like bit of a codesmell to be honest but I also can't think of a better alternative atm.

maxandersen:
> Yeah I played a bit more with that demo last night, and when resizing then you can see the raw ascii codes, and my terminal was misbehavior when I exited the run-demo app.

sounds like #61

maxandersen:
ok this is making resizing worse and better at same time :) let me record video.

maxandersen:
video got too big :)

when i resize on ghostty it scrambles with asciiprinting quite fast and I hit this:

```
Uqx8dF6SuxDtV8sHERBzV+48KxKOVYa6ZVB4mEnkYJRVMUFLREqhVkLE2RwFWqeMI1AKI2+1pB9kOzLcGQ3UfkYpl1JgWAf8LaXdv04E9YBwigK0D0gLAPyQDGweRA0kUclUlzvnmT370t/xdiINsI6T+JKLBezyHUIR3p457WiItAH+RBNJycGrH70W6SuKg5+RlqjYxcp/4vv6NX0yf/8Jn7TzICvnWP3Qu707sIt5oAp5DEkHZCcwL/CNRePHFF7zM2adCUh+OXMvzk0ge30AKRwHyPJBv0kw68BYkdUGVAWnTqrfuB9iNulHcqv9q/Hn9s1EycRAXoDbPm8qOfEv6x7tRD2U+VAe6p/qh3eQEMt4pQXwOoNVmyKsWCTQeException in thread "SIGWINCH handler" Exception in thread "SIGWINCH handler" java.lang.IndexOutOfBoundsException
        at java.base/java.io.FileOutputStream.writeBytes(Native Method)
        at java.base/java.io.FileOutputStream.write(FileOutputStream.java:352)
        at org.jline.utils.FastBufferedOutputStream.flushBuffer(FastBufferedOutputStream.java:51)
        at org.jline.utils.FastBufferedOutputStream.write(FastBufferedOutputStream.java:43)
        at java.base/java.io.FilterOutputStream.write(FilterOutputStream.java:110)
        at dev.tamboui.backend.jline.JLineBackend.writeRaw(JLineBackend.java:180)
        at dev.tamboui.terminal.Terminal$1.write(Terminal.java:58)
        at java.base/java.io.OutputStream.write(OutputStream.java:124)
        at dev.tamboui.image.protocol.KittyProtocol.sendChunked(KittyProtocol.java:127)
        at dev.tamboui.image.protocol.KittyProtocol.render(KittyProtocol.java:65)
        at dev.tamboui.image.Image.render(Image.java:144)
        at dev.tamboui.terminal.Frame.renderWidget(Frame.java:95)
        at dev.tamboui.demo.ImageDemo.renderImage(ImageDemo.java:307)
        at dev.tamboui.demo.ImageDemo.ui(ImageDemo.java:206)
        at dev.tamboui.terminal.Terminal.draw(Terminal.java:114)
        at dev.tamboui.demo.ImageDemo.lambda$run$0(ImageDemo.java:107)
        at dev.tamboui.backend.jline.JLineBackend.lambda$onResize$0(JLineBackend.java:190)
        at org.jline.terminal.impl.AbstractTerminal.raise(AbstractTerminal.java:92)
        at org.jline.terminal.impl.PosixSysTerminal.lambda$handle$1(PosixSysTerminal.java:68)
        at org.jline.utils.Signals.lambda$register$1(Signals.java:52)
        at jdk.proxy1/jdk.proxy1.$Proxy0.handle(Unknown Source)
        at jdk.unsupported/sun.misc.Signal$InternalMiscHandler.handle(Signal.java:198)
        at java.base/jdk.internal.misc.Signal$1.run(Signal.java:218)
        at java.base/java.lang.Thread.run(Thread.java:1474)
java.lang.IndexOutOfBoundsException
        at java.base/java.io.FileOutputStream.writeBytes(Native Method)
        at java.base/java.io.FileOutputStream.write(FileOutputStream.java:352)
        at org.jline.utils.FastBufferedOutputStream.flushBuffer(FastBufferedOutputStream.java:51)
        at org.jline.utils.FastBufferedOutputStream.write(FastBufferedOutputStream.java:43)
        ```

maxandersen:
resizing in iterm2 kinda works better - but now I get bells on every resize + it starts detecting half-block as best 

https://github.com/user-attachments/assets/6a07b060-8527-49b1-849f-a86f0a228ca0

maxandersen:
fwiw rio does similar to iterm. bells and scrambled content on resize

melix:
mmm, ImageIO uses a file cache, maybe that's the reason, let me see...

tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
melix: It's only a "code smell" because I want it explicitly opt-in. If we exposed raw output to all widgets, that wouldn't be needed. But I think it's dangerous to expose.

maxandersen:
fwiw using timg.sh on video (to trigger redraws) does cause some visual render glitches but nothing like what seeing here. no visual bells and no dump of ascii.

so i assume there is something "non-guarded" in our tamboui image widget


https://github.com/user-attachments/assets/630eb07c-283f-497b-b85d-ce3c13de5b6e

melix:
@maxandersen can you check again?

maxandersen:
> @maxandersen can you check again?

a 1000% better - or "as good as timg works".

...but why does less redraw fix it?

melix:
> ...but why does less redraw fix it?

It was a leftover from a previous implementation: basically before, you could have unsupported protocols not clearing anything, which required us to clear the buffer, but now this is redundant. The other problem was that redraw should never be called from the sigwinch thread.

maxandersen:
> It was a leftover from a previous implementation: basically before, you could have unsupported protocols not clearing anything, which required us to clear the buffer, but now this is redundant. The other problem was that redraw should never be called from the sigwinch thread.

awesome - I can see the first part in this pr (more consistent clearing / redraws) - but missing how redraws were happening in sigwinch/signal handler thread?  just sideeffect of not having demo do drawing itself and let widget do it all?

PR #66, opened 2026-01-10
First iteration on documentation

melix:
This commit adds documentation to TamboUI. It makes use of Asciidoctor with a custom CSS for a more polished look. At the moment, the documentation DOES NOT include screenshots or animations. We will probably want to generate these automatically and publish them alongside the docs.

The documentation can be generated by calling:

    ./gradlew asciidoc

Then open the `docs/build/docs/index.html` file.

Preview:

[docs.tar.gz](https://github.com/user-attachments/files/24545555/docs.tar.gz)

@maxandersen at the moment this doesn't deal with docs publishing. I know you have ideas and that you mentioned that SVG couldn't be rendered on GitHub Pages. Is that true? Sounds strange. But anyway we should have a way to automate screenshots.

### DIFF
--- a/docs/src/docs/asciidoc/developer-guide.adoc
+++ b/docs/src/docs/asciidoc/developer-guide.adoc
@@ -1,0 +1,243 @@
+= Developer Guide
+:doctype: book
+:icons: font
+:source-highlighter: highlight.js
+:toc: left
+:toclevels: 3
+
+include::_attributes.adoc[]
+
+This guide covers how to create custom widgets and components for {project-name}.
+
+== Creating Custom Widgets
+
+{project-name} provides two widget interfaces depending on whether your widget needs external state.
+
+=== Stateless Widgets
+
+For widgets that render the same output given the same inputs:
+
+[source,java]
+----
+public interface Widget {
+    void render(Rect area, Buffer buffer);
+}
+----
+
+Example: A custom separator widget:
+
+[source,java]
+----
+public class Separator implements Widget {
+    private final Style style;
+
+    public Separator(Style style) {
+        this.style = style;
+    }
+
+    @Override
+    public void render(Rect area, Buffer buffer) {
+        // Fill the entire width with a horizontal line
+        for (int x = area.x(); x < area.right(); x++) {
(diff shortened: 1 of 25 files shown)


maxandersen:
I've been working on getting VHS support to asciidotor diagrams. 

https://github.com/asciidoctor/asciidoctor-diagram/pull/502

Crossing fingers for it being released next week but otherwise I'll setup gem using my branch.

maxandersen:
About SVG - it's true. They did render it some years back but then someone found ways to execute random code in svgs so the stopped allowing it on GitHub site. 

You can of course on GitHub pages.

melix:
Ah, but we're only interested in GH pages I guess?

maxandersen:
> Ah, but we're only interested in GH pages I guess?

it would be nice if readme's had some - but yeah for that it need be .gif/.png

docs/src/docs/asciidoc/developer-guide.adoc
```diff
@@ -0,0 +1,243 @@
...
+        .rounded()
+        .borderColor(borderColor)
+        .fill();
+    }
+}
+----
+
+=== Using Components
+
+Components require an ID:
```
maxandersen, line 189: Isn't it more "Components require an ID if you want to be able to style them with CSS and/or handle keyevents" ?

melix: It is, but that's the same. The `Component` class is for this and makes sure you set an id.

maxandersen:
looks good - I'm wondering if we should put it in antora styled dir so overtime can handle versions better but lets get these docs in as is and we can iterate from there.

maxandersen:
merging so can try include it on tamboui.dev now.

melix:
> looks good - I'm wondering if we should put it in antora styled dir so overtime can handle versions better but lets get these docs in as is and we can iterate from there.

The thing which bothers me with Antora is that this means bringing in another toolchain (npm).

maxandersen:
Yeah. I agree. Though the Gradle plugin does greatly simplify it so at least easy to use.

PR #67, opened 2026-01-11
fix: make all .tape demos work + add build validation

maxandersen:
This started as just adding image-demo.tape but then I realized https://github.com/tamboui/tamboui.dev/actions been failing for a few days, cause: jbang aliases been added, but they didn't build as no //DEPS in them. end result: gallary of svg's are not being updated.

note: reason I don't just want to add a global config for jbang to be aware of the deps is a) not all demos need all the dependencies b) by having it embedded in the demos you can run or even install these demos directly.

So I fixed all the examples AND added steps to github action that will:

- refresh jbang catalog (should be a noop)
- then build all examples and fail if any erros
- fail if the refresh changed jbang-catalog.json

~~Ultimately this should be available as a gradle task but preferred having this in now rather than waiting for me to figure out gradle fun to fix it :)~~

its done as part of updateJBangCatalog now with optional -Pjbang.verifyBuilds=true flag

below is copilot summary which explains things well imo:

**Workflow and Build Improvements:**

* The GitHub Actions workflow (`.github/workflows/gradle.yml`) now automatically updates the `jbang-catalog.json` file after building and verifies that the catalog is up to date, failing the workflow if it is not. It also runs a build verification step for all catalog entries when requested.
* The Gradle `UpdateJBangCatalogTask` has been extended to support a `verifyBuilds` property, which, when enabled, runs `jbang build` for each catalog alias to ensure all demos build successfully. This is wired into the workflow and build script. [[1]](diffhunk://#diff-4a8b3838c7db88ddd554d847604fa6dd4b5b527386ec20a5d0346fb991842575R11-R26) [[2]](diffhunk://#diff-4a8b3838c7db88ddd554d847604fa6dd4b5b527386ec20a5d0346fb991842575R66-R74) [[3]](diffhunk://#diff-4a8b3838c7db88ddd554d847604fa6dd4b5b527386ec20a5d0346fb991842575R96-R100) [[4]](diffhunk://#diff-4a8b3838c7db88ddd554d847604fa6dd4b5b527386ec20a5d0346fb991842575R215-R281) [[5]](diffhunk://#diff-490ab1b8f49525e837c7920813dfa3f91e3faf8f066b12bf43d82c2cf3944d3dR44)

**Demo Catalog and Entry Point Updates:**

* The `jbang-catalog.json` file has been updated to add new demos: `inline-progress-demo`, `modular-demo`, and `image-demo`, making them available for discovery and execution via JBang. [[1]](diffhunk://#diff-2afd8b2d140cbc69c6f9b16119375480e5bc19d582e7f0c2e9739ce829901124R10-R12) [[2]](diffhunk://#diff-2afd8b2d140cbc69c6f9b16119375480e5bc19d582e7f0c2e9739ce829901124R22-R24)
* Several demo entry point files now explicitly declare their dependencies using `//DEPS` comments, improving reproducibility and clarity for users running the demos directly. [[1]](diffhunk://#diff-5ac9e40234ef062285b247d6e84767bbbda12077f6d83dc0487b45335c044e66R1-R2) [[2]](diffhunk://#diff-e6b3f2f20d8fba798f53d22d1caec10dd39b6c637bf80f12b1a2c2d77b50ecb1R1-R2) [[3]](diffhunk://#diff-f9b6b348aef4e81df4721bb6d4b0755aa71b3fae81c494d33e13d7de2d2d31feR4) [[4]](diffhunk://#diff-2c8174e8b5b10ce6f546957e5a4d94fa29b6a7e0c79d51c42d425666f8a81129R1-R4) [[5]](diffhunk://#diff-933b22ac4db368ccd41626c683afb04312641a2c85edf45726ee19950b57486eR1-R3) [[6]](diffhunk://#diff-d50beeb14d46478810362dd1a0d1ca2207861f0797924e03fb35d316b7664f0dR1-R4) [[7]](diffhunk://#diff-79e553fc0e991a24cc91fb3eded23554baa44cf151983cd22340e0e62d3ffc7fR1-R2)

**Documentation Enhancements:**

* The `AGENTS.md` documentation has been updated to provide guidelines for creating and maintaining demo `.tape` files, ensuring consistency and clarity in demo video generation.

### DIFF
--- a/buildSrc/src/main/java/dev/tamboui/build/UpdateJBangCatalogTask.java
+++ b/buildSrc/src/main/java/dev/tamboui/build/UpdateJBangCatalogTask.java
@@ -12,15 +12,17 @@ import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputDirectory;
 import org.gradle.api.tasks.Optional;
+import org.gradle.api.tasks.options.Option;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.PathSensitivity;
 import org.gradle.api.tasks.TaskAction;
+import org.gradle.process.ExecOperations;
+import org.gradle.process.ExecResult;
 
-import java.io.BufferedReader;
+import javax.inject.Inject;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -37,6 +39,14 @@ import java.util.stream.Stream;
 public abstract class UpdateJBangCatalogTask extends DefaultTask {
 
     private static final String DEMO_SELECTOR = "demo-selector";
+
+    /**
+     * Returns the ExecOperations service for executing external processes.
+     *
+     * @return the ExecOperations service
+     */
+    @Inject
+    protected abstract ExecOperations getExecOperations();
 
     /**
      * Returns the root project directory.
@@ -68,7 +78,8 @@ public abstract class UpdateJBangCatalogTask extends DefaultTask {
      *
      * @return the verify builds property
      */
-    @Input
+    @Option(option = "verify-builds", description = "Verify builds after writing the catalog")
+    @Input 
     @Optional
     public abstract Property<Boolean> getVerifyBuilds();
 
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/dev/tamboui/build/UpdateJBangCatalogTask.java
```diff
@@ -193,4 +212,71 @@ private void writeCatalog(Map<String, String> aliases) {
...
+
+    /**
+     * Builds a single alias using jbang build command.
+     *
+     * @param aliasName the alias name to build
+     * @return true if the build succeeded (exit code 0), false otherwise
+     */
+    private boolean buildAlias(String aliasName) {
+        try {
+            ProcessBuilder processBuilder = new ProcessBuilder(
```
melix: Should use the `ExecOperations` instead.

buildSrc/src/main/kotlin/dev.tamboui.parent.gradle.kts
```diff
@@ -41,4 +41,5 @@ tasks.register<UpdateJBangCatalogTask>("updateJBangCatalog") {
     projectDir = layout.projectDirectory
     modules =  rootProject.subprojects.map { it.name }
     catalogFile = layout.projectDirectory.file("jbang-catalog.json")
+    verifyBuilds = providers.gradleProperty("jbang.verifyBuilds").map { it.toBoolean() }.orElse(false)
```
melix: You could use an `@Option` on the task instead, which would make the property available via a command line flag (e.g `./gradlew updateJBangCatalog --verify-builds`

maxandersen:
First - Having backend or at least buffer be able to save would be useful no matter what :)

ATM I expect the .tape to move inside or even embedded in the adoc files and have generation run from there. I have on my to-do to try that with asciidotor diagrams. 

Second - Not sure how to automate further as you'll not be able to figure out automatically what a demo can do / should show ?

Third - One thing I do miss in vhs is to make assertions to have it spot errors. That part isn't well handled in vhs.

buildSrc/src/main/java/dev/tamboui/build/UpdateJBangCatalogTask.java
maxandersen: fixed

buildSrc/src/main/kotlin/dev.tamboui.parent.gradle.kts
maxandersen: done

maxandersen:
i fixed the suggestions - .tape/alternative conversation we can continue here or in other dedicated issue/chat.

PR #70, opened 2026-01-11
feat: add pilot style test fixture

maxandersen:
This commit add a textual inspired pilot testing API.

lets do things like:

```
// Test pressing the B key
            pilot.press('b');
            pilot.pause();
            assertEquals(RgbAppExample.BackgroundColor.BLUE, app.getCurrentColor());

            // Test pressing the X key (no binding, so no change)
            pilot.press('x');
            pilot.pause();
            assertEquals(RgbAppExample.BackgroundColor.BLUE, app.getCurrentColor());

			pilot.click(3,4);

            // Quit
            pilot.press('q');
            pilot.pause();
```

idea  is you start your runner and then the pilot becomes available and drives the event loop - it is NOT simulating/emulating at the level of the OS - its at the tamboui fwk level making it not require any terminal etc. nor does it let you test OS integration - but it does let you test/simulate that your app is wired up properly to your TUI.

wdyt?

some of the code is still messy and dumb - but pushing it to get a set of eyes/feedback on the idea.

specifically on if testfixtures is the right place for this.

### DIFF
--- a/tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/app/ToolkitPilot.java
+++ b/tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/app/ToolkitPilot.java
@@ -53,7 +53,7 @@ public final class ToolkitPilot implements Pilot {
     @Override
     public void press(KeyCode keyCode, KeyModifiers modifiers) {
         dev.tamboui.tui.event.KeyEvent event = dev.tamboui.tui.event.KeyEvent.ofKey(keyCode, modifiers);
-        injectEvent(event);
+        dispatch(event);
     }
 
     @Override
@@ -64,7 +64,7 @@ public final class ToolkitPilot implements Pilot {
     @Override
     public void press(char c, KeyModifiers modifiers) {
         dev.tamboui.tui.event.KeyEvent event = dev.tamboui.tui.event.KeyEvent.ofChar(c, modifiers);
-        injectEvent(event);
+        dispatch(event);
     }
 
     @Override
@@ -91,13 +91,13 @@ public final class ToolkitPilot implements Pilot {
     @Override
     public void mousePress(MouseButton button, int x, int y) {
         dev.tamboui.tui.event.MouseEvent event = dev.tamboui.tui.event.MouseEvent.press(button, x, y);
-        injectEvent(event);
+        dispatch(event);
     }
 
     @Override
     public void mouseRelease(MouseButton button, int x, int y) {
         dev.tamboui.tui.event.MouseEvent event = dev.tamboui.tui.event.MouseEvent.release(button, x, y);
-        injectEvent(event);
+        dispatch(event);
     }
 
     @Override
@@ -122,13 +122,13 @@ public final class ToolkitPilot implements Pilot {
     @Override
     public void mouseMove(int x, int y) {
         dev.tamboui.tui.event.MouseEvent event = dev.tamboui.tui.event.MouseEvent.move(x, y);
-        injectEvent(event);
+        dispatch(event);
     }
 
     @Override
     public void resize(int width, int height) {
         dev.tamboui.tui.event.ResizeEvent event = dev.tamboui.tui.event.ResizeEvent.of(width, height);
-        injectEvent(event);
+        dispatch(event);
     }
 
     @Override
@@ -154,8 +154,8 @@ public final class ToolkitPilot implements Pilot {
         }
     }
 
-    private void injectEvent(dev.tamboui.tui.event.Event event) {
-        tuiRunner.injectEvent(event);
+    private void dispatch(dev.tamboui.tui.event.Event event) {
+        tuiRunner.dispatch(event);
         pause();
     }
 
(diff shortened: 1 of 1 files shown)


tamboui-toolkit/src/main/java/dev/tamboui/toolkit/focus/FocusManager.java
```diff
@@ -80,6 +97,7 @@ public void registerFocusable(String elementId, Rect area) {
     public void clearFocusables() {
         focusOrder.clear();
         focusableAreas.clear();
+        allElementAreas.clear();
```
melix, line 100: The name of the method becomes confusing with this, because you're also clearing "all elements"

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/focus/FocusManager.java
```diff
@@ -176,4 +194,17 @@ public Rect focusedArea() {
...
+     * Returns a map of all registered focusable element IDs to their areas.
+     * <p>
+     * This method is useful for testing and widget selection, allowing
+     * test code to find elements by ID and interact with them.
+     *
+     * @return an unmodifiable map of element IDs to their areas
+     */
+    public Map<String, Rect> focusableAreas() {
+        // Return all elements with IDs (not just focusable ones) for widget selection
+        return Collections.unmodifiableMap(new LinkedHashMap<>(allElementAreas));
```
melix, line 208: :x: This doesn't do what the javadocs say, nor the method name say. Also you probably don't need to create a copy since you're wrapping in `unmodifiableMap`.

tamboui-tui/src/testFixtures/java/dev/tamboui/tui/pilot/Pilot.java
```diff
@@ -0,0 +1,266 @@
...
+ * <pre>{@code
+ * try (TestRunner test = TestRunner.runTest(...)) {
+ *     Pilot pilot = test.pilot();
+ *     pilot.press('q');
+ *     pilot.click(10, 5);
+ *     pilot.resize(100, 30);
+ * }
+ * }</pre>
+ */
+public interface Pilot {
```
melix, line 30: Should probably live under testFixtures directory (e.g apply the java-test-fixtures plugin to the build script similar to what is done in core)

tamboui-tui/src/testFixtures/java/dev/tamboui/tui/pilot/TestBackend.java
```diff
@@ -0,0 +1,193 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A headless backend implementation for testing TUI applications.
+ * <p>
+ * This backend doesn't interact with a real terminal, making it suitable
+ * for automated testing. It maintains a configurable terminal size and
+ * provides a queue for injecting events programmatically.
+ */
+public final class TestBackend implements Backend {
```
melix, line 24: Should probably live under testFixtures directory (e.g apply the java-test-fixtures plugin to the build script similar to what is done in core)

tamboui-tui/src/testFixtures/java/dev/tamboui/tui/pilot/TestRunner.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * <pre>{@code
+ * try (TestRunner test = TestRunner.runTest(...)) {
+ *     Pilot pilot = test.pilot();
+ *     pilot.press('q');
+ * }
+ * }</pre>
+ */
+@SuppressWarnings("try")
+public interface TestRunner extends AutoCloseable {
```
melix, line 21: Should probably live under testFixtures directory (e.g apply the java-test-fixtures plugin to the build script similar to what is done in core)

tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
```diff
@@ -255,6 +288,17 @@ public Backend backend() {
         return backend;
     }
 
+    /**
+     * Injects an event into the event queue primarily for testing purposes
+     *
+     * @param event the event to inject
+     */
+    public void injectEvent(Event event) {
```
melix: This could be useful beyond testing I guess. The name `injectEvent` is a bit strange. Probably something like `pushEvent`, `sendEvent` or `acceptEvent` would make more sense to me.

melix (commented):
I like the idea. I think we could add more test fixtures.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/focus/FocusManager.java
maxandersen: yeah so i wasn't sure whats right/wrong here.

The issue is that I had to find a place to map id to elements and FocusManager have that info - but only for  focusable elements.

so as a quick'n'dirty added the other elements to it too. For now mainly used by testing to lookup elements by ID - but I could imagine it being useful elsewhere.

Should we repurpose FocusManager and expand its role or make a new thing that keep track of these?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/focus/FocusManager.java
maxandersen: yeah this really is not needed to happen - there really should just be a findElement() 

how does the css system deal with this - just realizing it probably need similar lookups?

maxandersen: actually - this is just broken. getting a Rect for id is not useful when really after the element...

maxandersen:
replaced with #189

PR #72, opened 2026-01-11
fix: Ctrl+C handling to respect bindings instead of always generating SIGINT

melix:
I got confused, because it seemed that apps were exiting with CTRL+C even if we didn't bind it. Two related bugs were fixed:

1. JLineBackend: Disable ISIG flag after entering raw mode so Ctrl+C goes through the event system instead of generating SIGINT. This allows quit behavior to be controlled by bindings.

2. TextAreaElement/Toolkit: Don't consume CHAR events with Ctrl or Alt modifiers. These are control sequences, not text input. This allows Ctrl+C (and other Ctrl combinations) to bubble up and be matched against bindings.


melix:
IMO handling this is trivial. In addition, since it's dependent on the backend, we should keep it as is.

PR #73, opened 2026-01-12
feat: Introduce automatic asciinema demos

melix:
This commit adds a backend which can export a demo as asciinema animation. For this, it uses the docs/video/*.tape files as the source for keys/mouse events. A Gradle task is responsible for collecting the demo metadata and generating an asciidoc page which lists all the demos and shows them using the Asciinema player. This is more accurate than SVGs, while being more compact and diff friendly.

If a demo misses a `.tape`, then the build will fail, so that we make sure that all demos are properly wired.

To see the result, run: `./gradlew asciidoc`, then serve the documentation using a local http server, e.g:

```
cd docs/build/docs
python -m http.server 8080
```

(because it needs to load files from the server in Asciinema player).

@maxandersen please do NOT merge this until it is approved by @graemerocher , that's a legal requirement for contributions made during work hours.

### DIFF
--- a/tamboui-image/demos/image-demo/src/main/java/dev/tamboui/demo/ImageDemo.java
+++ b/tamboui-image/demos/image-demo/src/main/java/dev/tamboui/demo/ImageDemo.java
@@ -22,6 +22,7 @@ import dev.tamboui.image.protocol.ImageProtocol;
 import dev.tamboui.image.protocol.ITermProtocol;
 import dev.tamboui.image.protocol.KittyProtocol;
 import dev.tamboui.image.protocol.SixelProtocol;
+import dev.tamboui.internal.record.RecordingBackend;
 import dev.tamboui.layout.Constraint;
 import dev.tamboui.layout.Layout;
 import dev.tamboui.layout.Rect;
@@ -43,6 +44,7 @@ import dev.tamboui.widgets.paragraph.Paragraph;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.EnumSet;
 
 /**
  * Demo TUI application showcasing image rendering capabilities.
@@ -81,7 +83,16 @@ public class ImageDemo {
     }
 
     public ImageDemo(String customImagePath) throws IOException {
-        this.capabilities = TerminalImageCapabilities.detect();
+        try (var backend = BackendFactory.create()) {
+            if (backend instanceof RecordingBackend) {
+                this.capabilities = TerminalImageCapabilities.withSupport(EnumSet.of(
+                        TerminalImageProtocol.HALF_BLOCK,
+                        TerminalImageProtocol.BRAILLE
+                ));
+            } else {
+                this.capabilities = TerminalImageCapabilities.detect();
+            }
+        }
         this.currentProtocol = capabilities.bestProtocol();
 
         // Load image from path or bundled resource


maxandersen:
not a asciinema fan as i moved to vhs for much nicer svg/mp4s....but it intrigues me so definitely taking a look :)

melix:
I spent quite some time trying to automate SVG generation and never got any decent result. The outcome depends on the browser, the files are (much) larger, so mixed feelings about it.

maxandersen:
whats the gradle command that builds this?

maxandersen:
ah gradle assciidoc

maxandersen:
so i of course run it with `jbang httpd@jbangdev` (no need for python :)

anyhow it works; but of course quite annoying have to serve it out...means can't share it easily... 

But ignoring that - looks decent but the rendering is "worse" from my pov in one area - I have lines between all the horizontal characters.



is there a way to avoid that 1-2 pixel on every line - makes it look bad ?

4200 fps is awesome on the rgb demo :)

docs/video/image-demo.tape
```diff
@@ -0,0 +1,11 @@
+Source shared_.tape
+
+# Setup
```
maxandersen, line 3: image-demo doesn't render anything but is autodetected to iterm2 which sounds like a bug ?

the vhs doesn't support either but braille and fallback does show up.

tamboui-core/src/main/java/dev/tamboui/internal/record/RecordingConfig.java
```diff
@@ -0,0 +1,87 @@
...
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.internal.record;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Configuration for SVG recording, loaded from system properties.
```
maxandersen:
```suggestion
 * Configuration for Asciinema (.cast) recording, loaded from system properties.
```

melix:
> is there a way to avoid that 1-2 pixel on every line - makes it look bad ?

Yes, fixed

docs/video/image-demo.tape
melix: I think we'll need to force the detected mode to half-block. This is probably because it gets the terminal capabilities from the running environment, but then asciinema doesn't support that.

melix: fixed

maxandersen:
ok, i'm very closed to being sold on this - the speed of generation and sizes of the output (about 1/3) is a win. 

Other thing I spotted is  that its as if colors or characters aren't rendered fully?

Here is the blocks demo looks in recording vs how it looks in my terminal:



any idea?

melix:
I think that comes from the fact the italic style is applied to the borders and asciinema player supports this. It may actually reveal an issue (maybe we shouldn't apply italic to the border itself)

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -100,7 +101,15 @@ public static Backend create() throws IOException {
             );
         }
 
-        return providers.get(0).create();
+        Backend backend = providers.get(0).create();
```
maxandersen, line 104: this is my only real "code smell" - that its hardly coupled in the backendfactory and the recording mechnism is in the core - but its also super useful :)

maxandersen:
> I think that comes from the fact the italic style is applied to the borders and asciinema player supports this. It may actually reveal an issue (maybe we shouldn't apply italic to the border itself)

true, it should only apply to the text lines - but its also missing the color change somehow.

melix:
Yes I actually used `vhs` to record a .tape file, that's useful.

PR #75, opened 2026-01-12
feat: Add `fit()` to `Panel`

melix:
This commit adds a fit method to Panel, which as the name states, computes the height to fit its content. This is a convenience method that I missed when playing with the API.


maxandersen:
Curious why you needed it? 

Would need go away if we had more complete layout mechanics ?

melix:
Yes, probably. Inside a column, several panels, the top and bottom ones were growing and I didn't want that.

PR #81, opened 2026-01-13
fix: Ordered backend creation

melix:
This commit changes how backend creation works. It is possible to have multiple backends on classpath, in which case, by default, they are tried _in order_. The first one that doesn't fail creation is used.

It is possible for a user to override that behavior by setting the `tamboui.backend` system property or `TAMBOUI_BACKEND` env var, in which case, the user can pick one, or pick many by ordering them (e.g `jline,panama`).

This fixes the docs generation that fail when using the panama backend (because it needs a real terminal).

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
@@ -17,8 +17,11 @@ import java.util.stream.StreamSupport;
  * Factory for creating {@link Backend} instances using the {@link ServiceLoader} mechanism.
  * <p>
  * This factory discovers {@link BackendProvider} implementations on the classpath
- * and uses them to create backend instances. Applications should include exactly
- * one backend provider on the classpath (e.g., tamboui-jline).
+ * and uses them to create backend instances. When multiple providers are available,
+ * they are tried in order until one successfully creates a backend.
+ * <p>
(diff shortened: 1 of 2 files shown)


tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -112,6 +85,73 @@ public static Backend create() throws IOException {
...
+     * @throws IllegalStateException if a specified provider is not found
+     */
+    private static List<BackendProvider> resolveProviders(String providerSpec, List<BackendProvider> allProviders) {
+        List<BackendProvider> resolved = new java.util.ArrayList<>();
+        for (String spec : providerSpec.split(",")) {
+            String trimmedSpec = spec.trim();
+            if (trimmedSpec.isEmpty()) {
+                continue;
+            }
+            boolean isClassName = trimmedSpec.contains(".");
```
maxandersen, line 103: not blocking this pr - but im thinking if we wont load classname unless there is match in the providers is there any reason to provide that option?

graemerocher (approved):

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
melix: I'm not sure we'd use FQCN in any case, we can remove in another PR

PR #82, opened 2026-01-13
fix: Overflow.CLIP didn't clip

melix:
I wasn't sure if it was me misusing the API but after checking, I saw that the CLIP mode actually didn't clip. Actually, it clipped at the terminal width. This caused weird behavior where having a text item inside a paragraph/panel could write _outside_ the panel.

### DIFF
--- a/tamboui-widgets/src/test/java/dev/tamboui/widgets/list/ListWidgetTest.java
+++ b/tamboui-widgets/src/test/java/dev/tamboui/widgets/list/ListWidgetTest.java
@@ -244,11 +244,9 @@ class ListWidgetTest {
         list.render(area, buffer, state);
 
         // Should show "Hello" (5 chars)
-        assertThat(buffer.get(0, 0).symbol()).isEqualTo("H");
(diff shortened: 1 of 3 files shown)


tamboui-widgets/src/test/java/dev/tamboui/widgets/list/ListWidgetTest.java
```diff
@@ -222,4 +223,159 @@ void withoutRepeatHighlightSymbol() {
...
+            .highlightSymbol("")
+            .build();
+        Rect area = new Rect(0, 0, 5, 1);
+        Buffer buffer = Buffer.empty(area);
+        ListState state = new ListState();
+
+        list.render(area, buffer, state);
+
+        // Should show "Hello" (5 chars)
+        assertThat(buffer.get(0, 0).symbol()).isEqualTo("H");
```
maxandersen: tip: if you use the assertj for buffer this can be much easier read and the errror/fail messages much more informative :)

PR #83, opened 2026-01-13
feat: Automated docs publishing

melix:
This commit sets up the build to publish documentation automatically. It will publish the SNAPSHOT version of docs into the `main` directory, while version specific docs will be published in the corresponding folder.

This will require setting up an SSH key for this to work on CI.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,38 @@
+name: Publish documentation
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+      - '[0-9]*'
+jobs:
+  publish_documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 1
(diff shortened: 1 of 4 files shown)


melix:
Deploy key added

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,38 @@
...
+      - 'v*'
+      - '[0-9]*'
+jobs:
+  publish_documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 1
+          ssh-key: "${{ secrets.DOCS_SSH_PRIVATE_KEY }}"
```
maxandersen, line 16: does this really need to be ssh key and not just upgraded token?

melix: AFAIK we need an ssh key here because we're pushing to a different repo.

maxandersen: seems it must be a finegrained PAT key. so thats about the same amount of work so this is fine.

PR #84, opened 2026-01-13
feat: add hyperlink

maxandersen:
adds osc-8 hyperlinks as a Style. Makes it possible to put links almost anywhere - demos shows paragraph and block titles. even handle wrapping of urls.

pushing this for comments.

Reason I put it on draft is that in my further testing I noticed that Paragraph seem to now apply styles everywhere instead of only where it starts and begins...resulting in LOTS of links being kept in buffer.

This might be sideeffect of changes in here.

note: reason I put it in core is hyperlinks are needed/useful for markdown/up rendering so good to have as first class citizen.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/style/Hyperlink.java
+++ b/tamboui-core/src/main/java/dev/tamboui/style/Hyperlink.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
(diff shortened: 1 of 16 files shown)


melix:
Works fine. Got confused in the first place because the first terminal I tried was Rio, which doesn't support hyperlinks. But gnome terminal and Ghostty work fine :)

maxandersen:
lets get this in - merge if you are ok.

as discussed there is seomthign with background and styling but i think thats complementary to osc8/link handling.

PR #87, opened 2026-01-13
feat: Flex layout with solver

### LINKED ISSUE #26: Flex layout is just an enum, no implementation

melix:
This is an alternative implementation to #31, which uses a proper Cassowary solver. I made the decision to add a solver in our code, rather than use an external library, to remain in the "no external dependency unless strictly required" effort.

The solver implementation algorithm was AI generated. The demo shows the flex layout in action with several constraints. Performance is expected to be very good, because we have a very low number of constraints in a TUI.

Fixes #26

### DIFF
--- a/demos/flex-demo/src/main/java/dev/tamboui/demo/flex/FlexDemo.java
+++ b/demos/flex-demo/src/main/java/dev/tamboui/demo/flex/FlexDemo.java
@@ -32,7 +32,7 @@ public class FlexDemo {
 
     private static final Flex[] FLEX_MODES = {
             Flex.START, Flex.CENTER, Flex.END,
-            Flex.SPACE_BETWEEN, Flex.SPACE_AROUND
+            Flex.SPACE_BETWEEN, Flex.SPACE_AROUND, Flex.SPACE_EVENLY
     };
 
     private static final String[] EXAMPLE_NAMES = {
--- a/demos/rflex-demo/build.gradle.kts
+++ b/demos/rflex-demo/build.gradle.kts
@@ -1,0 +1,19 @@
+plugins {
+    id("dev.tamboui.demo-project")
+}
+
+description = "Demo showcasing Flex layout modes (Ratatui's Flex demo)"
+
+dependencies {
+    implementation(projects.tambouiToolkit)
+    runtimeOnly(projects.tambouiJline)
+}
+
+application {
+    mainClass.set("dev.tamboui.demo.flex.RFlexDemo")
+}
+
+demo {
+    displayName = "Ratatui Flex Layout Demo"
+    tags = setOf("layout", "flex", "constraints", "cassowary")
+}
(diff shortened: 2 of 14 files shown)


graemerocher (approved):

demos/flex-demo/src/main/java/dev/tamboui/demo/flex/FlexDemo.java
```diff
@@ -0,0 +1,365 @@
...
+ *   <li>1-4: Switch between examples</li>
+ *   <li>Left/Right arrows: Cycle through Flex modes</li>
+ *   <li>q: Quit</li>
+ * </ul>
+ */
+public class FlexDemo {
+
+    private static final Flex[] FLEX_MODES = {
+            Flex.START, Flex.CENTER, Flex.END,
+            Flex.SPACE_BETWEEN, Flex.SPACE_AROUND
```
maxandersen: I noticed SPACE_EVENLY exist in ratatui. should we have it?

maxandersen: also - remove Flex.LEGACY (in Flex.java as I think thats just a ratatui left over?)

melix: I think it makes sense

maxandersen:
i've not gone through them all but I pushed a commit with RFlexDemo which is a close port of ratatui's flex demo and here I'm spotting some few glitches.



see how in the lower row Fill(1) is missing.

And if you try the various modes it definitely lays out very differently.

This might be intentional - or a bug/missing feature - but the current demo doesn't test a lot of what this one show so thought would be interestign to at least grok as this layout is going to be painful to alter over time :)

maxandersen:
Note sure if intentional but  Constaint.fill(0) is allowed in ratatui...we have >1 ...

also unrleated but our rendering is not merging background style afaics which is why the labels are on default background but ratatui has a more clam look. I thought I fixed that but its definitely back or not merged :)

melix:
I'm investigating the issues you mentioned, thanks!

melix:
@maxandersen all issues fixed. The demo looks exactly like the ratatui one now:

diff

tamboui-core/src/main/java/dev/tamboui/layout/cassowary/Expression.java
```diff
@@ -0,0 +1,371 @@
...
+ * // Create constraint: 2*x + 3*y + 10 == 0
+ * CassowaryConstraint c = expr.equalTo(0, Strength.REQUIRED);
+ * </pre>
+ */
+public final class Expression {
+
+    private static final Expression ZERO = new Expression(Collections.emptyList(), 0.0);
+
+    private final List<Term> terms;
+    private final double constant;
```
maxandersen, line 39: for now we'll work with double - but i do see a few one-off cell renderings i'm wondering if the use of doubles is the cause and we should look into having a Fraction representation instead.

See "Fraction" in https://www.textualize.io/blog/7-things-ive-learned-building-a-modern-tui-framework/

PR #88, opened 2026-01-14
fix: Rework ListContainer to RichList

melix:
This commit refactors the ListContainer element to support rich styled elements as items. The previous implementation, which lived in toolkit, only supported `ListItem`, which is at the widget level. In practice, this was cumbersome and unintuitive, because it wasn't obvious why you had to limit yourself to a Text element with Style in list.

The new component was renamed to `RichList` to make it clearer, and demos were adjusted. A new "Rich List" demo was added.

In the process I discovered that some elements were wrongly using a private API (casting to DefaultRenderContext) which was a mistake. A note has been added about the use of that API to avoid such mistakes in the future.

### DIFF
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -13,11 +13,11 @@
     "inline-progress-demo": {
       "script-ref": "demos/inline-progress-demo/src/main/java/dev/tamboui/demo/InlineProgressDemo.java"
     },
+    "listelement-demo": {
+      "script-ref": "demos/listelement-demo/src/main/java/dev/tamboui/demo/ListElementDemo.java"
+    },
     "modular-demo": {
       "script-ref": "demos/modular-demo/src/main/java/dev/tamboui/demo/modular/ModularDemo.java"
-    },
-    "rich-list-demo": {
-      "script-ref": "demos/rich-list-demo/src/main/java/dev/tamboui/demo/ListElementDemo.java"
     },
     "css-demo": {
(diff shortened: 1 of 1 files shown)


graemerocher (approved):

maxandersen:
haven't tried running it yet just looked at code for now:

1. Good idea to have ability beyond just List with text/style
2. RichList ....not loving the name...see 3
3. is there a reason why we don't have ListWidget allow listing any Widget? then RichList won't be richer :) but just add the higher level things ListContainer stil works

melix:
Regarding `ListWidget` accepting any widget. That's a real challenge, since a `Widget` doesn't have any intrinsic dimension: it simply renders in a `Rect`. Therefore the container cannot know in advance the size of its components. We should see widgets as "primitives". It works with `Text` because we know its height is 1. The higher level abstraction is the `Element` model. I can rename `RichList` to `ListElement` for consistency if you prefer, though.

maxandersen:
> Regarding `ListWidget` accepting any widget. That's a real challenge, since a `Widget` doesn't have any intrinsic dimension: it simply renders in a `Rect`. Therefore the container cannot know in advance the size of its components. We should see widgets as "primitives". It works with `Text` because we know its height is 1. 

ah of course - makes sense. 

> The higher level abstraction is the `Element` model. I can rename `RichList` to `ListElement` for consistency if you prefer, though.

+1

PR #89, opened 2026-01-14
fix: poll timeout vs tick rate

melix:
This commit makes it so that the _effective_ poll timeout cannot be greater than the tick rate. Without this, the following could happen, say, for example if tick = 25ms and poll = 100ms

1. the user doesn't type anything for 100ms
2. when poll timeout is reached, 4 ticks are sent in a row, causing 4 quick redraws

Now, the behavior is much more predictable:

1. the poll timeout must be <= tick timeout
2. if the user types quickly, < poll timeout, then there's a UI refresh for each key press (responsive UI)
3. if poll timeout is reached, a tick event is sent

In addition, I reduced the default tick/poll to 40ms, which is effectively 25 fps.

Last but not least, all TuiRunner applications gained a hidden overlay, which can be activated by pressing CTRL+SHIFT+F12, which will display the FPS. This can only be done at the toolkit level though.

### DIFF
--- a/tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
+++ b/tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
@@ -10,11 +10,11 @@ import dev.tamboui.terminal.BackendFactory;
 import dev.tamboui.terminal.Frame;
 import dev.tamboui.terminal.Terminal;
 import dev.tamboui.tui.bindings.ActionHandler;
+import dev.tamboui.tui.bindings.Actions;
 import dev.tamboui.tui.bindings.Bindings;
 import dev.tamboui.tui.bindings.BindingSets;
 import dev.tamboui.tui.event.Event;
 import dev.tamboui.tui.event.EventParser;
-import dev.tamboui.tui.event.KeyCode;
 import dev.tamboui.tui.event.KeyEvent;
 import dev.tamboui.tui.event.TickEvent;
 import dev.tamboui.tui.overlay.FpsOverlay;
@@ -211,14 +211,11 @@ public final class TuiRunner implements AutoCloseable {
             Event event = pollEvent(effectivePollTimeout);
 
             if (event != null) {
-                // Handle FPS overlay toggle: CTRL+SHIFT+F12
-                if (event instanceof KeyEvent) {
-                    KeyEvent keyEvent = (KeyEvent) event;
-                    if (keyEvent.code() == KeyCode.F12 && keyEvent.hasCtrl() && keyEvent.hasShift()) {
-                        fpsOverlay.toggle();
-                        terminal.draw(wrappedRenderer::render);
-                        continue;
-                    }
+                // Handle FPS overlay toggle
+                if (config.bindings().matches(event, Actions.TOGGLE_FPS_OVERLAY)) {
+                    fpsOverlay.toggle();
+                    terminal.draw(wrappedRenderer::render);
+                    continue;
                 }
 
                 boolean shouldRedraw = handler.handle(event, this);
--- a/tamboui-tui/src/main/java/dev/tamboui/tui/TuiConfig.java
+++ b/tamboui-tui/src/main/java/dev/tamboui/tui/TuiConfig.java
@@ -175,8 +175,6 @@ public final class TuiConfig {
 
     /**
      * Returns whether the FPS overlay is enabled.
-     * <p>
-     * When enabled, press F12 to toggle overlay visibility.
      *
      * @return true if FPS overlay is enabled
      */
@@ -382,8 +380,6 @@ public final class TuiConfig {
 
         /**
          * Enables or disables the FPS overlay.
-         * <p>
-         * When enabled, press F12 to toggle overlay visibility.
          *
          * @param enabled true to enable the FPS overlay
          * @return this builder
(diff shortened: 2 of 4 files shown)


tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/RenderContext.java
```diff
@@ -173,6 +175,28 @@ default Style childStyle(String childName, ChildPosition position, dev.tamboui.c
         return currentStyle();  // fallback when no CSS engine
     }
 
+    /**
+     * Renders a child element and registers it for event routing.
+     * <p>
+     * Container elements should use this method instead of calling {@code child.render()}
```
maxandersen, line 181: would it make sense to do something like #3, an explicit mount - or is this truly "rendering" step?

tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
```diff
@@ -172,16 +186,35 @@ public static TuiRunner create(TuiConfig config) throws Exception {
...
         // Initial draw
-        terminal.draw(renderer::render);
+        terminal.draw(wrappedRenderer::render);
 
         while (running.get()) {
-            Event event = pollEvent(config.pollTimeout());
+            Event event = pollEvent(effectivePollTimeout);
 
             if (event != null) {
+                // Handle FPS overlay toggle: CTRL+SHIFT+F12
```
maxandersen: it says F12 some places, ctrl+shift+F12 other places.

...I love F'keys but to use them I have to press an additional key so it becomes ctrl+shift+fn+F12....thats quite a mouthfull.

how about ctrl+shift+0 ?

tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
```diff
@@ -172,16 +186,35 @@ public static TuiRunner create(TuiConfig config) throws Exception {
...
+                if (event instanceof KeyEvent) {
+                    KeyEvent keyEvent = (KeyEvent) event;
+                    if (keyEvent.code() == KeyCode.F12 && keyEvent.hasCtrl() && keyEvent.hasShift()) {
```
maxandersen: shouldn't this be mapped via (default) bindings ? so users can remap the key?

tamboui-tui/src/main/java/dev/tamboui/tui/overlay/FpsOverlay.java
```diff
@@ -0,0 +1,186 @@
...
+import dev.tamboui.widgets.block.Block;
+import dev.tamboui.widgets.block.BorderType;
+import dev.tamboui.widgets.block.Borders;
+import dev.tamboui.widgets.block.Title;
+import dev.tamboui.widgets.paragraph.Paragraph;
+
+import java.time.Duration;
+
+/**
+ * FPS overlay that displays performance metrics.
```
maxandersen, line 24: this feels like it almost could be a widget with a state to manual manage when not having toolkit available...but that can be done separately

maxandersen (changes requested):
changing to take a look - the textareademo is not reacting to resizes properly. lots of content dissapears.

tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
melix: Trying hard to avoid shortcuts which are likely to be bound to panel management in Gnome/KDE/...

maxandersen: ah do - anyhow specifically on macbook keyboards it turned out to be 3 keys in corner then f12 so its not that hard there.

worse on my moonlander keyboard with my custom layout that uses an F key about once a year :)

tamboui-tui/src/main/java/dev/tamboui/tui/bindings/BindingSets.java
```diff
@@ -51,16 +76,42 @@ public final class BindingSets {
     private static final Map<String, MouseButton> MOUSE_BUTTONS = createMouseButtonMap();
     private static final Map<String, MouseEventKind> MOUSE_KINDS = createMouseKindMap();
 
-    // Predefined binding sets - loaded from properties files
-    private static final Bindings STANDARD = loadBuiltIn("standard.properties", null);
-    private static final Bindings VIM = loadBuiltIn("vim.properties", STANDARD);
-    private static final Bindings EMACS = loadBuiltIn("emacs.properties", STANDARD);
-    private static final Bindings INTELLIJ = loadBuiltIn("intellij.properties", STANDARD);
-    private static final Bindings VSCODE = loadBuiltIn("vscode.properties", STANDARD);
+    private static final Path USER_CONFIG_DIR = Paths.get(System.getProperty("user.home"), ".tamboui");
```
maxandersen: not a fan we do this as hardcoded but I don't have better suggestion and at least nice to have.

But we should probably be good citizens and use `~/.config/.tamboui`

maxandersen (approved):
I'm not a fan of we default always look for override files and no ability to adjust/isolate applications and kind a wish that was separate PR.

I'm approving the rest. the redraw issue is gone and its still smooth as butter :)

But I would suggest we separate out the override facility in separate PR or at least have a way to turn it off or somehow avoid it programmatically.

melix:
Removed the override, will do in a separate PR

PR #90, opened 2026-01-14
feat: core capability reporting by core jar

maxandersen:
This adds basic diagnostic printing to tamboui-core that when run you get something like:

```
TamboUI capability report

== tamboui-core:properties
java.version: 25.0.1
java.vendor: Eclipse Adoptium
os.name: Mac OS X
os.arch: aarch64
os.version: 26.0
tamboui.backend: null

== tamboui-core:environment
TERM: xterm-256color
COLORTERM: truecolor
TERM_PROGRAM: vscode
TERM_PROGRAM_VERSION: 2.2.44
LC_ALL: null
LANG: da_DK.UTF-8
TAMBOUI_BACKEND: null

== tamboui-core:features
backend.present: true
backend.provider.jline: true
backend.provider.panama: true
backend.count: 2
backend.providers: jline (dev.tamboui.backend.jline.JLineBackendProvider), panama (dev.tamboui.backend.panama.PanamaBackendProvider)

== tamboui-image:features
native.kitty: false
native.iterm2: true
native.sixel: false
native.half_block: true
native.braille: true
native.none: false
native.any: true
bestSupport: ITERM2
supportedProtocols: ITERM2, HALF_BLOCK, BRAILLE
```

it uses service provider to let other modules provide info. i.e. above is from running:

`jbang --deps dev.tamboui:t
amboui-image:LATEST,dev.tamboui:tamboui-jline:LATEST,dev.tamboui:tamboui-panama-backend:
LATEST dev.tamboui:tamboui-core:LATEST`

I'm not sure we want it split up in sections/features/values or if a Map<String,String> suffices...

Idea is though that when we have decent println styling will make the printing look niceer - but for now to add basic feature in of being able to run core jar, or explicit use `dev.tamboui.capability.Capabilities` as main class.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
+++ b/tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
@@ -67,7 +67,7 @@ public final class SafeServiceLoader {
                 consecutiveErrors = 0;
 
                 loaded.add(provider);
-            } catch (ServiceConfigurationError | LinkageError e) {
+            } catch (ServiceConfigurationError | LinkageError  e) {
                 if (onError != null) {
                     onError.accept(e);
                 }


tamboui-core/src/main/java/dev/tamboui/capability/core/CoreCapabilityProvider.java
```diff
@@ -0,0 +1,152 @@
...
+public final class CoreCapabilityProvider implements CapabilityProvider {
+
+    @Override
+    public String source() {
+        return "tamboui-core";
+    }
+
+    @Override
+    public void contribute(CapabilityReportBuilder report) {
+        report.section(source(), "properties")
```
melix:
```suggestion
        report.section(source(), "system properties")
```

The rationale being that without this, it can be confusing for properties like `tamboui.backend`, which are not the effective selected backend, but the system property.

tamboui-core/src/main/java/dev/tamboui/capability/core/CoreCapabilityProvider.java
```diff
@@ -0,0 +1,152 @@
...
+    private static List<BackendProvider> discoverBackends() {
+        ServiceLoader<BackendProvider> loader = ServiceLoader.load(BackendProvider.class);
+        List<BackendProvider> providers = new ArrayList<>();
+        for (BackendProvider provider : loader) {
+            providers.add(provider);
+        }
+        return providers;
+    }
+
+    private static String joinProviderDescriptions(List<BackendProvider> providers, CapabilityReportBuilder.Section out) {
```
melix: Missing Java 21? ;)

tamboui-core/src/main/java/dev/tamboui/capability/core/CoreCapabilityProvider.java
```diff
@@ -0,0 +1,152 @@
...
+            joiner.add(providerName + " (" + providerClass + ")");
+        }
+        return joiner.toString();
+    }
+
+    /**
+     * Best-effort: uses BackendProvider#name() if present (newer versions),
+     * otherwise returns null.
+     */
+    private static String backendProviderName(BackendProvider provider) {
```
melix: :x: This is wrong. I guess LLM assumed we need backwards compatibility but that's not true we haven't released anything. Use of reflection should be forbidden in our AGENTS.md (I thought it was actually mentioned).

tamboui-core/src/main/java/dev/tamboui/capability/core/CoreCapabilityProvider.java
```diff
@@ -0,0 +1,152 @@
...
+            simple = simple.substring(lastDot + 1);
+        }
+        if (simple.endsWith("BackendProvider")) {
+            simple = simple.substring(0, simple.length() - "BackendProvider".length());
+        }
+        return simple.toLowerCase(Locale.ROOT);
+    }
+
+    private static String sanitizeFeatureToken(String s) {
+        String lower = s.toLowerCase(Locale.ROOT);
```
melix: :x: This is system dependent: the rendering will depend on the system locale

tamboui-core/src/main/java/dev/tamboui/capability/CapabilityRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+public final class CapabilityRegistry {
+
+    private final List<CapabilityProvider> providers;
+
+    private CapabilityRegistry(List<CapabilityProvider> providers) {
+        this.providers = Collections.unmodifiableList(providers);
+    }
+
+    public static CapabilityRegistry load() {
+        return load(Thread.currentThread().getContextClassLoader());
```
melix: :?: Why using the context classloader here?

tamboui-image/src/main/java/dev/tamboui/image/capability/ImageCapabilityProvider.java
```diff
@@ -0,0 +1,40 @@
...
+    public void contribute(CapabilityReportBuilder report) {
+        TerminalImageCapabilities caps = TerminalImageCapabilities.detect();
+        Set<TerminalImageProtocol> supported = caps.supportedProtocols();
+
+        CapabilityReportBuilder.Section images = report.section(source(), "features")
+                .kv("bestSupport", caps.bestSupport())
+                .kv("supportedProtocols", supported.stream().map(Enum::name).collect(Collectors.joining(", ")));
+
+        for (TerminalImageProtocol protocol : TerminalImageProtocol.values()) {
+            images.feature("native." + protocol.name().toLowerCase(), caps.supports(protocol));
```
melix: :x: The name `native.` is confusing. Maybe something like `protocol.<NAME>.supported`

tamboui-image/src/main/java/dev/tamboui/image/capability/ImageCapabilityProvider.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+        images.feature("native.any", caps.supportsNativeImages());
```
melix: :x: Same here, we should probably rename the `supportsNativeImages` since it's even more confusing (didn't realize when adding the feature)

tamboui-core/build.gradle.kts
```diff
@@ -5,6 +5,12 @@ plugins {
 
 description = "Core types and abstractions for TamboUI TUI library"
 
+tasks.named<org.gradle.jvm.tasks.Jar>("jar") {
+    manifest {
+        attributes("Main-Class" to "dev.tamboui.capability.Capabilities")
```
melix: mmm, okay but this looks like an arbitrary choice.

tamboui-core/src/main/java/dev/tamboui/capability/CapabilityRegistry.java
maxandersen: thats bogus - dont vibe code at 1 am! :) i'll clean this up.

tamboui-core/build.gradle.kts
maxandersen: arbitrary in what way? want something like dev.tamboui.Main that just delegate to Capabilities?

melix: Yes, that would be better. Maybe just me, but a library with an entry point makes me raise an eyebrow ;)

maxandersen: done

maxandersen:
I cleaned up but i'm wondering if I should boil this down to

Capabilities {
    Map<String, List<String>> env;
    Map<String, List<String>> properties;
    Map<String, Map<String,Object> features;
}

and use dotted notation to separate the sections rather thant the full typesafety now present?

melix:
My gut feeling is that a capability report with untyped (aka stringly typed) capabilities is only useful for printing.

maxandersen:
> My gut feeling is that a capability report with untyped (aka stringly typed) capabilities is only useful for printing.

yes - same here; which is why i was leaning to just have this feature be stringly typed for printing/diagnostics and let each module have its own specific detection which it then can stringify on request.

maxandersen:
ok, this have been updated to simplify to a stringified model but this also made me bump into #108 thus this introduces a SafeServiceProviderLoader that wont fail if a provider can't be loaded and I've updated BackendFactory and the provider lookup in Capabilities to also not fail so it will still print capability info even if a module is requiring other java version or native libs not available or some classes missing from a dependency etc.

maxandersen:
If I hard fail it will fail loading Panama on any Java below Java 22 and not even see jline.

melix:
But why looping? If it fails once you already have the information it fails?

maxandersen:
Its the case imagine you have another provider that fails to load.

melix:
I'm sorry but I really don't see why you would be doing this. Either a backend fails to load and we should try the next one, or it doesn't and we don't have a problem. If a backend fails to load, if it fails once, I see no reason we would try again, it's likely it will fail exactly the same so a waste of time. Unless I'm missing something obvious?

maxandersen:
What retry loop is it you think is bad ? Can you put comment on the code ?

We can definitely lower the count but it's just that we can't really know. 

Ie. Today it's Panama and jline3. Tomorrow it's Panama, jline3 and jline4 maybe some os specific backend. Just stopping and call hasnext on first error would make it fail too early.

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
```diff
@@ -0,0 +1,111 @@
...
+                if (!it.hasNext()) {
+                    break;
+                }
+
+                S provider = it.next(); // may throw
+                consecutiveErrors = 0;
+
+                // De-dup: protects against odd iterator recovery behavior and multiple resources.
+                String impl = provider.getClass().getName();
+                if (seen.add(impl)) {
```
melix: This seems like a hack. I don't understand why this is needed.

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
```diff
@@ -0,0 +1,111 @@
...
+                    loaded.add(provider);
+                }
+
+            } catch (ServiceConfigurationError e) {
+                if (onError != null) {
+                    onError.accept(e);
+                }
+                if (++consecutiveErrors > MAX_CONSECUTIVE_ERRORS) {
```
melix: Why is this needed? Why do you need a count? In the beginning I thought this was a max retry count but it seems this would simply stop all loading if we have more than MAX_CONSECUTIVE_ERRORS?

maxandersen:
Gah! I've not pushed my cleaned up version. :/ 

I'll do that when back at laptop. 

You are fully correct it's overly cautious in this one.

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
maxandersen: its not needed. removed.

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
maxandersen: I lowered the max to 4  from 16. 

The code is retrying rather than stopping on first fail to avoid one real and one theoretical issue:

The real one:
We can't know (unless we lookup service files ourself and count) how many providers there are available.
Lets say you have a case where you are on Java 9 on Windows with panama, someones-crazy-jni-static-init-load-working-only-on-linux, jline4, jline3...and only jline3 can load.

Then hasNext() will on panama and jline4 will fail due to UnsupportedClassVersionError,
where "someones-crazy-jni-static-init-load-working-only-on-linux" will fail with Linkage on .next().
Each of them will be recoverable and it.hasNext() will eventually return true when it gets to jline3.

Thus it will trigger 3 exceptions, and require 3 consecutive retries to find the working one.

If the ordering is done differently - it could be 2 or less. If you add more backends - its whatever number of backends present that will be the maxium of retries needed.

Thus to cover this fully it should keep retrying forever. no knowable limit.

The theoretical one:

You could imagine some funky scenario where the mere loading of a provider brings the jvm in some weird state and the partial guarded code inside ServiceLoader to have the iterator always move forward
fails - then the loop will never end. It will keep retrying. 

That coupled with that there probably is a small finite number of possible backend providers we can set the number fairly low. 

Does that help or you still feel I'm being crazy ? :)

maxandersen:
pushed simplified version. should be good to go now.

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
melix: hasNext would throw? Why would it throw?

maxandersen: its the hasNext that fails with UnsupportedClass error as its the one trying to check if the next element is present, see from #108:

```
Exception in thread "main" java.lang.UnsupportedClassVersionError: dev/tamboui/backend/panama/PanamaBackendProvider has been compiled by a more recent version of the Java Runtime (class file version 66.0), this version of the Java Runtime only recognizes class file versions up to 55.0
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
        ...<class loader stack>
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:527)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:398)
        at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(ServiceLoader.java:1210)
        at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1221)
        at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1265)
        at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1300)
        at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1385) <---- hasNext() NOT next() fails
        at java.base/java.util.Iterator.forEachRemaining(Iterator.java:132)
```

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
```diff
@@ -0,0 +1,89 @@
...
+
+        int consecutiveErrors = 0;
+
+        while (true) {
+            try {
+                if (!it.hasNext()) {
+                    break;
+                }
+
+                S provider = it.next(); // may throw
```
melix, line 66: If this throws, the iterator would still be advanced, no? So the next iteration should be the next item in line. So say first 3 throw, the 4th would still be picked? Or are you saying that this could go infinite? In which case we'd never get a backend in the end, so why catch?

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
melix: Ok, so, if `hasNext()` fails 4 times in a row, I get the iterator cursor didn't advance and we're exiting the loop. Then in that case our service list will be empty even if there was one later that would work? Sorry if my questions are stupid, trying to understand if it really does what it's supposed to do :)

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
maxandersen: Yes, first 3 throws the 4th would be picked - yes; that is what would happen in the normal circumstance of having lets say 3 potential backends that require higher java version or missing some runtime dependency.

My understanding is that .hasNext() loads the class, .next() instantiates. Thus both can fail in a recoverable way (i.e. .hasNext()/.next() will work as should.

Thus need to catch both.

And i also can't tell how many times I need to catch...if I did I wouldn't need to have the error counter.

Coincidentally that counter also ensures we will always exit this loop no matter how weird erros or bad state might be in the serviceloaded code.

so- we must catch both for good reasons.

melix: Ok so I thought you were saying that `hasNext()` would fail but _not advance the cursor_, in which case, I don't see how this would improve the situation. If it _does_ then it's ok.

melix (approved):

tamboui-core/src/main/java/dev/tamboui/util/SafeServiceLoader.java
maxandersen: it does in all most likely cases - just the javadoc is stating it is not guaranteed. Which is just to say - don't just keep calling .hasNext() as it might fail - though very unlikely :)

PR #92, opened 2026-01-15
feat: add demo showing form/table constrain solved layout

maxandersen:
current flex demo are pure vertical/horizontal - this adds form + table demo; the table layouts is particular awsome IMO.

Unfortunatly you have to resize the window to see it in effect which our recorder does not (yet?) support so users have to run them manually to see it.


melix (approved):
Tested and approved!

PR #94, opened 2026-01-15
feat: Full CSS property coverage

melix:
With the new flex layout, it became evident that we needed to support more properties in CSS, the full constraints: width, height, filling, flex, ...

This is what this PR is doing. The custom component demo, which shows a CSS editor, is the perfect playground for this. It shows that we can completely get rid of programmatic constraint setup, only use CSS classes, then update live.

### DIFF
--- a/docs/src/docs/asciidoc/styling.adoc
+++ b/docs/src/docs/asciidoc/styling.adoc
@@ -186,6 +186,8 @@ Panel {
 
 === Properties
 
+==== Style Properties
+
 [cols="1,2,2",options="header"]
 |===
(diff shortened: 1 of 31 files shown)


maxandersen (approved):
this looks awesome.

the names we might adjust but i'll need to use this more in anger to be a judge of that. 
Are they mostly using web css naming conventions ?

melix:
Yes, except simplified because we don't have `min-width`, `max-width`, etc.. but this could be added if we feel the need. With what we have and limited widths/heights anyway, this looked sufficient to me.

Issue #91, opened 2026-01-15
consider use of Fraction type instead of double in layouts/pixel precision?


maxandersen:
as lesson from https://www.textualize.io/blog/7-things-ive-learned-building-a-modern-tui-framework/ use of Fraction type (which can be quite simple on longs and/or BigInteger) can avoid one-off cell errors when we start adding/multiplying with doubles.

Mainly in layout engine - but might also be relevant other places.

Additional win is one can retain a more logical relevant representation "1/3" instead "0.333333333%"

PR #97, opened 2026-01-15
fix: Fix wrapping of text in ListElement

melix:
Technically speaking, this commit does a bit more, since there was a more general problem. I noticed that when adding a text() element into a ListElement, then the overflow policy wasn't respected.

This commit fixes it, making sure that all containers provide their expected height.

(another issue found implementing a real app ;))

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/layout/cassowary/LayoutSolver.java
+++ b/tamboui-core/src/main/java/dev/tamboui/layout/cassowary/LayoutSolver.java
@@ -88,12 +88,15 @@ public final class LayoutSolver {
         // Solve and extract results
         solver.updateVariables();
 
-        int[] result = new int[n];
+        // Get float values and round with constraint preservation.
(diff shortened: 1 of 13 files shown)


tamboui-core/src/main/java/dev/tamboui/layout/cassowary/LayoutSolver.java
```diff
@@ -88,12 +88,15 @@ public int[] solve(List<Constraint> constraints, int available, int spacing, Fle
         // Solve and extract results
         solver.updateVariables();
 
-        int[] result = new int[n];
+        // Get float values and round with constraint preservation.
+        // Note: Using doubles can cause cumulative rounding errors. A more robust
+        // solution would use exact Fraction arithmetic - see issue #91.
```
maxandersen, line 93: was just about to comment this felt like #91  :)

graemerocher (approved):

PR #100, opened 2026-01-16
feat: Graceful error management

melix:
The rationale for this change is simple: at the moment, if, for some reason, an error happens during UI rendering (for example a widget throws an exception), the whole application crashes. This can be even more problematic in setups where the TUI needs to capture the standard output and error streams, to avoid UI trashing, because in that case, no exception would ever be shown to the user, with an application that could have its UI die without error.

To fix this, this PR introduces error management at the TuiRunner level. An error handler can be set, which allows different strategies, like displaying a standard error panel instead, logging to a file, etc.

On top of that, the Toolkit runner provides a failsafe mode for Element rendering, where a failing Element area can be replaced completely, allowing the _rest_ of the UI to work properly. To illustrate this, a demo has been added.

### DIFF
--- a/tamboui-tui/src/main/java/dev/tamboui/tui/error/ErrorContext.java
+++ b/tamboui-tui/src/main/java/dev/tamboui/tui/error/ErrorContext.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.tui.error;
+
+import java.io.PrintStream;
+
+/**
+ * Context provided to error handlers when a render error occurs.
+ * <p>
+ * Provides access to error output streams and control over the runner.
+ */
+public interface ErrorContext {
+
+    /**
+     * Returns the output stream for logging fatal errors.
+     * <p>
+     * This stream is configured via {@code TuiConfig.errorOutput()} and defaults
+     * to {@code System.err} (captured at config creation time). When the TUI has
+     * captured standard streams, this provides a way to log errors to an external
+     * destination.
+     *
+     * @return the error output stream
+     */
+    PrintStream errorOutput();
+
+    /**
+     * Signals the runner to stop.
+     * <p>
+     * Call this when the error handler determines the application should quit.
+     * The runner will perform cleanup and exit the main loop.
+     */
+    void quit();
+}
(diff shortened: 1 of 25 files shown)


graemerocher (approved):

tamboui-tui/src/main/java/dev/tamboui/tui/error/ErrorContext.java
```diff
@@ -0,0 +1,35 @@
...
+     * Returns the output stream for logging fatal errors.
+     * <p>
+     * This stream is configured via {@code TuiConfig.errorOutput()} and defaults
+     * to {@code System.err} (captured at config creation time). When the TUI has
+     * captured standard streams, this provides a way to log errors to an external
+     * destination.
+     *
+     * @return the error output stream
+     */
+    PrintStream errorOutput();
```
maxandersen, line 26: I might consider we made this a bit more highlevel to allow styled reporting; but we dont have that yet (is in my markup branch) so lets just keep it at printstream for now.

tamboui-tui/src/main/java/dev/tamboui/tui/error/RenderError.java
```diff
@@ -0,0 +1,113 @@
...
+ * Captures the exception that occurred during rendering along with a timestamp
+ * and a pre-formatted stack trace for display.
+ */
+public final class RenderError {
+
+    private final Throwable cause;
+    private final long timestamp;
+    private final String formattedStackTrace;
+
+    private RenderError(Throwable cause, long timestamp, String formattedStackTrace) {
```
maxandersen, line 23: why is there a need for passing in the formattedStacktrace here?

isn't Throwable enough which can then be styled accordingly when needed?

tamboui-tui/src/main/java/dev/tamboui/tui/error/ErrorContext.java
melix: Well this needs to be a stream because that's the "last resort" where you can print an error when the UI failed

tamboui-tui/src/main/java/dev/tamboui/tui/error/RenderError.java
melix: Yes I think we can simplify this

melix: Actually I think it's better like this, because the error can be used during rendering, and lazily formatting the stack trace will introduce a lot of overhead. Better format it once and reuse.

PR #101, opened 2026-01-16
fix: Simplify focusability API and always register elements for events

melix:
The previous API had several pain points:
  - Calling focusable() without id() caused silent failures (element not registered in focus chain)
  - TextInputElement required manual focusable()/id() unlike TextAreaElement
  - IDs could be changed after being set, causing unexpected behavior
  - needsEventRouting() added complexity with little performance benefit

Changes:
  - Auto-generate UUID for focusable elements at render time if no ID set
  - Make element IDs immutable once set (throws if changed)
  - TextInputElement is now always focusable with auto-generated ID
  - Component auto-generates ID instead of throwing on missing ID
  - Remove needsEventRouting() - all elements now register with EventRouter
  - Fix TextInputElement.handleKeyEvent() to check focused parameter
  - Add documentation clarifying focusable() vs onKeyEvent() relationship

The API is now more intuitive: call focusable() and it just works. It's orthogonal to event handling.

### DIFF
--- a/demos/error-handling-demo/build.gradle.kts
+++ b/demos/error-handling-demo/build.gradle.kts
@@ -1,0 +1,19 @@
+plugins {
+    id("dev.tamboui.demo-project")
+}
+
+description = "Demo showcasing error handling and fault-tolerant rendering"
+
+dependencies {
+    implementation(projects.tambouiToolkit)
+    runtimeOnly(projects.tambouiJline)
+}
+
+application {
+    mainClass.set("dev.tamboui.demo.errorhandling.ErrorHandlingDemo")
+}
+
+demo {
+    displayName = "Error Handling Demo"
+    tags = setOf("toolkit", "error-handling", "fault-tolerant", "exceptions")
+}
--- a/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
+++ b/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
@@ -1,0 +1,125 @@
+//DEPS dev.tamboui:tamboui-toolkit:LATEST
+//DEPS dev.tamboui:tamboui-jline:LATEST
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.demo.errorhandling;
+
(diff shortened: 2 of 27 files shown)


graemerocher (approved):

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/component/Component.java
```diff
@@ -67,9 +69,9 @@ protected boolean isFocused() {
 
     @Override
     protected final void renderContent(Frame frame, Rect area, RenderContext renderContext) {
+        // Auto-generate ID if not set (Component is always focusable)
         if (elementId == null) {
-            throw new IllegalStateException(
-                    "Component " + getClass().getSimpleName() + " must have an id. Use .id(\"myId\") when creating the component.");
+            elementId = UUID.randomUUID().toString();
```
maxandersen: do we really need a a full UUID ? those are slow ...

can't we make it an id derived from  tree path somehow ?

maxandersen: or have parent container enforce id's based on child position?

melix: I wouldn't use parent/child relationship. Another faster solution is simply to use an `AtomicLong`. WDYT?

maxandersen: I was thinking toIdentityString but realized AtomicLong is excellent as it gets to be short and assuming your rendering is deterministic the number should be the same between restarts - at least for non-dynamic constructs. 

Would you put the AtomicLong on Frame or as  a full global one?

I'm fine with either as long as we leave it as implementation detail and just say that if no id assigned we'll assign for you - nothing guaranteed about it.

btw. to avoid some random overlap maybe do "_tid_" + counter.getAndIncrement() so less likely to collide and its visible if dealing with a "random" assigned one?

melix: I would use an external IdGenerator class which handles static state.

maxandersen (approved):
+1

Maybe move IdGenerator into element or component package ?

melix:
> Maybe move IdGenerator into element or component package ?

I put it in a different package "just in case" we need other ids. No strong opinion though.

PR #102, opened 2026-01-16
fix: Implement Fraction type

### LINKED ISSUE #91: consider use of Fraction type instead of double in layouts/pixel precision?

melix:
This avoids cumulative rounding errors in the solver.

Fixes #91

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/layout/Constraint.java
+++ b/tamboui-core/src/main/java/dev/tamboui/layout/Constraint.java
@@ -131,6 +131,15 @@ public interface Constraint {
             return denominator;
         }
 
+        /**
+         * Converts this ratio to a Fraction for exact arithmetic.
+         *
+         * @return a Fraction representing this ratio
+         */
+        public Fraction toFraction() {
+            return Fraction.of(numerator, denominator);
+        }
+
         @Override
         public boolean equals(Object o) {
             if (this == o) {
--- a/tamboui-core/src/main/java/dev/tamboui/layout/Fraction.java
+++ b/tamboui-core/src/main/java/dev/tamboui/layout/Fraction.java
@@ -2,7 +2,7 @@
  * Copyright (c) 2025 TamboUI Contributors
  * SPDX-License-Identifier: MIT
  */
-package dev.tamboui.layout.cassowary;
+package dev.tamboui.layout;
 
 /**
  * An immutable exact rational number represented as a fraction.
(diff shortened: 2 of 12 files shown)


tamboui-core/src/main/java/dev/tamboui/layout/Fraction.java
```diff
@@ -0,0 +1,436 @@
...
+    /** The fraction representing zero (0/1). */
+    public static final Fraction ZERO = new Fraction(0, 1);
+
+    /** The fraction representing one (1/1). */
+    public static final Fraction ONE = new Fraction(1, 1);
+
+    /** The fraction representing negative one (-1/1). */
+    public static final Fraction NEG_ONE = new Fraction(-1, 1);
+
+    private final long numerator;
```
maxandersen, line 35: just realizing we have Constraint.Ratio that has the same setup - should those two merge?

Also - kinda feels this Fraction is usable outside cassowary so maybe move to .layout directly ?

tamboui-core/src/main/java/dev/tamboui/layout/cassowary/LayoutSolver.java
```diff
@@ -244,31 +249,57 @@ private void addEqualSizeTendency(Variable[] sizes) {
...
      *
-     * @param floatSizes the float sizes from the solver
-     * @param target     the maximum sum (available space)
+     * @param fractionSizes the exact Fraction sizes from the solver
+     * @param target        the maximum sum (available space)
      * @return integer sizes that sum to at most target
      */
-    private int[] roundWithConstraint(double[] floatSizes, int target) {
-        int n = floatSizes.length;
+    private int[] roundWithConstraint(Fraction[] fractionSizes, int target) {
```
maxandersen, line 265: hmm - I wouldn't have expected this method to become more complex.

tamboui-core/src/main/java/dev/tamboui/layout/Constraint.java
```diff
@@ -131,6 +131,15 @@ public int denominator() {
             return denominator;
         }
 
+        /**
+         * Converts this ratio to a Fraction for exact arithmetic.
+         *
+         * @return a Fraction representing this ratio
+         */
+        public Fraction toFraction() {
```
maxandersen, line 139: any reason to keep ratio as a type? can't just it directly in the ratio methods?

melix: Yes, one represents a `Constraint` (data class), the other one is a utility for computing.

maxandersen: ok :)

PR #105, opened 2026-01-17
fix: Consistent CSS styling

melix:
This commit refactors styling in Elements to make it more consistent:

- all Element implementations provide CSS properties to resolve styles
- there is a base method in `StyledElement` which performs the resolution, avoiding code duplication, with the right priority (explicit > CSS > default).
- some elements were only partially styleable using CSS, now all of them should be

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -157,6 +157,69 @@ class MyApp extends ToolkitApp {
 - keep each .tape short and focus. 
 - Try cover all demo features, but be aware they should not be too long as then the `.svg` renderings gets too big.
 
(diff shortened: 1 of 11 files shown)


maxandersen (approved):
Looks to make sense.

as a side note- I'm wondering if we could somehow make the css class/ids/attributes self documenting somehow? having something that can keep track on the properties looked up and be able to print out/show - even generate into api widget docs?

melix:
Yes, I want to be able to do something like this, but need to finish fixing the API/consistency/usability issues first.

PR #106, opened 2026-01-17
fix: Fix missing CSS gaps

melix:
This builds on top of #105 and fixes some missing CSS features, in particular:

- the ability to have lists for selectors
- matching attributes

This PR also systematically tests elements for rendering, which revealed a few more bugs.


melix:
At the moment I don't think it's possible. We don't have a "stable" element registry, persisted between frames. But it's an interesting idea to investigate.

PR #107, opened 2026-01-18
aggregated javadoc + markdown script

maxandersen:
This does a few things for javadoc:

1. a jbang script to convert all html to markdown javadoc - it is not yet applied; but can be experimented with. Run it with `jbang .jbang/ConvertJavaDocs.java --write` 

2. add javadoc style that matches theme of current asciidoctor guides, including hororing dark/white mode with toggle + javascript guided github link (with ability to set branch/tag to use for releases)

3. adds an aggregated javadoc bundle which are linked into main asciidoc generation.

image

The commits are separated out as I believe #1 and #2 are trivial. Things gets more dicy with #3.

Is where the code ballons a bit as the github link have to deal with aggregated so cant just rely on everything being in the same module. So now there is a index generated for package to github folder based path that gets injected into the generate site.

Then there is that we go from 19 mb to 35 mb, so close to double size just beacuse of javadoc.

Other concern is that for reasons I haven't figured out, if i include the panama module the build fails. no idea why. Best guess is that javadoc fails as panama is 100% a module, where others are only optional modules in /java11 so it works. 

I tried reducing warn/logs + use Java 25 where we were using Java 22 before but didn't help.

wdyt ? keep the aggregation in or out ?

### DIFF
--- a/buildSrc/src/main/kotlin/dev.tamboui.docs.gradle.kts
+++ b/buildSrc/src/main/kotlin/dev.tamboui.docs.gradle.kts
@@ -1,4 +1,7 @@
 import dev.tamboui.build.GenerateDemosGalleryTask
+import org.gradle.api.tasks.javadoc.Javadoc
+import org.gradle.external.javadoc.StandardJavadocDocletOptions
+import java.io.File
 
 plugins {
     id("org.asciidoctor.jvm.convert")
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/dev.tamboui.docs.gradle.kts
```diff
@@ -9,6 +12,44 @@ repositories {
...
+
+fun selectedJavadocProjects(): List<Project> {
+    // Include everything relevant, and explicitly exclude what you don't want.
+    // This scales better as modules are added/removed.
+    //
+    // Example:
+    //   -Ptamboui.docs.javadocExcludeModules=tamboui-panama-backend,tamboui-image
+    // Default: exclude the panama backend (it is JDK/toolchain sensitive and not part of the
+    // published website API surface yet). Set the property to an empty string to include it.
+    val configuredExcludes = providers.gradleProperty("tamboui.docs.javadocExcludeModules")
```
melix, line 38: :x: This is a Gradle antipattern. Things to include should either be declared as proper dependencies, or have automatic detection. The code here is excessively complex and shouldn't have to be. Relying on "magic" system properties for configuration is not a good practice.

buildSrc/src/main/kotlin/dev.tamboui.docs.gradle.kts
```diff
@@ -9,6 +12,44 @@ repositories {
...
+        .orElse("tamboui-panama-backend")
+        .get()
+    val excludedNames = mutableSetOf<String>()
+    if (configuredExcludes.isNotBlank()) {
+        excludedNames.addAll(configuredExcludes.split(',').map { it.trim() }.filter { it.isNotEmpty() })
+    }
+
+    return rootProject.subprojects
```
melix, line 46: :x: You should not do this. We have plugins that tell us what each project kind is. If we want to use automatic inclusion that's what should be used, not fragile project paths.

buildSrc/src/main/kotlin/dev.tamboui.docs.gradle.kts
```diff
@@ -57,8 +98,130 @@ val prepareAsciidocSources = tasks.register<Sync>("prepareAsciidocSources") {
...
 }
 
+/**
+ * Aggregated Javadoc for the website.
+ *
+ * We generate a single unified Javadoc across the relevant library modules (excluding demos),
+ * and then publish it alongside the Asciidoctor site under {@code javadoc/}.
+ */
+val aggregateJavadocOutputDir = layout.buildDirectory.dir("generated/javadoc")
+val aggregateJavadoc = tasks.register<Javadoc>("aggregateJavadoc") {
```
melix, line 108: :x: This is way too complicated to live in a build script. Complex build logic should be isolated in tasks, not live in build scripts when it's open door to hacks and breaking build caching or up-to-date checking.

tamboui-panama-backend/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
 description = "Panama FFI backend for TamboUI TUI library"
 
 tasks.withType<JavaCompile>().configureEach {
-    options.release = 22
+    options.release = 25
```
melix, line 8: Why?

melix (changes requested):
I'm not against it, but IMO this PR mixes too many things together and the implementation is not very clean IMHO, this will bite us. The javadoc aggregation implementation should be simpler. I have done this in other projects, I can create a separate PR just for that aspect if you want.

maxandersen:
> I'm not against it, but IMO this PR mixes too many things together and the implementation is not very clean IMHO, this will bite us. The javadoc aggregation implementation should be simpler. I have done this in other projects, I can create a separate PR just for that aspect if you want.

yes. I had a much simpler one but it failed because of following:

1. querying gradle resulted in errors about concurrency access - I probably did it wrong ;)
2. the panama module triggers a fail i just couldn't uncover why; but it hits duplicated package names or multi module-info errors.

Thats why I split this into separate commits as I wasn't happy for the aggregation.

If you know of a way that works with modules and all I'm all ears for a PR for that.

tamboui-panama-backend/build.gradle.kts
maxandersen: was an attempt because the -X's flags was failing anyway when not building on java 25.

maxandersen:
replaced with #116

PR #109, opened 2026-01-18
fix: Prioritize keyboard input over queued tick events in TuiRunner

melix:
Changed pollEvent() to check for terminal input (keyboard/mouse) with a non-blocking read before checking the event queue. This ensures user input is always processed even when tick events are continuously queued by the scheduler.

Previously, if tick events were produced faster than the main loop could process them, keyboard input would never be read because the queue was always non-empty. This was highlighted by the JTop demo which appeared unresponsive to keystrokes.


maxandersen:
So I have a slow computer and didn't see it ? :)

melix:
That's the other way around: the demo uses the system monitor library _during_ the render loop (which will be fixed in a separate PR). This isn't good because apparently it's quite slow under Linux.

PR #110, opened 2026-01-18
feat: Javadoc aggregation

melix:
Look at the second commit, because this PR is built on top of #109


davsclaus:
some of the .javas source files does not have a top license header.

And the license header may have (c) year in the past.

maxandersen:
"filters out module-info.java (which will break javadoc generation
because there's no single module info)."

yeah - so thats probably root cause i got stuck last night. Thats so annoying javadoc assumes one .jar = one javadoc "site".

Means you loose the module info for the docs - annoying but yeah, will have to do because having to go each javadoc site manually is a pain.

maxandersen:
Merge when dependent one in and I'll update the style etc. pr to use this.

melix:
@davsclaus good catch about the license headers, they are incorrect. Let me fix.

PR #112, opened 2026-01-18
fix: Performance optimizations

melix:
These problems were highlighted by the JTopDemo:

1. the Jtop demo was doing expensive work on the render loop (getting system statistics), which was causing slowness in rendering
2. the layout that the demo uses, with many processes and bars created tons of constraints. The layout solver had a hard time dealing with this. I introduced a layout cache, which makes it **dramatically faster** (from 8fps to 1000+).

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/layout/Layout.java
+++ b/tamboui-core/src/main/java/dev/tamboui/layout/Layout.java
@@ -6,6 +6,7 @@ package dev.tamboui.layout;
 
 import static dev.tamboui.util.CollectionUtil.listCopyOf;
 
+import dev.tamboui.layout.cassowary.LayoutCache;
 import dev.tamboui.layout.cassowary.LayoutSolver;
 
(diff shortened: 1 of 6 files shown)


maxandersen:
btw. its java 25 with panama gives me 0.2 fps
java 21 with jline is 0.3 fps ...

something definitely not working as expected.

melix:
uh!?

maxandersen:
ok - its not this specific PR causing it - its #109 that kills me. if I revert that on main I get 4 fps on main.
if i revert it on this branch I get ~25 fps.

poking at it with jvisualvm  - haven't concluded yet where it goes wrong except I can see readEvent is spending *alot* of cpu withuot the revert.

melix:
Ok so I would suggest we merge this one, because it makes a dramatic difference, and investigate why #109 has opposite effects for you.

maxandersen:
fyi, i pasted some jvisualm om zulip

PR #113, opened 2026-01-18
fix: Rename JLine backend

melix:
This commit changes the jline backend name to jline3. For consistency, it has been moved into a `tamboui-jline3-backend` directory, similar to the `tamboui-panama-backend` directory. The name "jline3" is in prevision of a future JLine 4 backend.

Unfortunately, using a digit in a module name is trigerring a warning, so the "3" is converted to "_three". Ugly but no choice.


maxandersen (approved):

maxandersen:
That three is such a weird Jpms limit. But how about we just embrace and use same name as artifact name ?

melix:
uhhh, honestly I wouldn't. I mean, yes, for the sake of consistency, but for the sake of my mental sanity, I'd rather not :D

maxandersen:
is it "[module] module name component jline3 should avoid terminal digits" you are concerned about ? 

Then that is just jpms linter being dumb/pedantic (IMO). 

Its just to try avoid users start using versions in names as jpms tries to avoid versions. Something I and reality of software distribution completely disagree with :) jline2, jline3, and soon jline4 is where this makes perfect sense. 

It is a perfectly valid name imo.

melix:
I completely agree with you. But this warning yelled at us on every build is painful.

maxandersen:
> I completely agree with you. But this warning yelled at us on every build is painful.

and here i'm struggling to get to spot errors as doclint warnings is 100's of lines :)

PR #116, opened 2026-01-18
Styled Javadoc w/darklight mode + github links

maxandersen:
This replaces #107

 - adds a API link to asciidoc header that links to /api to get aggregated javadoc  
 - adds a javadoc css that mimicks main site, sharing dark/light mode styles
 - adds .js that adds a dark/light mode switch 
 - .js also adds a GitHub link on each javadoc page that point to github main branch by default, but can be overwritten in the build using `tamboui.githubRef` property
 - for aggregated there is a slightly different github lookup as can't deduce location based on gradle module name; this generates a small index that gets injected only in the aggregated javadoc to handle that mapping.
 
 In end both individual .javadoc jars and the main site have consistent styles and functionallity.


maxandersen:
p.s. i have asciinema player switching theme too betwee light and dark but lets not over complicate this PR.

melix:
I will push some changes to the Gradle configuration, there are things which caught my eye.

maxandersen:
tahnks for the clean up!

PR #117, opened 2026-01-18
use markdown in javadoc

maxandersen:
POC showing how it will look if move to use markdown in all html.

given we are building with java 25 I was wondering if we can't just use markdown everywhere as the html output doesn't care about what java version you are running with.

marking as draft as haven't triple checked if all modules actually works.


maxandersen:
> This looks interesting. My eyes will have to get used to the `///` though :D

:)

i did a quick test in intellij and it seems to not care wether its one or the other. if javadoc jar is present it takes it from there - if not javadoc jar it renders it nicely wether html or markdown.

my main worry is ides that might have logic like "your project is java 17, we wont render markdown even in depednent sources then" ...but i feel like thats their problem to fix :)

PR #118, opened 2026-01-18
fix: Use separate thread for key events

melix:
The fix in #109 only worked with the Panama backend, because setting timeout to 0 meant non-blocking. But with JLine, this just waits indefinitely. Therefore, the solution was making things worse with the JLine backend.

This commit reworks the fix so that we read key events in a separate, dedicated thread. The main loop with then prioritize key events, for responsiveness, before rendering events. Therefore, if a render is slow, the key events will still be processed.

### DIFF
--- a/tamboui-tui/src/main/java/dev/tamboui/tui/TerminalInputReader.java
+++ b/tamboui-tui/src/main/java/dev/tamboui/tui/TerminalInputReader.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.tui;
+
+import dev.tamboui.terminal.Backend;
+import dev.tamboui.tui.bindings.Bindings;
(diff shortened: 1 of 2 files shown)


maxandersen:
whats weird is my mac mini dont seem to be as effected - going to test more on my laptop.

It doesn't surprise me we with jline need to do this but it somehow bothers me that it has to be default/enforced because if i grok it right - if app would not tick automatically this extra thread would not be necessary, true?

That said - definitely think this is a valid approach - and wondering if a base version of this should be lower in core to be used when not doing full toolkit?

anyhow - let me try out the various ones on laptop

maxandersen (approved):
this definitely seem needed for jline and as such probably make sense to generalize and move further down in stack - or at least document how to do.

might not be possible/feasible to put in the backend but feels like it would be nice panama wasnt enforced to use threads if not needed.

melix:
In practice I think it's no harm to have a thread dedicated to key events. We should probably have one for render events too, but that's a much bigger change I think.

PR #119, opened 2026-01-19
feat: Add markup text elements with CSS class-based effect targeting

melix:
- Add MarkupParser for BBCode-style markup in tamboui-core
  - Add MarkupTextAreaElement that renders tags as CSS classes
  - Add RichTextAreaElement/RichTextState for scrollable rich text
  - Enable TFX effects to target markup segments via CSS selectors
  - Add thread safety to ElementRegistry and ToolkitRunner
  - Add RichTextDemo showcasing animated markup text

This PR also has some thread safety changes when resizing, that could be backported to main if this branch is not merged.

### DIFF
--- a/docs/src/docs/asciidoc/markup.adoc
+++ b/docs/src/docs/asciidoc/markup.adoc
@@ -136,7 +136,7 @@ Combine multiple styles in a single tag by separating them with spaces:
 [bold underlined yellow]Warning![/]
 ----
 
-The first token becomes the primary tag name (used for CSS class targeting).
+All tokens in a compound style become CSS classes for targeting (see <<CSS Class Targeting>>).
 
 === Background Colors
 
@@ -263,22 +263,36 @@ Text t4 = MarkupParser.parse("[error italic]Message[/]", resolver);
 
 == CSS Class Targeting
 
-All tags (including compound styles) create a CSS class that can be targeted with stylesheets.
-The primary tag name (first token in compound styles) becomes the class:
+All tokens in a tag become CSS classes that can be targeted with stylesheets.
+For compound styles, every token (except `on`) becomes a CSS class:
 
 [source,java]
 ----
 Text text = MarkupParser.parse("[error]message[/]");
 // The span has Tags extension containing "error"
 
-Text compound = MarkupParser.parse("[error bold]message[/]");
-// The span has Tags extension containing "error" (primary tag)
+Text compound = MarkupParser.parse("[bold underlined yellow]Warning![/]");
+// The span has Tags extension containing "bold", "underlined", "yellow"
+
+Text background = MarkupParser.parse("[white on blue]text[/]");
+// The span has Tags extension containing "white", "blue" (not "on")
 ----
 
-This allows external CSS to style markup-based text:
+This allows external CSS to target any style token:
 
 [source,css]
 ----
+/* Target all bold text */
+.bold {
+    text-style: bold;
+}
+
+/* Target all yellow text */
+.yellow {
+    color: #ffcc00;
+}
+
+/* Target error messages */
 .error {
     color: red;
     text-style: bold;
--- a/tamboui-core/src/main/java/dev/tamboui/text/MarkupParser.java
+++ b/tamboui-core/src/main/java/dev/tamboui/text/MarkupParser.java
@@ -291,11 +291,20 @@ public final class MarkupParser {
             // Flush current text with current style
             flushCurrentText();
 
-            // Use first token as the tag name for Tags (for CSS targeting)
-            String primaryTag = tagName.split("\\s+")[0];
+            // Parse all tokens for CSS class targeting
+            String[] tokens = tagName.split("\\s+");
+            String primaryTag = tokens[0];  // First token used for closing tag matching
 
-            // Create a Tags extension for this tag name
-            Style tagStyle = Style.EMPTY.withExtension(Tags.class, Tags.of(primaryTag));
+            // Collect all tokens as CSS class tags (excluding "on" keyword)
+            List<String> tagList = new ArrayList<>();
+            for (String token : tokens) {
+                if (!token.isEmpty() && !"on".equals(token)) {
+                    tagList.add(token);
+                }
+            }
+
+            // Create a Tags extension with all tokens
+            Style tagStyle = Style.EMPTY.withExtension(Tags.class, Tags.of(tagList.toArray(new String[0])));
 
             // Check for link tag
             if ("link".equals(primaryTag) && attribute != null) {
--- a/tamboui-core/src/test/java/dev/tamboui/text/MarkupParserTest.java
+++ b/tamboui-core/src/test/java/dev/tamboui/text/MarkupParserTest.java
@@ -377,13 +377,39 @@ class MarkupParserTest {
     }
 
     @Test
-    @DisplayName("compound style preserves Tags for CSS targeting")
+    @DisplayName("compound style preserves all Tags for CSS targeting")
     void compoundStylePreservesTags() {
         Text text = MarkupParser.parse("[bold red]styled[/]");
 
         Tags tags = text.lines().get(0).spans().get(0).style()
             .extension(Tags.class, Tags.empty());
+        // All tokens become CSS classes
         assertThat(tags.contains("bold")).isTrue();
+        assertThat(tags.contains("red")).isTrue();
+    }
+
+    @Test
+    @DisplayName("compound style with multiple tokens all become tags")
+    void compoundStyleAllTokensBecomeTags() {
+        Text text = MarkupParser.parse("[bold underlined yellow]Warning![/]");
+
+        Tags tags = text.lines().get(0).spans().get(0).style()
+            .extension(Tags.class, Tags.empty());
+        assertThat(tags.contains("bold")).isTrue();
+        assertThat(tags.contains("underlined")).isTrue();
+        assertThat(tags.contains("yellow")).isTrue();
+    }
+
+    @Test
+    @DisplayName("compound style with background excludes 'on' keyword from tags")
+    void compoundStyleExcludesOnKeyword() {
+        Text text = MarkupParser.parse("[white on blue]highlighted[/]");
+
+        Tags tags = text.lines().get(0).spans().get(0).style()
+            .extension(Tags.class, Tags.empty());
+        assertThat(tags.contains("white")).isTrue();
+        assertThat(tags.contains("blue")).isTrue();
+        assertThat(tags.contains("on")).isFalse();
     }
 
     @Test
@@ -433,23 +459,27 @@ class MarkupParserTest {
         Style style = text.lines().get(0).spans().get(0).style();
         assertThat(style.fg()).contains(Color.RED);
         assertThat(style.bg()).isEmpty();  // "foo" is not a color
-        // Primary tag is "red" for CSS targeting
+        // All tokens (except "on") become CSS classes
         Tags tags = style.extension(Tags.class, Tags.empty());
         assertThat(tags.contains("red")).isTrue();
+        assertThat(tags.contains("foo")).isTrue();
+        assertThat(tags.contains("on")).isFalse();
     }
 
     @Test
-    @DisplayName("custom tag with background color - primary tag used for CSS")
+    @DisplayName("custom tag with background color - all tokens become CSS classes")
     void customTagWithBackgroundColor() {
-        // [foo on red] - "foo" becomes the CSS class, red becomes background
+        // [foo on red] - both "foo" and "red" become CSS classes, red becomes background
         Text text = MarkupParser.parse("[foo on red]text[/]");
 
         Style style = text.lines().get(0).spans().get(0).style();
         assertThat(style.fg()).isEmpty();  // "foo" is not a color
         assertThat(style.bg()).contains(Color.RED);
-        // Primary tag is "foo" for CSS targeting
+        // All tokens (except "on") become CSS classes
         Tags tags = style.extension(Tags.class, Tags.empty());
         assertThat(tags.contains("foo")).isTrue();
+        assertThat(tags.contains("red")).isTrue();
+        assertThat(tags.contains("on")).isFalse();
     }
 
     @Test


tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
```diff
@@ -145,28 +147,34 @@ public void run(Supplier<Element> elementSupplier) throws Exception {
...
-                // Auto-focus first focusable element if nothing is focused
-                if (focusManager.focusedId() == null && !focusManager.focusOrder().isEmpty()) {
-                    focusManager.setFocus(focusManager.focusOrder().get(0));
-                }
-
-                // Apply post-render processors (e.g., effects, overlays)
-                for (ToolkitPostRenderProcessor processor : postRenderProcessors) {
-                    processor.process(frame, elementRegistry, lastElapsed);
+                // Synchronize rendering to prevent concurrent access from scheduler thread
+                renderLock.lock();
```
maxandersen: I assume this is not related to markup ?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/text/MarkupSegmentParser.java
```diff
@@ -0,0 +1,440 @@
...
+    private MarkupSegmentParser() {
+        // Utility class
+    }
+
+    /**
+     * A segment of text with its associated tags and computed style.
+     */
+    public static final class Segment {
+        private final String text;
+        private final Set<String> tags;
```
maxandersen: couldn't these be extensions on the style directly? and we wouldn't need a separate parser?

tamboui-toolkit/demos/richtext-demo/src/main/java/dev/tamboui/demo/RichTextDemo.java
```diff
@@ -0,0 +1,271 @@
...
+import dev.tamboui.tui.bindings.KeyTrigger;
+import dev.tamboui.tui.event.KeyCode;
+import dev.tamboui.widgets.input.TextAreaState;
+import dev.tamboui.widgets.tabs.TabsState;
+import dev.tamboui.text.MarkupParser;
+
+import static dev.tamboui.toolkit.Toolkit.*;
+
+/**
+ * Demo showcasing the RichText and MarkupText components.
```
maxandersen, line 33: this demo is definitely cool.

fun to define .error as a color but was kinda surprised that .red was not possible to redefine.

is that intentional?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
maxandersen: i see its same as #120 so i'll just ignore it here :)

tamboui-toolkit/demos/richtext-demo/src/main/java/dev/tamboui/demo/RichTextDemo.java
melix: Yes, the parser supports built-in colors. For everything not recognized we assume CSS class.

maxandersen: but why not treat them equal? why should css be able to reconfigure everything via selector but not the builtin styles/classes?

melix: Because the built-in styles represent factual things, like "red", "green", "underline". Allowing to override them would mean that you could have red "blue" or underline "bold". That doesn't quite make sense to me. That's why we have CSS on top.

maxandersen: you mean as in RED refers to ansi color #2 which terminal themes can override to be any color ?

Then yes, I agree RED's default behaviour should be that - but also seems useful being able to use tcss as ways to define overrides for such themes if you prefer being able to have always readable colors.

with terminal themes you risk your app gets unreadable - which is a user choice; but some apps might want to make use of more reliable setups.

This all relates to figure out how to do coloring/styles - https://textual.textualize.io/guide/design has interesting approach of setting base colors and then derive set of consistent variables from that which provides a much richer and consistent look...but ALSO have option of honoring ansi colors (ansi_colors: true option).

my thinking is we default to use ansi_colors so its "fits in" by default, but have option to use richer mechanics and for that to work you'll want to override the notion of "ansi color 2"/"red" i would think.

I agree "underline" is a different thing though...I could be swayed to keep that an absolute truth.

maxandersen: Note - how we handle overrides / consistency is complementary to this pr. 

In this I'm more curious to grok the need for two parsers/mechanisms

melix: Oh it's not required technically speaking. In fact, I want to consolidate both. Initially, there wasn't the segment version. It's an artifact of how the feature was built. The initial "rich area" was built at the widget level, which didn't let us target specific elements in TFX, since TFX works at the named style/id level, which is higher level than what we have in core/widgets. I think we have 2 options:

1. use an extension in `Style` to represent the tags. That would work, but it would assume that all tags relate to styling. In practice, this is almost always the case, but we could have more in the future.
2. add an extension mechanism to `Span`, similar to `Style`, to store metadata associated to a particular span. This would mean "this span was tagged as `foo`" and then you'd have the widget/element deal with this information to render styles (raw / using style engine).

Option 1 is what you chose for Hyperlink, but it's hardcoded somehow. Instead of having a `Hyperlink` extension type on `Style`, you have a property.

melix: So I kind of feel we should do 1. But implementation wise, there's a difference between URLs and tags. For tags, we want to accumulate (e.g if you do `[foo][bar]text[/bar][/foo]`. The current extension mechanism in `Style` overwrites extensions. Therefore, I would suggest to have a `Patchable` interface that extensions _can_ implement:

```java
public interface Patchable<T> {                                                                                                                                                                                                   
      T patch(T other);                                                                                                                                                                                                             
  } 
```

then the `Tags` extension would do this:

```java
public class Tags implements Patchable {
     @Override
     public Tags patch(Tags other) { ... }
``` 

The `Style` code would only have to differentiate between extensions implementing `Patchable` or not.

melix: Commit https://github.com/tamboui/tamboui/pull/119/commits/3d422951f5e6166f6b1780f31128fe9f88740bcc implements this strategy.

maxandersen: interesting approach.

The rationale for hyperlink at style is more to have the ansi hyperlinks as first class.

As long as we can still have notion of markup at core level I'm in :)

i'll take a look the new commit - Patchable feels like one could just have one interface but maybe im missing somethign :)

maxandersen:
just some comparisons between what this branch does vs maxandersen/markup branch:

### cc/markup vs. maxandersen/markup

| Capability | `cc/markup` (current branch) | `maxandersen/markup` (remote branch) |
|---|---|---|
| **Core entrypoint** | `dev.tamboui.text.MarkupParser.parse(...)` | `dev.tamboui.text.markup.Markup.render(...)` |
| **overlapping tags [red][blue][/red][/blue]** | ✅  | ✅ 
| **implicit close [/]** | ❌ | ✅ 
| **fg on bg [red on blue]** |  ❌ |  ✅  
| **styles in tags [bold red]** |  ❌ |  ✅  
| **Escaping** | `[[` → `[` and `]]` → `]` | Backslash escape `\[red]`  |
| **Emoji shortcodes (`:smiley:` etc.)** | ❌ No | ✅ Yes |
| **Theming (semantic tags like `[error]`)** | ❌ No core theme | ✅ Yes  |
| **Unknown tags behavior** | Preserved as metadata (`Tags`) with no visual style unless resolver/TCSS adds style | Theme/style-parser driven; malformed/mismatched tags raise `MarkupParseException` |
| **Error handling** | Forgiving (generally doesn’t throw on mismatches) | Strict (throws `MarkupParseException` on mismatched closing tags, etc.) |
| **“Tags/classes” for higher-level targeting** | ✅ Yes: `dev.tamboui.style.Tags` stored as a `Style` extension; merged via `Patchable` | ❌ No equivalent “span classes” concept in the markup stack |
| **Style patching extensibility** | ✅ `Patchable` allows merge semantics for `Style` extensions (`Tags` unions) | ❌ No `Patchable`; no `Style` extension merging hook needed |
| **Extra console/ANSI tooling included** | Not present | ✅ `RichIO`, `TextAnsiRenderer`, OSC8 hyperlink rendering support |
| **Code location** | Mostly `tamboui-core` + toolkit elements (`MarkupTextElement`, `MarkupTextAreaElement`) | Mostly `tamboui-core` (markup/theme/emoji/ansi/console) + `demos/markup-demo` |


Overall I think we should merge these two approaches. Taking the bbcode/unicode handling + core/widget support from maxandersen and the generic tag/patchable approach from this branch allowing higherlevels to go further.

Theming and StyleResolver is afaics more or less interchangable - difference is about what can be overridden/defined.

Error handling is another - where I'm a bit not sure when to throw or not.

maxandersen:
btw. about escape -  I jsut did \[ as its what rich did so i could use their testsuite :) but [[ ]] is nice in java strings.
we can do [[ and consider optionally also have \ if find need.

melix:
Very interesting comparison. I'm a bit confused about the "semantic theming". In my approach you basically have it for free, since _any_ tag is converted. So not sure why it's marked as :x:

maxandersen:
> Very interesting comparison. I'm a bit confused about the "semantic theming". In my approach you basically have it for free, since _any_ tag is converted. So not sure why it's marked as ❌

yeah - maybe you are right. Just that this branch hardcodes the built-ins; where mine look them up in the Theme. So yes, definitely close/similar enough mechanics - just applied slightly different.

melix:
@maxandersen So I have made changes to the infrastructure, making this even more compelling. Basically I managed to move the logic to the core/widget. This means that the markup area can basically rely on existing widgets like `Paragraph`, while still allowing the TFX effects to target using CSS classes. In addition I have added the missing parser features from your branch. 

However, I feel like the theming/RichIO stuff that is mentioned in the table is orthogonal and most likely redundant: for theming we already have live switching of CSS, and for RichIO we already have the inlined display.

maxandersen:
I haven't had time to look deep on it - will do tomorre; so just a few comments on:

> However, I feel like the theming/RichIO stuff that is mentioned in the table is orthogonal and most likely redundant: for theming we already have live switching of CSS, and for RichIO we already have the inlined display.

Themes or StyleResolver is just because I couldn't see how to apply CSS without the full upper framework - and themes literally just is about being able to define the base 16 colors. And be able to load in various theme formats to apply.

But if there is a way to use CSS for that using core - definitely interested seeing how.

RichIO is not really redundant as its for the opposite of tuiruner/dsl layers - its jsut for printing out rich text / widgets in more classic console print usecases.

maxandersen:
couldn't resist - did you leave the emoji github names out or missed it?

melix:
> Themes or StyleResolver is just because I couldn't see how to apply CSS without the full upper framework - and themes literally just is about being able to define the base 16 colors. And be able to load in various theme formats to apply.

Right, for the base 16 colors. I'm not sure that "theme" is the right concept for this, as for me a theme is a full "CSS" (bare with me the name here). But in any case seems that this can be done independently.

> RichIO is not really redundant as its for the opposite of tuiruner/dsl layers - its jsut for printing out rich text / widgets in more classic console print usecases.

But `InlineDisplay` is part of core. I don't think you need `RichIO`, it's just about calling `Buffer#toAnsiString`. You can write:

```java
try (InlineDisplay display = InlineDisplay.create(1)) {                                                                      
      display.setLine(0, MarkupParser.parse("Hello [red]world[/red]!"));                                                       
  }
```

And if you really don't want to use this, you can do:

```java
Text text = MarkupParser.parse("Hello [red]world[/red]!");                                                                   
Buffer buf = Buffer.empty(Rect.of(80, 1));                                                                                   
buf.setLine(0, 0, text.lines().get(0));                                                                                      System.out.println(buf.toAnsiStringTrimmed());
```

Given that it's really straightforward, I'm not convinced we need _another_ utility for this.

melix:
> couldn't resist - did you leave the emoji github names out or missed it?

Yes I intentionally left that out for 2 reasons:

- from my tests, emoji rendering varies widely depending on the terminal and font being used (2 dimensions). Most often rendering blank. I'm a bit sad, but I don't want users to have false expectations and complain that it doesn't work
- this requires a "database" of emoji names, but you've already done that, but choosing the GH shortcodes is arbitrary (there are more schemes, e.g Slack or Discord, even if they share a lot in common).

maxandersen:
> > couldn't resist - did you leave the emoji github names out or missed it?
> 
> Yes I intentionally left that out for 2 reasons:
> 
> * from my tests, emoji rendering varies widely depending on the terminal and font being used (2 dimensions). Most often rendering blank. I'm a bit sad, but I don't want users to have false expectations and complain that it doesn't work

What emojii rendering was affected ? I'm assuming it was when you use wide/multi characters?

fonts, and terminal - yes some dont do it but its very few left these days - and by adding a syntax for emoji we can provide a "on|off" switch to allow fallback in those cases.

> * this requires a "database" of emoji names, but you've already done that, but choosing the GH shortcodes is arbitrary (there are more schemes, e.g Slack or Discord, even if they share a lot in common).

I can follow not adding the multi character into the default "database" - I do though wish we are able to handle them as they are more and more common; particularly in asia.

I looked up github, discord, slack emojii databases and they have large common set of names which we could use imo - slack and discord separates themselves from github mainly by each server allowing redefining or adding names which get converted into not just unicode emojii but also animations and images. I'm not aiming for that level - but having the common set available I think will be good - and in future expand the ability for :<name>: syntax to have custom lookup if there is ask/usecases found for it.

but we can do the unicode feature in separate pr.

maxandersen:
> > Themes or StyleResolver is just because I couldn't see how to apply CSS without the full upper framework - and themes literally just is about being able to define the base 16 colors. And be able to load in various theme formats to apply.
> 
> Right, for the base 16 colors. I'm not sure that "theme" is the right concept for this, as for me a theme is a full "CSS" (bare with me the name here). But in any case seems that this can be done independently.

Yeah, I just used theme here was that s the named used in Textual, ratatui, charm and various terminals - but yes not hung on the name :)

> 
> > RichIO is not really redundant as its for the opposite of tuiruner/dsl layers - its jsut for printing out rich text / widgets in more classic console print usecases.
> 
> But `InlineDisplay` is part of core. I don't think you need `RichIO`, it's just about calling `Buffer#toAnsiString`. You can write:
> 
> ```java
> try (InlineDisplay display = InlineDisplay.create(1)) {                                                                      
>       display.setLine(0, MarkupParser.parse("Hello [red]world[/red]!"));                                                       
>   }
> ```
> 
> And if you really don't want to use this, you can do:
> 
> ```java
> Text text = MarkupParser.parse("Hello [red]world[/red]!");                                                                   
> Buffer buf = Buffer.empty(Rect.of(80, 1));                                                                                   
> buf.setLine(0, 0, text.lines().get(0));                                                                                      System.out.println(buf.toAnsiStringTrimmed());
> ```
 
> Given that it's really straightforward, I'm not convinced we need _another_ utility for this.

Doing the above everytime wanting to print is for me noise compared to:

```
RichIO.println("Hello [red]world[/]!");
```

I like to have the ability to render tables/trees/markdown etc without user having to fully engage with out having to go "full tui"

Anyhow - this can be done separately and even put in a dedicated module as long as the core has the mechanics.

docs/src/docs/asciidoc/markup.adoc
```diff
@@ -0,0 +1,396 @@
...
+Combine multiple styles in a single tag by separating them with spaces:
+
+[source]
+----
+[bold red]Error message[/]
+[italic cyan]Hint text[/]
+[bold underlined yellow]Warning![/]
+----
+
+The first token becomes the primary tag name (used for CSS class targeting).
```
maxandersen: so [bold underlined yellow] gets to be `.bold` class?

shouldn't it have all the classes (.bold, .underlined, .yellow) ? or a combination (.bold_underlined_yellow)?

In any case I think it would be good if the docs explicilty stated with an example what selector/class would show up.

tamboui-core/src/main/java/dev/tamboui/buffer/Buffer.java
```diff
@@ -93,6 +95,19 @@ public static Buffer withLines(String... lines) {
         return buffer;
     }
 
+    /**
+     * Sets a listener to be notified when styled content is written to this buffer.
+     * <p>
+     * The listener receives the style and area for each span written to the buffer.
```
maxandersen, line 101: not fully following what this is for exactly?

maxandersen: what is it we can't do without this?

melix: That's a key aspect of the late changes on that PR. It allows CSS selectors to target both `Element` with ids/classes AND simple styled text (at the core level).

maxandersen: gotcha. kinda figured it out. it feels a bit weird to me having this "circular" connection betwen buffer and higher ups ...but I reckon the alternative is to tie it even harder together which is even worse.

maxandersen (approved):
+1 on putting this. We can add/iterate on emoji, style/themeing  seperatley!

https://github.com/tamboui/tamboui/pull/119#discussion_r2712061568 is still open - but also something i think we might evolve on once playing with it for real.

tamboui-core/src/main/java/dev/tamboui/buffer/Buffer.java
melix: Yeah and you haven't seen the intermediate steps which were even uglier. That's the simplest I could come up with which didn't feel too "invasive".

docs/src/docs/asciidoc/markup.adoc
melix: No I think you're right. All should become "classes", or "tags" as this stage (it only becomes a "class" in the context of CSS). I think the discrepancy comes from these built-in colors/types, which we'd want to replace. However, this comes at a cost. If `blue` becomes a tag, then it's technically equivalent to calling it `foo`, so what does `foo on bar` mean when both are tags?

maxandersen: > so what does `foo on bar` mean when both are tags?

thats where "foo_on_bar" might be more meaningfull - i.e. one class per tag combo...rather than individual classes.  so, [error] is class error, [blue on red underline] is [blue_on_red_underline] and [blue on red][underline] is blue_on_red and underline.

PR #120, opened 2026-01-19
fix: Thread safety

melix:
This commit fixes potential concurrent modificatione exceptions during rendering. The causes can be different: resizes triggering renders out of the main loop (difficult to change) or concurrent modification of the element registry.

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
@@ -230,8 +230,8 @@ public final class ToolkitRunner implements AutoCloseable {
     /**
      * Schedules an action to run after a delay.
      * <p>
-     * The action runs on a background thread. If the action modifies shared state,
-     * ensure proper synchronization. The UI will automatically redraw on the next
+     * The action is synchronized with the render loop to prevent concurrent
+     * modification of shared state. The UI will automatically redraw on the next
      * tick event after the action completes.
      *
      * <pre>{@code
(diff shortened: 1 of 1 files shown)


tamboui-tfx/src/main/java/dev/tamboui/tfx/EffectManager.java
```diff
@@ -93,15 +93,17 @@ public boolean isRunning() {
...
+
+        for (Effect effect : snapshot) {
             effect.process(duration, buffer, area);
             if (effect.done()) {
                 toRemove.add(effect);
             }
         }
-        
+
         effects.removeAll(toRemove);
```
maxandersen, line 107: doesn't this risk concurrent modifcation too?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/ToolkitRunner.java
```diff
@@ -145,28 +147,34 @@ public void run(Supplier<Element> elementSupplier) throws Exception {
...
-                // Auto-focus first focusable element if nothing is focused
-                if (focusManager.focusedId() == null && !focusManager.focusOrder().isEmpty()) {
-                    focusManager.setFocus(focusManager.focusOrder().get(0));
-                }
-
-                // Apply post-render processors (e.g., effects, overlays)
-                for (ToolkitPostRenderProcessor processor : postRenderProcessors) {
-                    processor.process(frame, elementRegistry, lastElapsed);
+                // Synchronize rendering to prevent concurrent access from scheduler thread
+                renderLock.lock();
```
maxandersen, line 151: shouldn't it just optimistically skip if someone already locked in the loop?

maxandersen:
I feel we are making this more complicated than it need to be - would a single queue of events not be better than having to sprinkle rentrant locks ?

melix:
If you go with a single queue of events you'll end up with the same problem as before, with key events blocking :/

maxandersen:
> If you go with a single queue of events you'll end up with the same problem as before, with key events blocking :/

sorry - I mean, that you would still have a separate thread to get the key events but it would then just push to event queue;
then there should never be need to have reentrant lock on draw/render loop afacis as only one thread would ever be in there.

melix:
Well as a follow-up to this I'm investigating how we can enforce a single UI render thread.

maxandersen:
replaced byt #121

PR #121, opened 2026-01-19
fix: Introduce proper UI thread

melix:
This commit reworks how we render, so that we guarantee that rendering is done only on the UI thread. Resize events, typically, which could be done asynchronously in a different thread, were causing problems, as well as scheduled events.

To fix this, this introduces the concept of a proper UI thread and UIRunnable. Automatic checks are performed to verify that rendering is only done on that thread, which lets us remove a few locks.

In addition, during testing I realized that the ToolkitEffects were not rendered at the right location after a resize, so I took advantage of this to fix it.

This is arguably an easier model to understand than the solution in #120.

Closes #120

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -104,7 +104,7 @@ class MyApp extends ToolkitApp {
 
 ### Threading Model
 
-- TamboUI uses a dedicated **render thread model** similar to JavaFX
+- TamboUI TUI framework uses a dedicated **render thread model** similar to JavaFX or Swing
 - All rendering and UI state modifications must happen on the render thread (the thread running `TuiRunner.run()`)
 - `RenderThread.isRenderThread()` checks if current thread is render thread; `RenderThread.checkRenderThread()` asserts it
 - `TuiRunner.runOnRenderThread(Runnable)` executes on render thread (immediately if already on it, queued otherwise)


docs/src/docs/asciidoc/api-levels.adoc
```diff
@@ -302,6 +302,86 @@ var config = TuiConfig.builder()
...
+==== The UI Thread
+
+When `TuiRunner.run()` starts, it marks the current thread as the UI thread.
+All event handling and rendering callbacks execute on this thread.
+When `run()` exits, the UI thread is cleared.
+
+[source,java]
+----
+// Check if on UI thread
+if (runner.isUiThread()) {
```
maxandersen: nit pick, but shouldn't it be `isUIThread()`?  or better yet - as UI is quite broad... isRenderThread() or isUIDispatchThread() ?

melix: I can live with `isRenderThread`

docs/src/docs/asciidoc/api-levels.adoc
```diff
@@ -302,6 +302,86 @@ var config = TuiConfig.builder()
     .build();
 ----
 
+=== Threading Model
+
+{project-name} uses a dedicated *render thread model* similar to other UI frameworks.
```
maxandersen, line 307: if we say "other UI frameworks" its best if we state which or at least give example. or just not mention there are other.

docs/src/docs/asciidoc/api-levels.adoc
```diff
@@ -302,6 +302,86 @@ var config = TuiConfig.builder()
...
+
+[source,java]
+----
+// Check if on render thread
+if (runner.isRenderThread()) {
+    // Safe to modify UI state
+}
+
+// Or use the static utility
+if (RenderThread.isRenderThread()) {
```
maxandersen, line 330: theoretically this would make it tricky to do multiple render threads for i.e. concurrent recording /replays of screens.... but i guess we can then change the tuirunner to know about that too if the need ariises :)

tamboui-tui/src/main/java/dev/tamboui/tui/RenderThread.java
```diff
@@ -0,0 +1,88 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.tui;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Utility class for render thread management.
+ * <p>
+ * TamboUI uses a dedicated render thread model similar to JavaFX. All rendering operations
```
maxandersen, line 12:
```suggestion
 * TamboUI TUI framework uses a dedicated render thread model similar to JavaFX or Swing. All rendering operations
```

AGENTS.md
```diff
@@ -96,20 +96,30 @@ class MyApp extends ToolkitApp {
...
-- Event types: `KeyEvent`, `MouseEvent`, `TickEvent`, `ResizeEvent`
+- Event types: `KeyEvent`, `MouseEvent`, `TickEvent`, `ResizeEvent`, `UiRunnable`
 - `KeyEvent` provides semantic methods (`isQuit()`, `isUp()`, `isDown()`, `isSelect()`, etc.) that respect configured bindings
 - Bindings map physical inputs to semantic actions; use `BindingSets.standard()`, `BindingSets.vim()`, or custom bindings
 - `@OnAction` annotation on Component methods handles actions; use annotation processor for compile-time generation
 - Toolkit elements handle events via `handleKeyEvent()`/`handleMouseEvent()` or handler lambdas
 
+### Threading Model
+
+- TamboUI uses a dedicated **render thread model** similar to JavaFX
```
maxandersen:
```suggestion
- TamboUI TUI framework uses a dedicated **render thread model** similar to JavaFX or Swing
```

maxandersen (approved):
Minor doc/precicions in the doc but overall good and I agree this is a much simpler model to reason about.

PR #122, opened 2026-01-20
feat: dark/light theme for asciinemaplayer

maxandersen:
this toggles the theme of asciinema player using css dark/light themes.

though not sure if good - as tamboui demos look to not use ansi colors in key places so some demos sees text stay white and not visible in light mode.

we should check if that happens in lightmode in terminal too- if yes, we shuold fix/find way to handle that.

### DIFF
--- a/docs/src/theme/tamboui.css
+++ b/docs/src/theme/tamboui.css
@@ -957,3 +957,78 @@ body.toc2.toc-left .prev-next-nav {
     background: rgba(30, 30, 46, 0.9);
     border-top: 1px solid var(--sl-color-hairline);
 }
+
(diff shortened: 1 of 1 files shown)


melix:
This works, however I'm not sure I would do it. One reason is that we have CSS styling demos and that it interferes (even if their tape files are wrong atm, they don't switch themes). The other reason is that as you have mentioned, some demos look weird with the new theme.

maxandersen:
@melix yes thats part of why I kept it separate. I do think though that this highlights its important we get the styling/theme mechanism done in a way that it is possible to blend in with ansiable themes ...AND be able to stay in your own color theme ...and our demos shouldn't mix and match as they seem to be doing now.

PR #124, opened 2026-01-21
fix: InlineDisplay should use backend

melix:
This commit updates InlineDisplay to that it renders using the backend. Before this, it was using System.out, which may be fine, but not for cases where typically the system out is captured, since in that case we wouldn't be able to display anything.


tamboui-core/src/main/java/dev/tamboui/inline/InlineDisplay.java
```diff
@@ -98,10 +99,36 @@ public static InlineDisplay create(int height) throws IOException {
      */
     public static InlineDisplay create(int height, int width) throws IOException {
         Backend backend = BackendFactory.create();
-        PrintWriter out = new PrintWriter(System.out, true);
+        PrintWriter out = createPrintWriter(backend);
```
maxandersen, line 116: shouldnt' we provide accessor that pass in Backend as not always case backend came from BackendFactory.

PR #126, opened 2026-01-21
feat: svg export of buffer

maxandersen:
first cut of svg export of a Buffer. port of textuals dump of console.

Not super happy of it being just a static operation - feels like we might want to add a SvgExportable interface onto Frame, Buffer and others ...ideas welcome.

added ctrl+s to CustomComponent for easy test.

Idea on this is that during asciinemacast/recorder we could use it to implement [Screenshot](https://github.com/charmbracelet/vhs#screenshot) to use in documentation.


maxandersen:
i've added screenshot ability which saves files relative to the .cast output file.

haven't figured out best way to move those files - any reason we dont just move/copy all of generated/ ?

maxandersen:
rebased with main - looks good to me. @melix works for you?

PR #127, opened 2026-01-21
refactor: Use type-safe properties everywhere

melix:
TL/DR: This PR sounds scary but it actually simplifies things a lot, by having a single concept for expressing styled properties, so that they go naturally with CSS. See the 2d commit, as this PR is built on top of another. The amount of code change is mostly related to shuffling things around and simplifications.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -140,6 +140,8 @@ class MyApp extends ToolkitApp {
 - You MUST use imports instead of fully qualified names in code
 - You SHOULD NOT name a method `getXXX` if it's not a simple getter returning a private field: prefer `computeXXX`, `fetchXXX`, `toXXX`, etc.
 - You MUST NOT add comments in source which cannot be understood without context
+- Tests SHOULD use the `BufferAssertions` test fixtures as much as possible
+- You CAN improve test fixtures to make tests more readable
 
 ## Testing Instructions
 
--- a/demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerDemo.java
+++ b/demos/filemanager-demo/src/main/java/dev/tamboui/demo/filemanager/FileManagerDemo.java
@@ -1,6 +1,9 @@
 //DEPS dev.tamboui:tamboui-toolkit:LATEST
 //DEPS dev.tamboui:tamboui-jline3-backend:LATEST
 //DEPS dev.tamboui:tamboui-image:LATEST
+//DEPS dev.tamboui:tamboui-panama-backend:LATEST
+
+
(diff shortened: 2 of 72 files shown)


docs/src/docs/asciidoc/developer-guide.adoc
```diff
@@ -209,6 +209,128 @@ try (var runner = ToolkitRunner.builder()
...
+public class MyGauge implements Widget {
+    // Define a custom property for the filled bar color
+    public static final PropertyDefinition<Color> BAR_COLOR =
+        PropertyDefinition.of("bar-color", ColorConverter.INSTANCE);
+
+    // ...
+}
+----
+
+This allows users to style your widget with CSS:
```
maxandersen, line 249: there is more code needed than just the declaration of the static final I assume?

docs/src/docs/asciidoc/styling.adoc
```diff
@@ -726,6 +726,151 @@ Override specificity with `!important`:
...
+    padding: 1;             /* NOT inherited - only Panel gets this */
+}
+
+/* Children need their own background if desired */
+Panel.sidebar Text {
+    background: black;      /* Must be set explicitly on Text */
+}
+----
+
+This matches standard CSS behavior: setting a background on a container doesn't automatically give all its children that same background.
```
maxandersen, line 880: color me surprised :)

...but css supports: 

background: inherited;

...will this be able to do that?

maxandersen: css also has transparent option but I reckon that doesn't really apply here as we dont (yet) have full layers of widgets?

tamboui-core/src/main/java/dev/tamboui/layout/BorderSet.java
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2025 TamboUI Contributors
  * SPDX-License-Identifier: MIT
  */
-package dev.tamboui.widgets.block;
+package dev.tamboui.layout;
```
maxandersen: hows Borders*.java related to layouts?

melix: The alternative is to keep this in the widget and don't make it part of standard properties.

maxandersen: Ok :)

docs/src/docs/asciidoc/developer-guide.adoc
melix: fixed

PR #128, opened 2026-01-21
fix: add customer borders to block demo

maxandersen:
I did this to test #123 so might as well share it as its a nice feature demo ;)


melix:
I like it! Maybe let's merge once (or if) #127 is merged, since it build on top.

melix:
uh? why was this closed and deleted?

melix:
Manually pushed.

PR #129, opened 2026-01-22
feat: add highlight using pygments

maxandersen:
I wanted to have a pure java highlighter but were not able to find one I was happy with (closest is/was textmate4e which I do think is doable but its dependency chain is massive so I wanted something that was lighter).

this adds a pygments.highlight(sourcetype, source) that converts into Text() with styles and tags. 

For now only done at lower levels, added demo and hook to Filemanager. 

note: it tries pygmentize, uvx, pipx in that order ...thus first call might be slow but afterwards its "just" a external process call "cost"

Could also be a candidate to be external repo - wdyt?

* note: I have a tree-sitter spike which is probably better more long term but it supports way less languages so stil wanted a basic highlight via pygment.


maxandersen:
hmm - not right diff...reopening soon

PR #130, opened 2026-01-22
feat: add highlight using pygments

maxandersen:
I wanted to have a pure java highlighter but were not able to find one I was happy with (closest is/was textmate4e which I do think is doable but its dependency chain is massive so I wanted something that was lighter).

this adds a pygments.highlight(sourcetype, source) that converts into Text() with styles and tags. 

For now only done at lower levels, added demo and hook to Filemanager. 

note: it tries pygmentize, uvx, pipx in that order ...thus first call might be slow but afterwards its "just" a external process call "cost"

Could also be a candidate to be external repo - wdyt?

* note: I have a tree-sitter spike which is probably better more long term but it supports way less languages so stil wanted a basic highlight via pygment.

### DIFF
--- a/tamboui-pygments/src/main/java/dev/tamboui/pygments/Pygments.java
+++ b/tamboui-pygments/src/main/java/dev/tamboui/pygments/Pygments.java
@@ -1,0 +1,528 @@
+/*
+ * Copyright (c) 2025 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.pygments;
+
+import dev.tamboui.style.Color;
+import dev.tamboui.style.Style;
+import dev.tamboui.style.Tags;
+import dev.tamboui.text.Text;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Syntax-highlights source code by invoking the {@code pygmentize} CLI with the {@code raw} formatter.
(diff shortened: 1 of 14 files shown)


tamboui-pygments/src/main/java/dev/tamboui/pygments/Pygments.java
```diff
@@ -0,0 +1,531 @@
...
+        /**
+         * @param tokenType full Pygments token type (e.g. {@code Token.Literal.String})
+         * @return a style to use for the given token type
+         */
+        Style resolve(String tokenType);
+    }
+
+    public static final TokenStyleResolver DEFAULT_STYLE_RESOLVER = createDefaultStyleResolver();
+
+    private static TokenStyleResolver createDefaultStyleResolver() {
```
maxandersen, line 405: atm just using colors i made up + tagging with "syntax-*". needs wiring up to whatever theming could be interesting.

davsclaus:
Very nice

I wonder if you have also looked at what jline - I think it has some basic java highlighting. And that you can also tweak colors for that.
Screenshot 2026-01-24 at 08 17 05

this is from camel-jbang and the edit command we have

camel edit foo.java

maxandersen:
i saw the highlight feature but didn't see many applications of it (i.e. different languages) ..got pointer to the code that setup the above?

maxandersen:
found the pr at https://github.com/apache/camel/commit/483f91ad8365f153f6f9b2078a3d7e4b107d59b1#diff-322a53fd0155c24a688e704ad9c27dae17302459a3548fe38f54dac99aadb004

this shows a nanorc config file format with regular expression matching that could be interesting....its only regular expression based but that could definitely be a "basic" option.

davsclaus:
yeah @gnodet would be the go to person about jline - i think it has highlighters out of the box for json/xml and other common styles.

maxandersen:
I see it at https://github.com/jline/jline3/blob/master/builtins/src/main/java/org/jline/builtins/SyntaxHighlighter.java

That has a lot of useful stuff but afaics the hooks to do this all translate to attributedStrings that doesn't have the type info afaics just the "naked" styles/colors. would be perfect to use if could have it somehow hooked up and get not just this string is red, but its red of class keyword or similar.

PR #131, opened 2026-01-22
feat: TUI and Toolkit modules for inline display

melix:
The inline display feature is nice, but lacked the higher level abstractions in TUI and toolkit. This made it effectively more complicated than it should be to use widgets or elements/components in the inline mode.

In the process, I have fixed an issue with focus management and some demo recordings. There are still bugs in the recording of the inline versions, but that can be fixed in another PR.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/internal/record/AnsiTerminalCapture.java
+++ b/tamboui-core/src/main/java/dev/tamboui/internal/record/AnsiTerminalCapture.java
@@ -4,68 +4,51 @@
  */
 package dev.tamboui.internal.record;
 
-import dev.tamboui.buffer.Buffer;
-import dev.tamboui.buffer.Cell;
-import dev.tamboui.layout.Rect;
-import dev.tamboui.style.AnsiColor;
-import dev.tamboui.style.Color;
-import dev.tamboui.style.Modifier;
-import dev.tamboui.style.Style;
-
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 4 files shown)


tamboui-core/src/main/java/dev/tamboui/internal/record/AnsiTerminalCapture.java
```diff
@@ -4,68 +4,51 @@
...
 import java.io.PrintStream;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
- * Captures System.out output by parsing ANSI escape sequences and maintaining
- * a virtual terminal buffer. Used for recording inline demos that write directly
- * to System.out rather than using Backend.draw().
+ * Captures raw bytes written to System.out for recording inline demos.
```
maxandersen, line 15: so...this is not reliant of anything in tamboui - it just works on raw java std out/err streams? thats nice.

tamboui-toolkit/demos/inline-toolkit-demo/src/main/java/dev/tamboui/demo/InlineToolkitDemo.java
```diff
@@ -0,0 +1,354 @@
...
+
+import static dev.tamboui.toolkit.Toolkit.*;
+
+/**
+ * NPM-style package installation demo using InlineApp.
+ * <p>
+ * This demo recreates the same multi-phase UI as InlineProgressDemo but using
+ * the toolkit's element-based DSL instead of manual widget composition.
+ * <p>
+ * Compare the ~400 lines of InlineProgressDemo with this ~200 line version
```
maxandersen, line 30: This is definitely nice :)

Am I right in assuming that inlineview height is static?

As in, if I want to have things scroll down - i.e. have the inline view be 3 lines high but then print a table to console and then have that  in the scroll buffer of terminal?

maxandersen (approved):

PR #132, opened 2026-01-22
fix: Tabs fit constraint and docs

melix:
This commit clarifies the documentation around layout and flex layout to make it clearer. In the process I found a bug in the tabs element, which didn't report its preferred width, breaking the layout.


melix:
Without the fix, the elements would always fill the available space, which made the flex layout inoperative.

maxandersen:
> Without the fix, the elements would always fill the available space, which made the flex layout inoperative.

make sense - jsut didnt spot a diff in the layout demo...is there a row that massages it ?

melix:
In the demo the components have fixed widths and didn't include nested components.

PR #136, opened 2026-01-23
fix: Wide chars unicode support

melix:
This commit fixes support for wide char characters (emojis, CJK, ...). The tables in `CharWidth` were checked against unicode.org.

I was a bit worried that this could introduce a performance regression, given the amount of work we have to do to check if characters are wide or not, but this seem to work relatively well. At least on my laptop I didn't notice a regression, all demos still run at requested frame rate.

The code was largely generated by Claude Code, with some requests from me to make `CharWidth` faster.

The result, however, largely differs based on which terminal is used. See for example the same demo:

Ghostty

image

Gnome Terminal

image

Rio

image

Hyper

image

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/inline/InlineDisplay.java
+++ b/tamboui-core/src/main/java/dev/tamboui/inline/InlineDisplay.java
@@ -322,6 +322,10 @@ public final class InlineDisplay implements AutoCloseable {
         try {
             // Move cursor to after the display area and restore cursor visibility
             backend.carriageReturn();
+            int toBottom = currentHeight - 1 - lastCursorY;
+            if (toBottom > 0) {
+                backend.moveCursorDown(toBottom);
+            }
             out.print(AnsiStringBuilder.RESET);  // Reset style
             backend.showCursor();
             out.print("\u001b[0 q");  // Reset cursor to default style


demos/unicode-inline-demo/src/main/java/dev/tamboui/demo/UnicodeDemo.java
```diff
@@ -0,0 +1,157 @@
+//DEPS dev.tamboui:tamboui-toolkit:LATEST
```
maxandersen, line 1: this demo is printing out unicode just fine but the cursor ends up inside the output ..not at the end. is that intentional/expected?

Image

davsclaus:
amazing to see the TUI power in 2026

Is there a theme for old classic green terminals with just upper case - That would take me back to my first job ;)

demos/unicode-inline-demo/src/main/java/dev/tamboui/demo/UnicodeDemo.java
melix: That seems to be specific to your terminal, I'm not seeing this behavior in Gnome Terminal, nor Ghostty :/

maxandersen:
we dont have full "theming" yet; on todo :)

maxandersen:
@graemerocher ok to merge? pretty please :)

PR #138, opened 2026-01-25
feat: add emoji replace to markup + default emoji set

maxandersen:
This brings back `:emojicode:` to MarkupParser with the diff Emoji.java is fallback EmojiResolver so if someone really wanted a different lookup or different set they could provide that.

Note @melix : the emoji-demo exposes what I think is a layout or width calculation bug where some ZWJ (zero width join) combinations does not render correctly before I resize/redraw. Try it out by typing in "greenland" and you'll see something like:

and then if i just resize one character:

voila - correct rendering.

I spent a god awful time thinking the characters were different for the two greenland emojis, until I added the codepoint column....they are the exact same...but rendered differently on first draw, but correct on second.

### DIFF
--- a/tamboui-toolkit/demos/emoji-demo/src/main/java/dev/tamboui/demo/EmojiDemo.java
+++ b/tamboui-toolkit/demos/emoji-demo/src/main/java/dev/tamboui/demo/EmojiDemo.java
@@ -141,13 +141,24 @@ public class EmojiDemo {
 
         try (var runner = ToolkitRunner.create(config)) {
             runner.run(() -> {
-                // Apply text filter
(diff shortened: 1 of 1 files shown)


maxandersen:
fyi:
btw. https://mitchellh.com/writing/grapheme-clusters-in-terminals is a good readmaxandersen: learned me about mode 2027 - which could be used to have emojicodes adapt to the terminal abilities.maxandersen: i.e. not return zwj

melix:
Actually it seems that there's a bug on both screenshots. In the first one there's a stale character on the second greenland line. On the 2d screenshot I feel like the white background should be applied to the whole line.

maxandersen:
thanks for the fixes and docs clarification - good stuff !

PR #140, opened 2026-01-25
feat: inherit property

### LINKED ISSUE #135: Consider making Overflow inheritable
Currently, you have to set `text-overflow: ...` on every element of a container (column, panel, ...) because Overflow is not inheritable: https://github.com/tamboui/tamboui/blob/225aa7360981e98cb43cd2c54dff59b84d7a0c41/tamboui-core/src/main/java/dev/tamboui/style/StandardProperties.java#L186

To me it would make sense to make it inheritable, wdyt?

melix:
This is a proposal to fix #135. This doesn't change the semantics of overflow and doesn't make it inherited by default. However, it introduces the `inherited` value, which, for a child, lets it pick the value from its parent (similar to CSS).

This alone, however, doesn't really address the original ticket, so I propose to add an `inheritable` property value which, when set, will change the semantics of a property from non-inherited to inherited.


maxandersen:
> This alone, however, doesn't really address the original ticket, so I propose to add an `inheritable` property value which, when set, will change the semantics of a property from non-inherited to inherited.

so that parents can force their will on children? that does feel like quite useful but also like something that can be over-used and make harder to control......also something I believe css deliberattly does not allow as it makes composition harder.

Like, sure - we can do it but it wouldn't be css "like" - we could also consider text-overflow to simply be inheritable and have different rules for text-overflow and overflow. The latter only affect text-overflow.

But how about instead, use structural selectos?

```css
#someid > * {
  text-overflow: ellipsis;
}
```

this is not parent inheritance (is not computed) but simply applied in case elements haven't applied.

you can of course still use Class:

```css
Panel > * {
  text-overflow: ellipsis;
}
```

and you are close to apply similar.

wdyt?

melix:
The parents cannot "force control" their children. It simply says "this is inherited unless a child says otherwise".

melix:
The reason I'm not super fan of structural selectors is that they tend to be messy and take more time to evaluate.

maxandersen:
But isn't structural selectors more or less the only effective way for users rather than widget writers be able to apply styles without having to know all the inner details?

I'm not inherently (no pun intended) against the idea of supporting "inheritable" property but do think structural selectors are useful constructs as allow fairly easy config of layout/colors from user space.

melix:
Maybe just me, but I'm already lost when there are more than 2 levels, e.g:

```
.foo Thing:selected {
   ...
}
```
vs
```
Thing:selected .foo {
...
}
```
so adding another level and I'm going crazy :D 
```
Thing:selected > * .foo {
...
}
```

maxandersen:
```
Thing:selected {
		text-overflow: word-wrap inheritable
		property-somethingelse: somevalue inheritable
}
```

vs

```
Thing:selected > * {
		text-overflow: word-wrap
		property-somethingelse: somevalue
}
```

The latter is something I can ask llm and google about ...the first one is custom and requires more text/typing 

So i'm not sure how crazy it is :)

anyhow - structured selection is superuseful for other things than mimicking inheritance so really a separate/different feature than this pr.

melix:
Right, just saying I'm not sure we should merge this if you think that this isn't the right way to go.

maxandersen:
yeah - my only main concern is that it would be interesting to grok why css in web doesn't do it this way...

melix:
Probably because of encapsulation, since it means a "deep" component is not in full control of how it renders? I can give the structural selector a try.

PR #141, opened 2026-01-25
Add mode 2027 support

melix:


tamboui-core/src/main/java/dev/tamboui/terminal/Mode2027Status.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Represents the status of Mode 2027 (grapheme cluster mode) support in a terminal.
+ * <p>
+ * Mode 2027 is a terminal feature that coordinates grapheme cluster handling between
+ * applications and terminals. When enabled, the terminal signals that it can handle
+ * complex grapheme clusters (ZWJ emoji sequences, regional indicators) as single
+ * display units.
+ */
+public enum Mode2027Status {
```
maxandersen, line 15: should we capture it as GraphmemeStatus rather than tied to specific protocol/code?

melix: Not sure I understand your comment: this _is_ tied to the support of a specific protocol?

melix:
> but who is intended to call it users or tamboui?

It is called internally by the backends during initialization.

PR #145, opened 2026-01-26
feat: Make base colors overridable

melix:
This commit makes it possible to override the default colors using a style engine (typically CSS). While it may sound weird, it allows for example overriding `red` to blue, etc. This is interesting for several use cases:

- accessibility, where we'd want users to change the default theme for daltonism or other visual deficiences
- theming, where we simply want to be able to override the styles, but even for widgets which use base ANSI colors

The goal of this commit is to avoid an independent "Theme" which would introduce a redundancy with the CSS styling. Instead, it tries to normalize regular styles by giving them names. For example, you can disable dimming by doing:

```
.dim {
   text-style: normal;
}
```

This works because default styles are mapped to style names.

Side note: if you use a named color on the right-hand side of the style, e.g:

```
   color: blue;
```

then `blue` doesn't refer to a style, but to a color.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/AnsiStringBuilder.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/AnsiStringBuilder.java
@@ -94,7 +94,10 @@ public final class AnsiStringBuilder {
      * @return the ANSI code string (without CSI prefix or 'm' suffix)
      */
     public static String colorToAnsiForeground(Color color) {
-        if (color instanceof Color.Reset) {
+        if (color instanceof Color.Named) {
+            // Named colors delegate to their default value
+            return colorToAnsiForeground(((Color.Named) color).defaultValue());
+        } else if (color instanceof Color.Reset) {
             return "39";
         } else if (color instanceof Color.Ansi) {
             AnsiColor c = ((Color.Ansi) color).color();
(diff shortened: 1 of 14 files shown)


tamboui-core/src/main/java/dev/tamboui/terminal/AnsiStringBuilder.java
```diff
@@ -116,7 +119,10 @@ public static String colorToAnsiForeground(Color color) {
      * @return the ANSI code string (without CSI prefix or 'm' suffix)
      */
     public static String colorToAnsiBackground(Color color) {
-        if (color instanceof Color.Reset) {
+        if (color instanceof Color.Named) {
```
maxandersen, line 122: any reason why color don't just implement this directly and remove need to do these instanceof's?

tamboui-panama-backend/src/main/java/dev/tamboui/backend/panama/PanamaBackend.java
```diff
@@ -383,7 +383,9 @@ private void applyStyle(Style style) {
     }
 
     private void appendColorToAnsi(Color color, boolean foreground) {
-        if (color instanceof Color.Reset) {
+        if (color instanceof Color.Named named) {
```
maxandersen, line 386: same as other places - should Color simply not just implement these directly or have a AnsiColor interface that does these?

maxandersen:
merging this as its awesome :) 

the color ansible is complementary / not specific to this pr so opened #150 to remember to consider or reject.

PR #148, opened 2026-01-26
feat: Add Columns element similar to Textual's

melix:
This commit introduces a Columns layout similar to the one found in Textual. All properties are configurable via CSS properties, including the column order, count and spacing.

This highlighted a bug in inline mode only with the Panama backend. A demo was added which does something similar to Textual's emoji list demo.

### DIFF
--- a/docs/src/docs/asciidoc/api-levels.adoc
+++ b/docs/src/docs/asciidoc/api-levels.adoc
@@ -529,7 +529,7 @@ public class MyApp extends ToolkitApp {
 }
 ----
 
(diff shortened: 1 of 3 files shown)


tamboui-widgets/src/main/java/dev/tamboui/widgets/columns/Columns.java
```diff
@@ -0,0 +1,296 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.widgets.columns;
```
maxandersen, line 5: what am I missing on why this (and layouts) isn't in core? 

whats the part that make it needed to be in widgets?

PR #155, opened 2026-01-27
fix: consistent exception handling

maxandersen:
this is draft for cleaning up exception handling.

Keeping draft as awaiting merge of all the layout and some questions of naming / approach to fix.

This introduces as a `TerminalException` which is a runtime exception.

This could also be named TambouiException - if objection we can go to that too - mainly just using it as a more descriptive exception than name of fwk exception.

Made the most obvious places use it directly and where already making speicic exceptions I just made TambouiException the base.

obvious win is a bunch of:

```
backend.onResize(() -> {
                try {
                    terminal.draw(this::render);
                } catch (IOException e) {
                    // Ignore
                }
            });
            ```
            
            is now just:
            
           ```
			backend.onResize(() -> {
                terminal.draw(this::render);
            });
           ```
            
Open questions:
1) is TerminalException good as base classname?
2) Should we wrap IllegalState and IllegalArgument exceptions ? (we use it in a lot of palces)
3) Should backend throw TerminalException rather than RuntimeException? (if yes, backends need to depend on core...which parts of panama does not yet)
4) should onResize wrap/ignore exceptions thrown? (all our code thus far ignored exceptions in resize)

### CODE REVIEW

**Comment by melix** [comment]
Maybe a base `TambouiException`, with a `RenderingException` subclass?

**Comment by maxandersen** [comment]
Yes I'm pro using specific types as it enables easier debugging (set breakpoint on specific type rather than generic one) 

I'm also fine with a base TambouiException but should avoid  new TambouiException and rather use specific types or add new ones when needed.

**Comment by maxandersen** [comment]
Another is IllegalLayoutException or similar which can be used in many of the places where IllegalArgument is used.

**Comment by melix** [comment]
You could make `TambouiException` abstract.

**Comment by maxandersen** [comment]
> You could make `TambouiException` abstract.

good idea - wdyt about IllegalArgumentException / IllegalStateException's ...I lean towards keep using IllegalArgumentException when its simple language/api breakages (pass negative % in or outofbounds value etc) thats literally what those exception types is for. for IllegalStateExcpetion  i think more tamboui/terminal specifci exceptions are usefull.

**Comment by melix** [comment]
> for IllegalStateExcpetion i think more tamboui/terminal specifci exceptions are useful

+1

**Comment by maxandersen** [comment]
pushed more cleanup.

Look if you think the naming / packages are meaningful/sane :)

File: docs/src/docs/asciidoc/core-concepts.adoc [hunk 1]
```diff
@@ -475,6 +475,17 @@ if (event instanceof TickEvent) {
 }
 ----
 
+=== Exceptions
+
+TamboUI tries to enforce a clear, consistent exception hierarchy for framework errors.
+
+* All general framework errors extend `TerminalException` (`dev.tamboui.TerminalException`). This is a `RuntimeException`.
```
melix: This seems to be out of date

tamboui-core/src/main/java/dev/tamboui/TamboUIException.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui;
```
melix: I would move the exceptions in a subpackage, because of Java modules, the split package stuff, it's better to have well defined packages rather than top level.

tamboui-core/src/main/java/dev/tamboui/errors/TerminalIOException.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Exception thrown when a terminal I/O operation fails.
+ * <p>
+ * This exception is used for errors that occur during terminal I/O operations,
+ * such as reading from or writing to the terminal, setting terminal attributes,
+ * or querying terminal state.
+ *
+ * @see TamboUIException
+ */
+public class TerminalIOException extends TamboUIException {
```
melix: I'm not sure I like the name much, but I have no better suggestion. Or, simply `UncheckedIOException`.

tamboui-core/src/main/java/dev/tamboui/TamboUIException.java
maxandersen: but which package would core exceptions 'fit' in? dev.tamboui.exceptions ?

melix: Yes, or `dev.tamboui.errors`

maxandersen: moved it.

tamboui-core/src/main/java/dev/tamboui/errors/TerminalIOException.java
maxandersen: alternatives I considered:

- TamboUIIOException - probably the rigth name, but just hard to read
- jdk UncheckedIOException - can't then catch TambouiException 
- tamboui UncheckedIOExcepton - very easy by accident throw or catch the wrong UnchekcedException
- RuntimeIOException - what UncheckedIOException should have been called imo :)

so keep TerminalIOException or use RuntimeIOException ?

melix: Agreed on `TamboUIIOException`. Alternatively, we could break the rule for readability and use `TambouiIOException`. Otherwise go with `RuntimeIOException`.

melix:
Just fix javadocs and we're good to go!

tamboui-core/src/main/java/dev/tamboui/errors/TerminalIOException.java
maxandersen: RuntimeIOException won

maxandersen:
@melix i also renamed errors->error as tui already had an error package.

PR #160, opened 2026-01-28
feat: initial aesh backend

maxandersen:
this adds a `aesh` backend - see https://aeshell.github.io/ 

aeash is on surface quite similar to jline but has quite clean and modular setup so at least in theory simpler/smaller :)

atm this depends on aesh-readline which drags in more than need for backend implementation be  but @stalep is going to refactor aesh to separate the tty terminal so we can depend on just that.

One of the nice things about aesh is that it has a terminal abstraction that allows serving the terminal out on http/websocket, ssh, etc..

Added basic demo for http/ssh. Work in progress

### DIFF
--- a/demos/aesh-ssh-http-demo/build.gradle.kts
+++ b/demos/aesh-ssh-http-demo/build.gradle.kts
@@ -7,6 +7,7 @@ description = "Demo showcasing TamboUI app accessible via SSH and HTTP/WebSocket
 demo {
     displayName = "Aesh SSH/HTTP Demo"
     tags = setOf("aesh", "ssh", "http", "websocket", "toolkit", "network")
+    internal = true
 }
 
 dependencies {
--- a/demos/aesh-ssh-http-demo/src/main/java/dev/tamboui/demo/aesh/AeshSshHttpDemo.java
+++ b/demos/aesh-ssh-http-demo/src/main/java/dev/tamboui/demo/aesh/AeshSshHttpDemo.java
@@ -11,6 +11,14 @@
  */
 package dev.tamboui.demo.aesh;
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import org.aesh.terminal.Connection;
+import org.aesh.terminal.http.netty.NettyWebsocketTtyBootstrap;
+import org.aesh.terminal.ssh.netty.NettySshTtyBootstrap;
+
 import dev.tamboui.backend.aesh.AeshBackend;
 import dev.tamboui.style.Color;
 import dev.tamboui.toolkit.app.ToolkitApp;
@@ -21,13 +29,6 @@ import dev.tamboui.tui.TuiConfig;
 import dev.tamboui.tui.event.KeyEvent;
 import dev.tamboui.tui.event.MouseEvent;
 import dev.tamboui.widgets.input.TextAreaState;
-import org.aesh.terminal.Connection;
-import org.aesh.terminal.ssh.netty.NettySshTtyBootstrap;
-import org.aesh.terminal.http.netty.NettyWebsocketTtyBootstrap;
-
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 
 import static dev.tamboui.toolkit.Toolkit.*;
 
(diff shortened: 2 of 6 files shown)


tamboui-aesh-backend/src/main/java/dev/tamboui/backend/aesh/AeshBackend.java
```diff
@@ -0,0 +1,425 @@
...
+
+        outputBuffer.append("m");
+    }
+
+    /**
+     * Returns the underlying TerminalConnection for advanced operations.
+     *
+     * @return the TerminalConnection instance
+     */
+    public TerminalConnection terminalConnection() {
```
melix: Is this actually used? Because since we're exposing `Backend` in any case, the interface, seems users wouldn't be able to call that anyway.

maxandersen: For now no. Its more if some want to utilise the aesh features to do terminal readline/commands that all starts with a connection they can get it. 

It should actually just return Connection.

tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
```diff
@@ -190,7 +190,7 @@ public static TuiRunner create() throws Exception {
      * @throws Exception if terminal initialization fails
      */
     public static TuiRunner create(TuiConfig config) throws Exception {
-        Backend backend = BackendFactory.create();
```
maxandersen, line 191: @melix this is to allow user to setup the backend - needed for ssh/http from aesh as needs context the backendfactory dont have.

let me know if works for you or you can think of better approach.

demos/aesh-ssh-http-demo/src/main/java/dev/tamboui/demo/aesh/AeshSshHttpDemo.java
```diff
@@ -0,0 +1,190 @@
...
+ *   <li>An HTTP/WebSocket server on port 8080</li>
+ * </ul>
+ * <p>
+ * Connect via SSH:
+ * <pre>{@code ssh -p 2222 user@localhost}</pre>
+ * <p>
+ * Connect via WebSocket:
+ * Open http://localhost:8080 in a browser with WebSocket terminal support
+ */
+public class AeshSshHttpDemo implements java.util.function.Consumer<Connection> {
```
maxandersen, line 50: @melix this is rushed code but lets you see how it works.

run this and you can ssh / http into the app.

I haven't done anything to separate state or created multiple apps - in theory that sohuld be possible.

tamboui-tui/src/main/java/dev/tamboui/tui/TuiRunner.java
melix: Looks good to me. The fact that we ignore the backend in the config would sound like a bug anyway.

demos/aesh-ssh-http-demo/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+dependencies {
+    implementation(projects.tambouiToolkit)
+    implementation(projects.tambouiAeshBackend)
+    implementation(libs.aesh.terminal.ssh)
+    implementation(libs.aesh.terminal.http)
+    
+    // Apache SSHD for SSH server
+    implementation("org.apache.sshd:sshd-core:2.14.0")
```
melix: :x: Should use the version catalog

demos/aesh-ssh-http-demo/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+    implementation("org.apache.sshd:sshd-netty:2.14.0")
+    
+    // Netty for HTTP/WebSocket server
+    implementation("io.netty:netty-all:4.1.81.Final")
```
melix: :x: Should use the version catalog

melix:
Something that's unclear to me is how this will play with https://github.com/tamboui/tamboui/pull/184. Since the demo brings Netty and that the new aesh backend will be included in the fat jar, all demos will pick a random backend, probably, but also the jar will jump from 8MB to much more.

maxandersen:
@melix about #184 ...yeah, I would for now leave it out of the aggregation?

maxandersen:
i marked the demo as internal as demo tape not relevant here too.

merging as the backend in tuiconfig is needed for other prs waiting.

PR #163, opened 2026-01-29
feat: Add Spinner widget

melix:
Add Spinner widget (and SpinnerElement to toolkit) for displaying animated loading spinners. This spinner offers a number of default styles but also supports custom frame sets, similarly to custom border styles for blocks.

Styles and framesets are also configurable from CSS.


graemerocher (approved):

docs/src/docs/asciidoc/widgets.adoc
```diff
@@ -473,6 +477,83 @@ Calendar calendar = Calendar.builder()
...
+}
+
+/* spinner-frames overrides spinner-style when both are set */
+.fancy-spinner {
+    spinner-style: dots;
+    spinner-frames: "*" "+" "x" "+";  /* This takes precedence */
+}
+----
+
+NOTE: Programmatic styles set via `spinnerStyle()`, `frameSet()`, or `frames()` take precedence over CSS styles.
```
maxandersen, line 553: is that same for other things? like colors? ...can users call setXYZ(null) to let css control it?

...that said - why not have css have highest precdence?

melix: That's indeed the same for all widgets. We could probably change the precedence.

PR #164, opened 2026-01-29
use minimal spotless config to sort/cleanup imports + basic format 

maxandersen:
This pr enables spotless to sort and cleanup imports + minimal java formatting using eclipse formatter.

i added .xml that does minimal to just have consistent width and it does affect html indention.

Maybe we should combine this with the #117 and avoid that problem completely.

If ok, I would add format check to build too that fails it but still completes the build/test so PR's does not need perfect formatting before push to main.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -27,9 +27,10 @@ jobs:
       - name: Set up JBang
         uses: jbangdev/setup-jbang@main
 
+      - name: Formatting check
+        run: ./gradlew --no-daemon spotlessCheck
       - name: Execute Gradle build
-        run: ./gradlew build
-
+        run: ./gradlew build -x spotlessCheck
       - name: Update jbang-catalog.json to include new demos
         run: |
          ./gradlew publishToMavenLocal


melix (approved):

maxandersen:
eh...not sure why gradle is just dumping a NPE on us in github :/

melix:
I can reproduce locally. So the thing is that your spotless changes broke the expectation. Now the module-info.java file for jline shows:

```java
@SuppressWarnings({"requires-transitive-automatic",
        "requires-automatic"}) module dev.tamboui.jline_three.backend {
```
with suppress warnings on a single line ( :vomiting_face: ). We can fix the `readModuleName` method to account for this (should we?), but even so, it seems that the PR broke more things, e.g after you'd get:

> /home/cchampeau/DEV/tamboui/tamboui-annotations/src/main/java11/module-info.java:13: error: package is empty or does not exist: dev.tamboui.annotations.bindings

Last but not least, spotless is a memory hog, so we'll need to update the memory settings in `gradle.properties` to `2g` probably.

maxandersen:
> with suppress warnings on a single line ( 🤮 ). We can fix the `readModuleName` method to account for this (should we?),

I'll see if i can get it to insert a new line. it does that in other places.

readModuleName should be fixed though as it is valid java.

> /home/cchampeau/DEV/tamboui/tamboui-annotations/src/main/java11/module-info.java:13: error: package is empty or does not exist: dev.tamboui.annotations.bindings

how do you get that one?

maxandersen:
ok, so redid this with a little bit of llm help.

1) fixed readmodulename to be more resilient 
2) changed order of compile so main is built before java11 as java11 requires the classes to be present for those packages to be found (not  fully sure why this only start failing now - but llm said:

"That’s because the old regex was accidentally grabbing the first “module …” it saw inside the Javadoc header of module-info.java (e.g. “This module provides…”), so readModuleName() could become "provides" instead of dev.tamboui.core. Then --patch-module provides=... patches the wrong module, and javac sees zero of your main packages → every exports looks “empty”."

so @melix if you can confirm the patch does what its still supposed to that would be great ;)

3) `<pre>{@code` can't contain other @something syntax to keep javac and javadoc happy so fixed those by using `&#64;` to allow for stable formatting.

haven't figured out yet why module and annotations stay on same line - but i'll keep trying though would prefer we get this in so things at least get sorted/stays consistent even if we change the formatting rules later.

maxandersen:
working on rebasing with main cleanly

maxandersen:
rebased so take a look

melix:
I'll do when I can. Power outage since a few hours here and return planned for noon tomorrow!

buildSrc/src/main/kotlin/dev.tamboui.java-library.gradle.kts
```diff
@@ -25,6 +25,10 @@ val java11 by sourceSets.creating {
 }
 
 tasks.named<JavaCompile>("compileJava11Java") {
+    // The module descriptor exports packages from the main sources, so those
+    // classes must be compiled before we compile module-info.java.
+    dependsOn(tasks.named("compileJava"))
```
melix, line 30: :x: Should be `mustRunAfter`. That said, this shouldn't be required at all 🤔

maxandersen: yeah i was confused too - remove it and module-info gets errors :/

melix: Actually I think the dependsOn is fine. The reason is because of the use of patch module.

maxandersen: ok, I'll merge then :)

maxandersen:
merging - atm split out spotlesscheck from main build and saved ~30 seconds. will open separate pr that spins of the format check in parallel run that should make the spotless overhead get close to zero.

PR #165, opened 2026-01-29
feat: Add Tree element

melix:
This commit introduces a tree element, similarly to the the list element. It belongs to the toolkit level because it needs to know about child element dimensions to render them properly.

The tree supports separation of model and view and lazy rendering. It is possible to display any kind of Element as node, using the `nodeRenderer(...)` call.

As usual, everything is configurable using CSS properties.


maxandersen:
cool - as i always i'm suspicious when widgets need to be toolkit level only..but that said:

nice addition - but rendering looks a bit weird for me:

```
 ▼ 📁 src                                                                                                                                  
│  │   ├── ▶ 📁 main                                                                                                                                                                          
│  │   └── ▼ 📁 test                                                                                                                                                                   
│  │       └── ▼ 📁 java                                                                                                                                                                                    
│  │           ├──   ☕ AppTest.java                                                                                                                                                                
│▶ │           └──   ☕ RouterTest.java   
```
     
how come child nodes has an 'empty' cell in front of their main icon?

maxandersen:
...and really would be nice to have a way to move the "splitters" in our layouts ;)

maxandersen:
question - particular reason to use TreeNode over TreeModel (similar to swing?)

maxandersen:
also....this tree element without children == List?

melix:
Can you share a screenshot? Not sure if GH renders that properly. Regarding `TreeNode`: I don't remember Swing, I actually don't think I ever built a Swing UI :D

melix:
> also....this tree element without children == List?

Not quite. But we could argue that a List is a Map where you don't care about the values :)

> as i always i'm suspicious when widgets need to be toolkit level only

That's a legitimate question. But I don't see how we could work around here: we have to know the dimensions of the children to tell the widgets "where" to render. But wouldn't be the first time I'm proved wrong :)

melix:
So, thought more about this, and I think we could provide a lower level `TreeWidget`, which renders `Widgets` as node, at the condition that the renderer tells which height it should occupy. Alternatively we'd need to introduce a `SizeableWidget` or something similar, which does something similar to `Element`, returning preferred width/height. However, that would defeat the simplicity of writing widgets today.

maxandersen:
> So, thought more about this, and I think we could provide a lower level `TreeWidget`, which renders `Widgets` as node, at the condition that the renderer tells which height it should occupy. Alternatively we'd need to introduce a `SizeableWidget` or something similar, which does something similar to `Element`, returning preferred width/height. However, that would defeat the simplicity of writing widgets today.

that makes sense to me (that just controls the area to render in for kids)

maxandersen:
> Can you share a screenshot? Not sure if GH renders that properly. 

image

I place cursor where there is that "extra" space which I assume is it drawing "no children" part.

> Regarding `TreeNode`: I don't remember Swing, I actually don't think I ever built a Swing UI :D

Swing has TreeModel https://docs.oracle.com/javase/8/docs/api/javax/swing/tree/TreeModel.html AND TreeNode https://docs.oracle.com/javase/8/docs/api/javax/swing/tree/TreeNode.html - I believe TreeNode is implemented using a TreeModel... TreeNode is the simplest and "just give me a tree" but the nice thing with TreeModel is that you can have arbitrary deep models as the widget itself doesn't manage any state and just look up based on objects. TreeModel (and TableModel) one of the real good things from swings approach imo

melix:
> that makes sense to me (that just controls the area to render in for kids)

What makes sense precisely?

maxandersen:
> > that makes sense to me (that just controls the area to render in for kids)
> 
> What makes sense precisely?

sorry - to have simplified treewidget where children are widgets expected to render within stated areas.

melix:
@maxandersen addressed review feedback

tamboui-core/src/main/java/dev/tamboui/layout/tree/TreeModel.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.layout.tree;
```
maxandersen: this isn't a layout is it, its a model? ..I kinda expected it to be in tamboui-widgets, under dev.tamboui.widgets.tree somewhere?

maxandersen (approved):
nice and clean. 

TreeNode implementing TreeModel hurts my old swing mind but I can't point to concrete issue ..and if we spot one I thinkj easy to migrate/move over to separation...but letes see real usage/feedback making it a problem first :)

PR #167, opened 2026-01-29
Fixes after discovering dock issues

melix:
This builds on top of #166 which revealed several issues. See 2d commit for details (1st one is the PR it builds on top of).

This also made me realize that our form abstraction is very basic, so a follow up PR with proper form handling is coming.

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/GaugeElement.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/GaugeElement.java
@@ -10,6 +10,7 @@ import dev.tamboui.layout.Rect;
 import dev.tamboui.style.Color;
 import dev.tamboui.style.Style;
 import dev.tamboui.terminal.Frame;
+import dev.tamboui.text.CharWidth;
 import dev.tamboui.widgets.block.Block;
 import dev.tamboui.widgets.block.BorderType;
 import dev.tamboui.widgets.block.Borders;
@@ -180,7 +181,7 @@ public final class GaugeElement extends StyledElement<GaugeElement> {
     @Override
     public int preferredWidth() {
         // Label width + some space for the gauge bar
-        int labelWidth = label != null ? label.length() : 4; // "75%" default
+        int labelWidth = label != null ? CharWidth.of(label) : 4; // "75%" default
         return labelWidth + 10;
     }
 
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/Panel.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/Panel.java
@@ -514,7 +514,7 @@ public final class Panel extends ContainerElement<Panel> {
                 height += ((Constraint.Length) c).value();
             } else {
                 // Use preferredHeight for proper calculation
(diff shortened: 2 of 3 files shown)


graemerocher (approved):

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
```diff
@@ -903,7 +904,7 @@ public Rect renderedArea() {
 
     @Override
     public EventResult handleKeyEvent(KeyEvent event, boolean focused) {
-        if (focused && keyHandler != null) {
+        if (keyHandler != null) {
```
maxandersen, line 907: is this intentional ? that elements will now *always* handle key events no matter if focused or not?

would it make sense to have handleUnfocusedKeyEent () to avoid elements by accident "swallowing" keys?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/GaugeElement.java
```diff
@@ -177,6 +177,19 @@ public GaugeElement borderColor(Color color) {
         return this;
     }
 
+    @Override
+    public int preferredWidth() {
+        // Label width + some space for the gauge bar
+        int labelWidth = label != null ? label.length() : 4; // "75%" default
```
maxandersen: should this not use CharWidth?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/ScrollbarElement.java
```diff
@@ -290,6 +290,18 @@ public ScrollbarElement endStyle(Style style) {
         return this;
     }
 
+    @Override
+    public int preferredWidth() {
+        // Scrollbar is 1 cell wide for vertical orientation
+        return orientation.isVertical() ? 1 : (state != null ? state.contentLength() : 10);
```
maxandersen: what happens if the content is big/huge...?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/Panel.java
```diff
@@ -504,14 +504,18 @@ private int computeContentHeight() {
...
+        // Children height: sum of child heights using preferredHeight
         for (Element child : children) {
             Constraint c = child.constraint();
             if (c instanceof Constraint.Length) {
                 height += ((Constraint.Length) c).value();
             } else {
-                // Default: 1 row per child
-                height += 1;
+                // Use preferredHeight for proper calculation
+                int preferred = child.preferredHeight(estimatedWidth, null);
```
maxandersen: should this not be child.preferredHeight(estimatedWidth) or child.preferredHeight(estimatedWidth, RenderContext.empty()) ?

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
melix: That's intentional and aligns with the javadocs declared behavior. That's actually one thing I fixed using the tracer :)

PR #169, opened 2026-01-29
feat: Add event handling tracing

melix:
This commit introduces a mechanism to trace and debug event handling in toolkit. This is because sometimes, the apps don't behave as we expect, and it's difficult to understand why, for example, a key event is not handled, or, more interestingly, what component decided to handle an event in place of another that you expected.

This also adds test fixtures so that we can test the behavior of an application in that regards.

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/event/EventTracer.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/event/EventTracer.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.toolkit.event;
+
+import dev.tamboui.tui.event.Event;
+
+/**
+ * Interface for tracing event routing decisions.
+ * <p>
+ * Implementations log detailed information about how events are routed
+ * through the element tree, which is useful for debugging focus and
+ * event handling issues.
+ * <p>
+ * Enable tracing by setting the {@code TAMBOUI_EVENT_TRACE} environment
+ * variable to a file path:
+ * <pre>
+ * TAMBOUI_EVENT_TRACE=/tmp/trace.log ./run-demo.sh my-demo
+ * </pre>
+ * <p>
+ * The trace file uses JSON Lines format (each line is a complete JSON object):
+ * <pre>
+ * {"ts":"2026-01-29T10:00:00.123","type":"route_start","event":"KeyEvent[TAB]","focused":"input1","elements":3}
+ * {"ts":"2026-01-29T10:00:00.124","type":"candidate","id":"input1","phase":"focused","decision":"tried"}
+ * {"ts":"2026-01-29T10:00:00.125","type":"route_end","event":"KeyEvent[TAB]","result":"HANDLED"}
+ * </pre>
+ *
+ * @see FileEventTracer
+ * @see EventTracerFactory
+ */
+public interface EventTracer extends AutoCloseable {
+
+    /**
+     * A no-op tracer that does nothing.
+     */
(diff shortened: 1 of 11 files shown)


graemerocher (approved):

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/event/EventTracer.java
```diff
@@ -0,0 +1,148 @@
...
+ * <pre>
+ * {"ts":"2026-01-29T10:00:00.123","type":"route_start","event":"KeyEvent[TAB]","focused":"input1","elements":3}
+ * {"ts":"2026-01-29T10:00:00.124","type":"candidate","id":"input1","phase":"focused","decision":"tried"}
+ * {"ts":"2026-01-29T10:00:00.125","type":"route_end","event":"KeyEvent[TAB]","result":"HANDLED"}
+ * </pre>
+ *
+ * @see FileEventTracer
+ * @see EventTracerFactory
+ */
+public interface EventTracer extends AutoCloseable {
```
maxandersen, line 32: couldn't this be at tui level?

melix: The tracer is specifically for debugging event *routing* through Elements and focus management, which are toolkit-level concepts. The tui layer only handles raw terminal I/O and doesn't know about Elements, FocusManager, or how events get dispatched to handlers.

If we wanted tui-level tracing (e.g., raw key codes from the terminal), that would be a separate concern. This tracer answers questions like "why didn't my button receive the click?" or "why did Tab go to the wrong element?" - all toolkit routing logic.

maxandersen: I agree and sorry for not being clearer - I wasn't saying to move the specifics of this down but that general notion of a "EventTracer" for TUI's makes sense to have...which the various layers then can submit things to.

I'm after that the default implmeention of such tracer would just use java logging to push them to and one could then use what ever logging fwk one have ...but also wanting to have a tamboui-dev piece which goes and connect to localhost:888444 (if you can figure out why those digits you are cool :) 
which then can provide a tui that provides nice ui for these and might know a bit more than just dumping strings of text.

melix: Right. I should have made it clear that this is the minimal work that saved hours of debugging. I'm fine if we expand that, it's just that debugging was driving me insane.

maxandersen: yeah so definitely +1 on getting these things tracable. 

just suggesting we add the TambouiTrace "sink" in lower levels which EventTracer with higher level apis on can delegate to for submitting them.

melix: To be honest I'm unclear what you want here. What would an `TambouiTrace` be? Wouldn't you end up implementing a logging framework?

maxandersen: its close, but similar with this here - could "just" be done via logging - the only difference is you have type safety to reason a bit better about it.

i.e. TuiRunner has the eventQueue ...so why is the Tracer not made available there and higher up in toolkit submit the higher level "messages"/events through it?

melix: Well it's not precisely because the event queue because if something blocks the event queue you'd get no trace. The point is to be able to debug what's going on when something fails to behave like we think it should.

maxandersen:
replaced by jfr

PR #171, opened 2026-01-30
feat: Form abstraction with validation and submission

melix:
The "Forms" demo was showing a form, but it was not doing anything particularly interesting. Worse, it showed that we had little to no abstraction on what a form is.

This commit fixes the problem by adding a proper form abstraction level. This starts with improving the widgets to include more form types (checkboxes, select, toggles) with customizable styles.

Then a FormState object coordinates the form content state and validation state. The `Validator` class offers a number of built-in validators, but others can be plugged in.

On top of that, the Toolkit module offers a higher level construct for form creation and submission.

### DIFF
--- a/demos/toolkit-export-demo/build.gradle.kts
+++ b/demos/toolkit-export-demo/build.gradle.kts
@@ -1,0 +1,18 @@
+plugins {
+    id("dev.tamboui.demo-project")
+}
+
+description = "Toolkit export demo: uses Element.renderedArea() for export regions. Press E to export."
+
+demo {
+    displayName = "Export (Toolkit)"
+    tags = setOf("export", "toolkit", "renderedArea", "crop", "svg", "html")
+}
+
+dependencies {
+    implementation(project(":tamboui-toolkit"))
+}
+
+application {
+    mainClass.set("dev.tamboui.demo.ToolkitExportDemo")
+}
--- a/demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
+++ b/demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
@@ -1,0 +1,226 @@
+//DEPS dev.tamboui:tamboui-toolkit:LATEST
+//DEPS dev.tamboui:tamboui-jline3-backend:LATEST
+/*
+ * Copyright (c) 2026 TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.demo;
+
+import dev.tamboui.buffer.Buffer;
+import static dev.tamboui.export.ExportRequest.export;
+import dev.tamboui.layout.Rect;
+import dev.tamboui.style.Color;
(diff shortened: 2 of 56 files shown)


demos/form-table-demo/src/main/java/dev/tamboui/demo/layout/FormTableDemo.java
```diff
@@ -29,22 +33,34 @@
...
     private static final int LABEL_WIDTH = 14;
     private static final AtomicInteger TAB_INDEX = new AtomicInteger(0);
     private static final TableState TABLE_STATE = new TableState();
 
-    // Persistent text input states for form fields
-    private static final TextInputState FULL_NAME = new TextInputState("Ada Lovelace");
-    private static final TextInputState EMAIL = new TextInputState("ada@analytical.io");
-    private static final TextInputState ROLE = new TextInputState("Research");
-    private static final TextInputState TIMEZONE = new TextInputState("UTC+1");
-    private static final TextInputState THEME = new TextInputState("Nord");
```
maxandersen, line 42: this seem to react to up/down no matter what field has focus.

demos/form-table-demo/src/main/java/dev/tamboui/demo/layout/FormTableDemo.java
```diff
@@ -29,22 +33,34 @@
...
+            // Security fields (boolean + text)
+            .booleanField("twoFa", true)
+            .textField("sessions", "3 active")
+            .build();
+
+    // Validation demo form state - validation results are stored here, not in FormFieldElement
+    private static final FormState VALIDATION_FORM = FormState.builder()
+            .textField("username", "")
+            .textField("email", "")
+            .textField("password", "")
```
maxandersen: should this not us Masked?

maxandersen:
is "save"/"cancel" in lower right supposed to work?

melix:
No it's placeholders, that were present and doing nothing before either ;)

demos/form-table-demo/src/main/java/dev/tamboui/demo/layout/FormTableDemo.java
melix: Maybe it would make sense to have a `password` type for this?

maxandersen: not everything that needs masking is a password() - we have a Masked widget already so maskedTextField(...) ?

or rather why not .text() .maskedText() ...is it important to say field?

melix: Or it could be a modifier on `.textField`. The `Field` is to avoid confusion with `text` used in other places. A `text` is immutable, a field is.

demos/form-table-demo/src/main/java/dev/tamboui/demo/layout/FormTableDemo.java
melix: Fixed in b5e95be

maxandersen:
looks good!

PR #172, opened 2026-01-30
feat: Add CSS style grid layout

melix:
As discussed, this adds the ability to define areas where to put widgets in the grid layout. The inspiration is CSS grid-template-area.

The API has been udpated so that you cannot use both `children` and `area` modes: you have to choose. It is possible to define the areas directly in CSS, however the components must be inserted using code.


melix:
I don't know if we should remove others. The dock one may look similar, but not having it would mean you'd have to redefine the grid template everywhere, when just using `.top(...)` is more natural.

PR #173, opened 2026-01-30
feat: lumis4j (rust->wasm/chicory) syntax highlighter

maxandersen:
very much a POC/draft showing using lumis4j which is brand new java port of lumis (rust) compiled to wasm and then built via chicory to jvm so we can use it as is.

currently its done by passing language guess + content to lumis4j which returns html with span class html which we then transform into bbcode which then gets very naively styled.

@andreaTP is looking into if we can get a more direct translation.

there is in theory 70 languages supported this way but not all work yet (project just existed in less than 24 hrs :)

### DIFF
--- a/tamboui-lumis4j/src/main/java/dev/tamboui/lumis4j/Lumis4jMarkup.java
+++ b/tamboui-lumis4j/src/main/java/dev/tamboui/lumis4j/Lumis4jMarkup.java
@@ -45,6 +45,7 @@ public final class Lumis4jMarkup implements AutoCloseable {
     m.put("htm", Lang.HTML.value());
     m.put("scss", Lang.SCSS.value());
     m.put("yml", Lang.YAML.value());
+    m.put("md", Lang.MARKDOWN.value());
     m.put("mdx", Lang.MARKDOWN.value());
     m.put("bash", Lang.BASH.value());
     m.put("h", Lang.C.value());
--- a/tamboui-lumis4j/src/test/java/dev/tamboui/lumis4j/Lumis4jMarkupTest.java
+++ b/tamboui-lumis4j/src/test/java/dev/tamboui/lumis4j/Lumis4jMarkupTest.java
@@ -4,13 +4,12 @@
  */
 package dev.tamboui.lumis4j;
 
-import java.beans.Transient;
 import java.time.Duration;
 
-import dev.tamboui.text.MarkupParser;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
+import dev.tamboui.text.MarkupParser;
 import dev.tamboui.text.Text;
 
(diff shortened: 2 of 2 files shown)


melix:
wow, how many levels of indirection do we have here? :D

andreaTP:
Sent an update: https://github.com/maxandersen/tamboui/pull/12

Among the others:
- moved the `bbcode` formatter [down to Rust](https://github.com/roastedroot/lumis4j/blob/main/wasm-build/src/bbcode.rs)
- set the [compilation target](https://github.com/roastedroot/lumis4j/blob/a0936055760610c57233ca93f06e487bcf923214/pom.xml#L47) to Java 8, I think this is the first time where I attempt this with Chicory, let me know if you notice anything wired
- enriched the API to enable better re-use of the Lumis instance
- enabled and embedded all the available languages ( the final resulting jar is ~10Mb now :sweat_smile: )

andreaTP:
> wow, how many levels of indirection do we have here? :D

Fair :slightly_smiling_face: I moved the bbcode converter down, closer to Tree Sitter, to remove at least one layer.

FWIW a working alternative is to [wrap `highlight.js`](https://github.com/andreaTP/highlightjs4j/) running on QuickJs, it works, but a few downsides:
 - seems much slower (I expect, even with all optimizations on, at least 2X slower than lumis)
 - depends on an "extra" runtime
 - we'd be back at converting html to bbcode

maxandersen:
Cool - not following the value of rust doing bbcode parsing ?

andreaTP:
Just directly emitting bbcode instead of translating from html, can remove if you don't see value, not fussy

maxandersen:
Ah. So generating bbcode not parsing ? That makes sense. 

Is it using class names like keyword, number etc. rather than specific color?

In any case definitely trying it out

andreaTP:
This is the code: https://github.com/roastedroot/lumis4j/blob/main/wasm-build/src/bbcode.rs
You are the main user of it, so feel free to make any change you like :+1:

maxandersen:
it works and I noticed no delays. pygral was 2-4s, here its more or less instant.

excllent!

maxandersen:
@andreaTP in this:

```java
LumisResult result = lumis.highlight(
            source,
            Theme.GITHUB_DARK,
            lumisLang,
            Formatter.BBCODE
            );
```
            
            Theme.GITHUB_DARK is irrelvant, right?

andreaTP:
@maxandersen yup, the API was still rough, this seems better: https://github.com/maxandersen/tamboui/pull/13

maxandersen:
found that we have a quote bug somewhere. test will fail for now.

use of [] in actual text (i.e.) json severely confuses markupparser...need to see where the isseue is on on bbcode side or markup issue

PR #174, opened 2026-02-01
feat: fluent export api with svg, html and text options

### LINKED ISSUE #152: clean up theme/svg exporter
right now svgexporter is separate class that inside has a Theme class - that one probably should go away and be a styleEngine or styleResolver that can be wired up rather than separate notion.

maxandersen:
just take snapshot - format decided based on filename:

buffer.export().toFile("default.html");

format specific settigns:
buffer.export().html().option(o->inlineStyles(true)).toFile("my.html");

cusotm format (user provded PDF format):
buffer.export().as(PDF).options(...).toFile('my.pdf");

This also enables to have screenshot in recorder to save to html/ansi/ascii in addition to svg.

Fixes #152 

@melix not happy with introduction of "Theme" in this (was internal before) - its really just here to lookup foreground/background colors when nothing set in buffer.

What could we introduce or reuse to have a "Theme" that css engine can be provider of but we can still reason about in core?

### DIFF
--- a/demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
+++ b/demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
@@ -57,11 +57,11 @@ public final class ToolkitExportDemo {
     /**
      * Entry point. Optional first argument: output directory for exported files.
      *
-     * @param args optional output path; default is current directory
+     * @param args optional output path; default is a temp directory
      * @throws Exception if export or I/O fails
      */
     public static void main(String[] args) throws Exception {
-        Path outDir = args.length > 0 ? Paths.get(args[0]) : Paths.get(".");
+        Path outDir = args.length > 0 ? Paths.get(args[0]) : Files.createTempDirectory("tamboui-export-");
         ToolkitExportDemo app = new ToolkitExportDemo(outDir);
 
         var config = TuiConfig.builder()
--- a/tamboui-core/demos/export-demo/src/main/java/dev/tamboui/demo/ExportDemo.java
+++ b/tamboui-core/demos/export-demo/src/main/java/dev/tamboui/demo/ExportDemo.java
@@ -56,12 +56,12 @@ public final class ExportDemo {
     /**
      * Entry point.
      *
-     * @param args optional output directory (default: current directory)
+     * @param args optional output directory (default: temp directory)
      * @throws IOException if writing export files fails
      */
     public static void main(String[] args) throws IOException {
-        Path outDir = args.length > 0 ? Paths.get(args[0]) : Paths.get(".");
-        if (!Files.isDirectory(outDir)) {
+        Path outDir = args.length > 0 ? Paths.get(args[0]) : Files.createTempDirectory("tamboui-export-");
+        if (args.length > 0 && !Files.isDirectory(outDir)) {
             System.err.println("Not a directory: " + outDir);
             System.exit(1);
         }


maxandersen:
p.s. lots of files but just a function of getting a cleaner fluent api...so not that much code - just type safe :)

demos/export-demo/export_demo.txt
```diff
@@ -0,0 +1,24 @@
+TamboUI Export Demo
```
maxandersen: doh - commited these by accident but then you can see the results :)

i'll remove these before pushing.

maxandersen:
slept on it and cleaned it up a bit more:

- added .crop(rect) to allow exporting just a subset - makes it trivial to save per element/widget.
- removed codeFormat option as we dont have proper template mechanism and its too easy to break even then
- added .chrome(boolean) to have svg with or without window/terminal decorations.

maxandersen:
It definitely needs to be in core imo. Otherwise we can't do the recordings nor do save or richconsole saving which want to add later too :) 

And yes, if you look at .export() it is just a shorthand for api that takes a buffer so it is not a hard requirement. 

Excellent example of where it would be great for Java to have extensions methods :) 
 
Im fine to not have buffer expose it directly - but I don't think it makes that big a difference.

demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
```diff
@@ -0,0 +1,229 @@
...
+            String name = p.getFileName().toString();
+            String url = p.toUri().toString();
+            markup.append("  [cyan][link=").append(url).append("]").append(name).append("[/link][/cyan]\n");
+        }
+        markup.append("[dim]Press Enter or Escape to close[/dim]");
+
+        int contentLines = markup.toString().split("\n").length;
+        // Wrap in panel with explicit height so dialog allocates space; dialog height = border(2) + title(1) + content
+        Element content = markupText(markup.toString()).length(contentLines);
+        int dialogHeight = 2 + contentLines;
```
maxandersen: Question @melix i couldn't get the panel/dialog to show up without manually setting height. 

What am I missing?

melix:
Well the difference is that you're committing to an API if you put it on `Buffer`. IMO doesn't belong here :)

docs/src/docs/asciidoc/developer-guide.adoc
```diff
@@ -142,6 +142,181 @@ if (x >= area.x() && x < area.right() &&
...
+// After rendering (e.g. in an action handler with access to buffer and element)
+Rect area = myPanel.renderedArea();
+if (area != null && !area.isEmpty()) {
+    buffer.export().crop(area).svg().toFile(Path.of("panel.svg"));
+}
+----
+
+=== Theme Colors
+
+Export formats use a default foreground/background when a cell style does not specify colors (e.g. after RESET). Use `ThemeColors.defaultTheme()` for the built-in dark theme, or create your own:
```
maxandersen: Theming is probably not the right word/name here as it's merely for export being able to lookup what to put as background/foreground. 

What could I use instead ? StyleResolver where "background-color" and "foreground-color" is assumed to exist - then StyleEngine could make such available configured by css?

demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
melix: Nothing, you shouldn't have to do this. Must be a bug in how `Dialog` computes its preferred height.

melix: @maxandersen pushed a commit on your branch which fixes that. Could be backported if we don't want to bind it to this PR.

docs/src/docs/asciidoc/developer-guide.adoc
melix: I indeed don't we need ThemeColors. We should use the existing StylePropertyResolver / PropertyDefinition pattern which should involve:

1. Creating new export-specific property definitions (EXPORT_FOREGROUND, EXPORT_BACKGROUND) with the same dark-theme defaults as the current ThemeColors.defaultTheme() values                                                      
2. Replacing the theme(ThemeColors) method in SvgOptions and HtmlOptions with a resolver(StylePropertyResolver) method                                                                                                             
3. Resolving the fg/bg colors in the exporters through the resolver, falling back to the property defaults when no resolver is provided

maxandersen: But those are not in core, right?

melix: They are, these are just interfaces. What is not is the CSS implementation.

maxandersen: Sorry misread on first go. I think I know what you mean now. Will update pr - but will be later due to house full of guests in a few minutes :)

demos/toolkit-export-demo/src/main/java/dev/tamboui/demo/ToolkitExportDemo.java
maxandersen: thanks!

docs/src/docs/asciidoc/developer-guide.adoc
maxandersen: pushed update for this - thanks to your excellent description just worked afaics :)

maxandersen:
> Well the difference is that you're committing to an API if you put it on `Buffer`. IMO doesn't belong here :)

well, its tedious to not have it there but I pushed version that removes it and lets you do:

```
import static dev.tamboui.export.ExportRequest.*;
export(buffer).svg().toFile("out.svg")
```
wdyt about that?

maxandersen:
default output is now a temporary directory. merging once test passes.

PR #175, opened 2026-02-01
chore: setup github action for reuse and concurrency

maxandersen:
This pull request introduces a new reusable GitHub Action for setting up the build environment and updates all relevant workflow files to use this shared setup. This change removes duplicated setup across documentation.yml, gradle.yml and publishing.yml.

Additionally, a new formatting check job is added to run spotlessCheck in parallel with the main build job.

Saves about 1-2 minutes.

The number of demos is now "hurting" the most as gradle + jbang now builds 56 jar/tar/zip files which is minium 168 * a few seconds...just for demos. This PR does not fix that - it just makes it more visible :/

### DIFF
--- a/buildSrc/src/main/java/dev/tamboui/build/UpdateJBangCatalogTask.java
+++ b/buildSrc/src/main/java/dev/tamboui/build/UpdateJBangCatalogTask.java
@@ -85,6 +85,30 @@ public abstract class UpdateJBangCatalogTask extends DefaultTask {
     public abstract Property<Boolean> getVerifyBuilds();
 
     /**
+     * Fail-fast mode for build verification.
+     */
+    @Option(option = "verify-fail-fast", description = "Fail fast if any build verification fails (not effective if parallel is enabled)")
+    @Input
+    @Optional
+    public abstract Property<Boolean> getFailFast();
+
+     /**
+     * Verbose mode for build verification.
+     */
+     @Option(option = "verify-verbose", description = "Verbose output for build verification")
+     @Input
+     @Optional
+     public abstract Property<Boolean> getVerbose();
+
+    /**
+     * Parallel mode for build verification.
+     */
+    @Option(option = "verify-parallel", description = "Verify aliases in parallel")
+    @Input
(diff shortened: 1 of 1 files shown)


maxandersen:
@melix while we dont have spotless active - this reorg still removes duplication so still reelvant. the format check will just for now not faild

.github/actions/shared-build-setup/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Set up JBang
+      uses: jbangdev/setup-jbang@main
+    
+    - name: setup-gradle
+      uses: gradle/actions/setup-gradle@748248ddd2a24f49513d8f472f81c3a07d4d50e1 # v4
+      with:
+          cache-overwrite-existing: true
+          add-job-summary: ${{ inputs.add-job-summary }}
+    - name: Setup upterm session
+      uses: owenthereal/action-upterm@v1
```
melix, line 44: Do we need this? I'm not super keen on adding 3rd party actions, especially SSH access...

.github/workflows/gradle.yml
```diff
@@ -1,31 +1,32 @@
...
 
 jobs:
+  format-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+      - uses: ./.github/actions/shared-build-setup
+      - name: check formatting and comment on PR if failed
+        id: check-formatting
+        run: ./gradlew --no-daemon spotlessCheck
```
melix: `--no-daemon` is useless, it will spawn a daemon anyway (even single use) and will kill the performance of subsequent builds if disabled

.github/actions/shared-build-setup/action.yml
maxandersen: we can remove it - its just there to enable and hook in to navigate failures.

melix:
Fixed the JBang task which wasn't failing if any demo was

PR #176, opened 2026-02-02
chore: enable import order/unused removal formatting

maxandersen:
as discussed enable import sort/removal of unused.


maxandersen:
prebased - if test passes we can push it

PR #179, opened 2026-02-04
Remove default implementation for preferred width/height

melix:
This commit removes the default implementation for preferred width and height in elements. It's currently too easy to forget to implement it, which causes all sorts of layout bugs (empty elements).

Therefore, this also fixes a number of missing implementations, and adds tests for these.

Note: there is a behavior change in `dock` now that we fixed the implementation. Since it now implements the preferred height, in order to fill the available space, you have to call `fill()` explicitly (defaults to `fit` as intended).

NB: This fixes a bug which appeared after simply updating to a newer SNAPSHOT version where the default layout strategy changed to `fit`

### DIFF
--- a/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
+++ b/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
@@ -126,6 +126,16 @@ public class ErrorHandlingDemo {
             }
 
             @Override
+            public int preferredWidth() {
+                return 0;
+            }
+
+            @Override
+            public int preferredHeight() {
+                return 0;
+            }
+
+            @Override
             public String id() {
                 return "faulty-panel";
             }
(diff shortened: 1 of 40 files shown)


graemerocher (approved):

demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
```diff
@@ -125,6 +125,16 @@ public void render(Frame frame, Rect area, RenderContext context) {
                 );
             }
 
+            @Override
+            public int preferredWidth() {
```
maxandersen, line 129: is return 0; really the right thing to do here?

maxandersen (approved):
i aprrove in principal to (at least for now) not have default implementation but is it really intentional to add that many return 0; methods?

demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
melix: I don't have any better suggestion, otherwise we'd have to return an arbitrary value. That said, it's likely going to be used in a panel that uses `fill`. In anycase that's not worse than before. As discussed earlier today, I'm not fully satisfied with the API, I feel like this is not right but can't wrap my head around what it should look like yet.

melix:
> add that many return 0; methods?

Yes, that's when we can't know. It falls back to the previous implementation. Which isn't perfect. I want to have something better, but this might require bigger changes.

PR #181, opened 2026-02-05
dev.tamboui.tui.Keys class is deprecated

### LINKED ISSUE #180: dev.tamboui.tui.Keys class is deprecated
## Issue

Code for the demo showed within the README.md or getting-started.md don't work as they use a deprecated class: ` dev.tamboui.tui.Keys class is deprecated`

cmoulliard:
- `dev.tamboui.tui.Keys` class is deprecated and should be replaced with `dev.tamboui.tui.event.KeyEvent`.
- Fix #180

### DIFF
--- a/docs/src/docs/asciidoc/getting-started.adoc
+++ b/docs/src/docs/asciidoc/getting-started.adoc
@@ -152,13 +152,14 @@ public class HelloTuiRunner {
     public static void main(String[] args) throws Exception {
         try (var tui = TuiRunner.create()) {
             tui.run(
-                (event, runner) -> {
-                    if (event instanceof KeyEvent && ((KeyEvent) event).isQuit()) {
-                        runner.quit();
-                        return false;
-                    }
-                    return false;
-                },
+                (event, runner) ->
+                    switch (event) {
+                        case KeyEvent k when k.isQuit() -> {
+                            runner.quit();
+                            yield false;
+                        }
+                        default -> false;
+                    },
                 frame -> {
                     var paragraph = Paragraph.builder()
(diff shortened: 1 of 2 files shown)


docs/src/docs/asciidoc/getting-started.adoc
```diff
@@ -153,7 +153,7 @@ public class HelloTuiRunner {
         try (var tui = TuiRunner.create()) {
             tui.run(
                 (event, runner) -> {
-                    if (event.isQuit()) {
+                    if (event instanceof KeyEvent && ((KeyEvent) event).isQuit()) {
```
maxandersen: switch (event) {
    case KeyEvent key when key.isQuit() -> {
        runner.quit();
        return false;
    }
    default -> {}
}

is cleaner.

cmoulliard: Fixed with: 3666cce

maxandersen:
thanks @cmoulliard i fixed it earlier today in #183 - i thought I had closed this; sorry about that! Thanks for opening/initial pr!

cmoulliard:
You add the link to your example in the getting.started but I think you missed to change the code in the README.md or to point to the same example as getting started !

melix:
tbh I think we should get rid of hardcoded samples in asciidoc and only include tested code (e.g from unit tests or demos), so that we guarantee that all docs compile. That should have been done since day 1, but...

maxandersen:
@cmoulliard you opened this so I couldn't rebase it directly. I've manually rebased and merged your changes for README to main. Thank you!

PR #182, opened 2026-02-05
fix: css matching landingpage

maxandersen:
this is so main landing page and content pages matches styles. 

see https://github.com/tamboui/tamboui.dev/pull/2


maxandersen:
@melix added in the various logo/icon variants to have a static place to refer to it.

PR #184, opened 2026-02-06
feat: Fat jar of the demos

### LINKED ISSUE #156: easier run-demo
I think run-demo.sh is doing the right thing but:

1) run-demo.sh doesn't run on windows.
2) it fails somewhat silently when running demos that doesnt match java
3) build system  force to java 25 in places not needed.
4) requires git clone of the repo

current idea:

make run-demo.sh a jbang script that runs on java 8, have run-demo.sh run that when need be.

which for the jvm version will utilize the jbang aliases - for native will use gradlew calls and still require clone.

This way we should be able to offer to run the demos on all platforms.

Still will require git clone of the repo but we could also have the build generate a snapshot of the metadata 
so dont need to access gradle toolkit at runtime...

Then it shuold be able to run from anywhere (minus the native variation)

melix:
This commit adds a Gradle subproject for aggregating all demos in a single, runnable fat jar. This will make running demos easier for example in Windows, by just calling java -jar or JBang.

Fixes #156


graemerocher (approved):

maxandersen:
looks good @melix!

1) shouldn't we use tamboui for this launcher ? ( opened #186 as one example)
2) Image demo fails to load image.

...will add more if i find more as i try the list.

PR #185, opened 2026-02-06
Consolidate schedulers

melix:
This commit fixes the issue that we could create multiple schedulers in a application. There wasn't either a consistency between how the schedulers were created in the toolkit and TUI helpers.

Now both use the same pattern, shared code, and we also let the user inject an external scheduler, which will be useful when integrating with frameworks like Micronaut, Quarkus or Spring Boot.


maxandersen:
looks good! 

and if i grok it right virtual threads could be used by that scheduledexectorservice to run the scheduledactions on if one wanted/needed to - or should code assume no action risk running concurrently?

melix:
mmm, not sure I get the question, but if something needs to run on the UI thread, in any case, virtual thread or not, it must use the `runOnRenderThread`

PR #186, opened 2026-02-06
interactive fat jar launcher

maxandersen:
quick'n'dirty launcher ui building on top of #184 

wdyt @melix feels sad the launcher isn't tamboui so made this:)


melix:
Looks good, only need to make code formatter happy :)

maxandersen:
fixed - had to also change tests css paths to have them pass.

Bit surprised that is not failing here in github actions? @melix any idea?

maxandersen:
this works for me - Its more verbose than i wnated but toolkit inline seems broken in comparison. will try isolate and open separately

PR #188, opened 2026-02-07
fix: Fix toolkit demo

### LINKED ISSUE #187: cannot close the toolkit-demo
the textfield in one of the tabs grabs all keys even its not in focus.

I think the changed focus routing is affecting it.

@melix ? is focus being too eagerly?

melix:
The todo component shouldn't handle keys when no modifier is used.

Fixes #187

### DIFF
--- a/docs/src/docs/asciidoc/bindings.adoc
+++ b/docs/src/docs/asciidoc/bindings.adoc
@@ -54,6 +54,53 @@ KeyTrigger.ctrl('u')                 // Ctrl+U
 KeyTrigger.alt('x')                  // Alt+X
 ----
 
+=== Terminal Compatibility
+
+[IMPORTANT]
+====
+**Ctrl key bindings have significant limitations** due to how terminals encode keyboard input.
+
+When you press a key combination like Ctrl+A, the terminal does not send "Ctrl" + "A" separately.
+Instead, it sends a single ASCII control character (byte value 1 for Ctrl+A).
+This encoding dates back to the 1970s and causes several problems:
+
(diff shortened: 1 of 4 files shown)


maxandersen:
But isn't x still stolen ?

melix:
Yes you're right, `x` needs similar fix. But the behavior is intended. All panels should handle events, and because `x` is captured by the text field, it handles it. In think choosing `x` was a bad idea in the beginning. That said, I also think that the key event handler should probably also have a way to figure out if it's focused. That's the kind of things which we need to improve, but there's no easy solution.

maxandersen:
Yeah I was expecting the text field to only get characters when actually In focus.

melix:
For that demo yes, but in general that's not expected.

melix:
@maxandersen found another bug that I fixed in the process (ALT parsing was broken)

docs/src/docs/asciidoc/bindings.adoc
```diff
@@ -54,6 +54,53 @@ KeyTrigger.ctrl('u')                 // Ctrl+U
...
 
+=== Terminal Compatibility
+
+[IMPORTANT]
+====
+**Ctrl key bindings have significant limitations** due to how terminals encode keyboard input.
+
+When you press a key combination like Ctrl+A, the terminal does not send "Ctrl" + "A" separately.
+Instead, it sends a single ASCII control character (byte value 1 for Ctrl+A).
+This encoding dates back to the 1970s and causes several problems:
```
maxandersen, line 65: yes, but aren't there actually terminals where extended sequences using esc [ 27 allow geting more info?

melix: There are, but, I don't think we should recommend to anyone to use Terminal specific apis.

PR #190, opened 2026-02-08
redo tracer / harness

maxandersen:
this is attempt of #169 with lower level sink but not really happy :) putting as draft for now.

rennamed to harness as after reviewing this feels more like a test fixture for us rather than users?

### DIFF
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -102,6 +102,9 @@ Architecture patterns for building maintainable TUI applications.
 === link:developer-guide.html[Developer Guide]
 Creating custom widgets and components.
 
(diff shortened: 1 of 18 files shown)


maxandersen:
marking as draft.

maxandersen:
thinking about this i'm starting to think we might just be better of having a utility method to format event info and serialize that into a specifc dev.tamboui.trace category .. then provide java util logging config file or simple method to call that will write this out to file and/or a websocket/http for easy remote "listening"

could still have the typesafe apis to enable other tools/frameworks to hook in a have more integrated view than just parse string json representations.

melix:
Been there :) That's why I had the other APIs because it was quite painful to hook :)

maxandersen:
this isnt the way. closing

PR #192, opened 2026-02-11
fix: Make it possible to fire events programmatically

melix:


tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
```diff
@@ -825,6 +830,55 @@ public T onAction(ActionHandler handler) {
...
+            MouseEventHandler prev = this.mouseHandler;
+            onMouseEvent(event -> {
+                if (trigger.matches(event)) {
+                    handler.accept(event);
+                    return EventResult.HANDLED;
+                }
+                return prev != null ? prev.handle(event) : EventResult.UNHANDLED;
+            });
+        }
+        return self();
```
maxandersen, line 882: so if I do .on(new ActionEvent(), (e) -> ..) it is meant to be ignored?

tamboui-tui/src/main/java/dev/tamboui/tui/event/ActionEvent.java
```diff
@@ -0,0 +1,57 @@
...
+package dev.tamboui.tui.event;
+
+/**
+ * Represents a programmatically fired action, as opposed to a raw input event.
+ * <p>
+ * ActionEvent is created when an action is fired via
+ * {@link dev.tamboui.tui.bindings.ActionHandler#fire(String)} without an
+ * originating input event. It carries the action name that was fired.
+ */
+public final class ActionEvent implements Event {
```
maxandersen: should an ActionEvent have a <T> source; ?

(jfx and swing had a Object source; but I was thinking ActionEvent<MyThingThatCausedThis> would be useful?

maxandersen: i.e. so the event can optionally carry information.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
melix: I don't understand what you mean. ActionEvent is an `Event`, not a trigger.

tamboui-tui/src/main/java/dev/tamboui/tui/event/ActionEvent.java
melix: We could. However a wise person said "don't add code unless you have a use case" :)

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
```diff
@@ -825,6 +830,55 @@ public T onAction(ActionHandler handler) {
...
+     * <strong>Note:</strong> Calling {@link #onKeyEvent(KeyEventHandler)} or
+     * {@link #onMouseEvent(MouseEventHandler)} after {@code on()} replaces the
+     * chained handler, same as calling those methods multiple times.
+     *
+     * @param trigger the input trigger to match (e.g., {@code MouseTrigger.click()}, {@code KeyTrigger.ch('r')})
+     * @param handler the handler to invoke when the trigger matches
+     * @return this element for chaining
+     */
+    public T on(InputTrigger trigger, Consumer<Event> handler) {
+        if (trigger instanceof KeyTrigger) {
```
maxandersen, line 862: should we assert these can't be null?

melix: We can probably throw NPEs.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
maxandersen: ah sorry - misread it. 

so it .on(new SomeInstanceOfInputTrigger(), (e) -> ...)  supposed to be ignored/pass through and not fail?

tamboui-tui/src/main/java/dev/tamboui/tui/event/ActionEvent.java
maxandersen: `.fire("deleteUser", list.get(user)); ` ?

`fire("doSomeAction", this)`

and "doSomeAction' or "deleteUser" can get context without having to look it up.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/element/StyledElement.java
melix: Right. If we make it fail this would prevent other elements to potentially handle the event (by overriding). Note that if we had sealed classes, we could ensure that all cases are treated..

PR #196, opened 2026-02-14
fix: do not try run  demos without a tape file

maxandersen:
demo without tape file (like aesh-ssh demo) gets stuck and never ends.

So added this to make it skip if no tape file (not just internal as we have some marked internal that have .tape files)


maxandersen:
merging as main builds are stuck without it. @melix complain if this was done wrong :)

PR #199, opened 2026-02-16
Docs update

melix:


maxandersen:
you wanna do the same with these which I moved earlier but I think shuold just follow this new format ?

src/docs/asciidoc/code/getting-started/HelloDsl.java
src/docs/asciidoc/code/getting-started/HelloImmediate.java
src/docs/asciidoc/code/getting-started/HelloInline.java
src/docs/asciidoc/code/getting-started/HelloTuiRunner.java

melix:
These were actually migrated, forgot to delete the files.

PR #200, opened 2026-02-16
fix: Single `preferredSize` method

melix:
This commit refactors the preferredHeight/Width methods so that we only have one. It also fixes the ambiguity that size 0 could either mean "no preference" and "invisible"/"hidden". A bug was actually visible in the flex layout demo, where the 2d tab showed empty cells.

### DIFF
--- a/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
+++ b/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
@@ -13,8 +13,8 @@ import dev.tamboui.style.Color;
 import dev.tamboui.terminal.Frame;
 import dev.tamboui.toolkit.app.ToolkitRunner;
 import dev.tamboui.toolkit.element.Element;
-import dev.tamboui.toolkit.element.PreferredSize;
 import dev.tamboui.toolkit.element.RenderContext;
+import dev.tamboui.toolkit.element.Size;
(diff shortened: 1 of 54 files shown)


maxandersen:
Til many files to navigate on mobile so can't check...but wouldn't a Size class be enough over a PreferredSize?

melix:
> Til many files to navigate on mobile so can't check...but wouldn't a Size class be enough over a PreferredSize?

I feared this would be overloaded, but it's indeed simpler. Fixed.

tamboui-toolkit/src/test/java/dev/tamboui/toolkit/elements/MarkupTextAreaElementTest.java
```diff
@@ -206,7 +206,7 @@ void scrollToLineWorks() {
     void preferredHeightReturnsLineCount() {
         MarkupTextAreaElement element = markupTextArea("Line 1\nLine 2\nLine 3");
 
-        assertThat(element.preferredHeight()).isEqualTo(3);
+        assertThat(element.preferredSize(-1, -1, null).heightOr(0)).isEqualTo(3);
```
maxandersen, line 209: should there be a preferredSize() method that does that so one dont have to keep repepating -1,-1,nulll for default values?

melix:
> shouldn't we have a prefferedSize() that doesn't take arguments

My idea was to NOT do this, otherwise it's too simple to miss the fact that it's context aware. The noise in the PR is actually in tests, for production code, you should get "real" data.

PR #203, opened 2026-02-17
fix: add link to source in demo launcher

maxandersen:
wanted to add a link to source from the demo launcher. it probably should be a small panel on the right but for now just a link in the listitem...BUT they are not being rendered atm.

@melix can you spot why  listitem is not applying the hyperlink styles?


maxandersen:
this works now! better that nothing :)

PR #211, opened 2026-02-18
Add .sdkmanrc file for Java 25.0.2 support

wfouche:


melix:
Thanks. I'd rather use the GraalVM version, since we have demos that compile to native images, I'd better use `25.0.2-graalce`

PR #215, opened 2026-02-18
feat: Add BOM

melix:
This commit introduces a BOM module (tamboui-bom), as stated :)


maxandersen:
cool...could we add a notice in the docs on how to use it from gradle /maven?

PR #216, opened 2026-02-18
Doom in terminal

maxandersen:
After https://www.youtube.com/watch?v=ERAp2vRHX3M some are asking to see the Doom code.

I didn't want to push it in as this is definitily vibe coded hard! It did help me find a few perf glitches.

To give some context this came about because of this conversation:

I then just told a llm agent (cursor. auto mode) to go implement doom to impress Cedric.
It came back with a ascii map walker - not great but good enough. Then challenged it to support
WAD files ...so it did ...and it ran very smoothly (60fps or so on macbook)

Then asked it to fix some glitches and include ceiling/floor rendering and here it still works ok in block text but in native image it goes bad. I haven't had time to look and isolate why - but I'm pretty sure its a mix of pushing terminal to its limits plus the code rendering way more than it need to.

It did show me that its not java that is the bottleneck in making this fast - its an IO challenge :)

Anyhow - putting this as a draft for others to poke at and prove the video was not lying.

- **Add Doom raycaster demo**
- **Add WAD loading and hi-res renderer**
- **fix: stop pulsing**
- **Add image-rendered WAD textures**
- **feat: add p for protocol switcher**
- **demos**
- **Add per-wall and flat textures**
- **Fix MapLoadResult defaults**
- **turn/speed**
- **mouse tracking**

### DIFF
--- a/demos/color-demo/build.gradle.kts
+++ b/demos/color-demo/build.gradle.kts
@@ -1,0 +1,14 @@
+plugins {
+    id("dev.tamboui.demo-project")
+}
+
+description = "Color modes demo - ANSI, indexed, and RGB"
+
+demo {
+    displayName = "Color modes demo"
+    internal = true
+}
+
+application {
+    mainClass.set("dev.tamboui.demo.ColorDemo")
+}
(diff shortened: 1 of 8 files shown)


franz1981:
@maxandersen i would be curious to see if #290 could help this as well ^^

maxandersen:
@franz1981 it does seem to be a bit more fluent in the highres image mode.

franz1981:
If we reports the FPS it should be very visible from there ❤️

PR #221, opened 2026-02-19
feat: Smarter ListWidget

melix:
This commit refactors ListWidget and ListElement so that they behave like TreeWidget/TreeElement. When ListWidget was designed, it followed the same pattern as in Ratatui, only allowing text as children. With this change, the list widget supports any kind of widget as child, similarly to TreeWidget. For this, it makes use of `SizedWidget`.

A common package has been created, which also consolidates duplicate ScrollBarPolicy.

This is an API breaking change.


docs/src/snippets/java/dev/tamboui/docs/snippets/WidgetsSnippets.java
```diff
@@ -146,20 +146,17 @@ void paragraphWidget() {
...
-
         // Create state
         ListState state = new ListState();
         state.select(0);  // Select first item
 
         // Build and render
         ListWidget list = ListWidget.builder()
-            .items(items)
+            .items(
+                ListItem.from("First item"),
```
maxandersen, line 156: just a general API idea - we use .from() *alot* which automatically means you are forced to list the full class name.

Should we have a convention of using lowercase version or some ohter version instead?

i.e. ListItem.item() or ListItem.listItem() so static import can be done and you can do item("first item") as well as row("...") and tableitem("sss") in same file?

tamboui-widgets/demos/list-demo/src/main/java/dev/tamboui/demo/ListDemo.java
```diff
@@ -289,15 +284,31 @@ private void renderBasicList(Frame frame, Rect area) {
      * Render a list with multiline items and custom styling.
      */
     private void renderMultilineList(Frame frame, Rect area) {
-        List<ListItem> items = getMultilineItems();
-
         ListWidget list = ListWidget.builder()
-            .items(items)
+            .items(
+                ListItem.from(Text.from(
```
maxandersen, line 289: good example where listItem(text(line())) would be cleaner.

docs/src/snippets/java/dev/tamboui/docs/snippets/WidgetsSnippets.java
melix: I get the idea, but I'm not sure I like it either, because if you _dont_ use a static import, you have `ListItem.listItem`, which makes my eye bleed as well.

tamboui-widgets/src/test/java/dev/tamboui/widgets/list/ListWidgetTest.java
```diff
@@ -136,239 +127,122 @@ void selectNextAtEnd() {
...
+        // Only 2 rows visible
+        Rect area = new Rect(0, 0, 20, 2);
         Buffer buffer = Buffer.empty(area);
         ListState state = new ListState();
-        state.select(0); // Select first item (should appear at bottom)
+        state.select(3); // Select 4th item (index 3)
 
         list.render(area, buffer, state);
 
-        // In bottom-to-top, first item should be at bottom (selected, so has reversed style)
```
maxandersen, line 158: on first look it feels like its removing tests over expanding - is that expected?

docs/src/snippets/java/dev/tamboui/docs/snippets/WidgetsSnippets.java
maxandersen: well, can still keep .from() as an alias.

maxandersen (approved):
current change looks good to me - assuming the reduced tests are expected.

melix:
Ok let's discuss the API changes in a separate PR. I also think we should split `Toolkit` into smaller pieces.

Issue #220, opened 2026-02-19
Windows not fully supported


wfouche:
The command to run the demos ("jbang run demos@tamboui") does not work on Windows. Pressing the Up/Down arrow keys causes A and B characters to be printed.

To reproduce, start the demo and press the DownArrow key three times.

<img width="939" height="402" alt="Image" src="https://github.com/user-attachments/assets/9a378e50-08ff-4080-8f36-fcbc41caa983" />

wfouche:
The root cause of this issue might just be a JLine initialization issue. Hopefully easy to fix.

A similar program using Mordant works fine from the Windows command-line.

```kotlin
//DEPS com.github.ajalt.mordant:mordant-jvm:3.0.2
//KOTLIN 2.3.10
import com.github.ajalt.mordant.rendering.TextColors.*
import com.github.ajalt.mordant.terminal.Terminal
import com.github.ajalt.mordant.input.interactiveSelectList

fun main() {
    val t = Terminal()

    t.println("--- Mission Control System ---")
    t.println("Please select your next destination using the arrow keys:\n")

    // Define the interactive prompt
    val selection = t.interactiveSelectList(
        entries = listOf(
            "Mars - The Red Planet",
            "Europa - Ocean Moon",
            "Titan - Methane Lakes",
            "Alpha Centauri - Interstellar Journey",
            "Abort Mission"
        ),
        title = "Available Destinations:"
    )

    // React to the user's choice
    when (selection) {
        "Abort Mission" -> t.println(red("Mission cancelled. Powering down..."))
        null -> t.println(yellow("No selection made.")) // Handles ESC or interrupts
        else -> {
            t.println("\n" + green("Success!") + " Navigation locked on to: $selection")
            t.println("Safe travels, Commander.")
        }
    }
}
```

`C:\>  jbang run test.kt`

<img width="525" height="189" alt="Image" src="https://github.com/user-attachments/assets/235c19a8-7d7b-4440-bfa1-c94514f152f8" />

maxandersen:
I also notice that some escape codes are being sent which seems inappropriate too.

`?2027;3$y`

That is coming from Mode2027Status.java.

PR #223, opened 2026-02-20
fix: windows friendly escape reading

maxandersen:
this is one way of fixing #220 but I'm not a fan of having to add this kind of code that gets enabled on all platforms....should we make it environment dependent ?

also, this does not fix fully the printing of esc code dumped to screen...now it just shows up as a `y' :)

It also on non panama or maybe just java version makes repaints/refreshes way more visibile on windows - so it does have some impact on refresh rates.

marking draft for now for input.


maxandersen:
Superseded by #360 which fixes the actual root cause — `WindowsTerminal.peek()` returning a flag instead of the real character. That fix makes the EventParser work correctly without needing parser-level workarounds.

PR #238, opened 2026-02-21
fix: add versioned dropdown in docs

maxandersen:
this adds a version selector on main doc page:

its done by page looking for /docs/versions.json - if available it will show the drop down.

the docs/versions.json is generated as part of gh pages publish.

Also added readme.md so i dont forget how to easily run this thing :)

Maybe controversial - and need to check if more changes needed for publish to work:
use https github urls so users without ssh keys setup can run this just fine - only at time of publish would key be needed.

current limits (why its in draft): 
- given 0.1.0 is already published the dropdown dissapears when picking that.
- javadoc is currently not having the dropdown at all but can look at doing similar.
- need to check publish still works.


melix:
Looks like a good idea to me

PR #240, opened 2026-02-21
fix: prevent auto-focus from overriding explicit focus clear (Escape)

KyleKreuter:
## Summary

When pressing **Escape** in a focused input field, `clearFocus()` correctly sets `focusedId` to `null`. However, on the very next render cycle, two auto-focus mechanisms immediately re-focus the first focusable element:

1. `FocusManager.registerFocusable()` — auto-focuses the first element when `focusedId == null`
2. `ToolkitRunner.run()` / `InlineToolkitRunner.run()` — explicitly sets focus to the first element when `currentFocus == null`

This makes it impossible to fully clear focus via Escape — the focus always snaps back to the first focusable element.

## Solution

Add a `focusExplicitlyCleared` flag to `FocusManager` that tracks whether focus was intentionally removed:

- **Set** by `clearFocus()` (Escape key, clicking outside focusable elements)
- **Reset** by any explicit focus action: `setFocus()`, `focusNext()`, `focusPrevious()`, `focusAt()` (Tab, Shift+Tab, mouse click on element)
- **Checked** in `registerFocusable()` and both runner render callbacks — auto-focus is skipped when the flag is set

## Changed files

- `FocusManager.java` — new `focusExplicitlyCleared` flag with getter, integrated into all focus mutation methods
- `ToolkitRunner.java` — check `isFocusExplicitlyCleared()` before auto-focusing
- `InlineToolkitRunner.java` — same check

## Test plan

- [x] Existing `tamboui-toolkit` tests pass (`./gradlew :tamboui-toolkit:test`)
- [x] Manual: Focus a text input → press Escape → verify no element is focused
- [x] Manual: After Escape clear, press Tab → verify focus returns to first element
- [x] Manual: After Escape clear, click on an input → verify it gets focused

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/InlineToolkitRunner.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/app/InlineToolkitRunner.java
@@ -160,9 +160,10 @@ public final class InlineToolkitRunner implements AutoCloseable {
                 }
 
                 // Auto-focus first focusable element if nothing is focused or focus is stale
+                // (but not if focus was explicitly cleared, e.g. via Escape)
                 String currentFocus = focusManager.focusedId();
                 List<String> focusOrder = focusManager.focusOrder();
-                if (!focusOrder.isEmpty()) {
+                if (!focusOrder.isEmpty() && !focusManager.isFocusExplicitlyCleared()) {
                     if (currentFocus == null || !focusOrder.contains(currentFocus)) {
                         focusManager.setFocus(focusOrder.get(0));
                     }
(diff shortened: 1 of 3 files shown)


maxandersen:
any of the demos that shows the issue ? or got a small reproducer for what you are seeing fixed with this?

KyleKreuter:
i am currentyl working on a spring integration and noticed this when two input fields are together.

KyleKreuter:
I am currently working on a [spring integration](https://github.com/KyleKreuter/tamboui-spring) and noticed this when two input fields are together: 

Without the fix
https://github.com/user-attachments/assets/7e53bb58-5722-4874-872a-748033fb9732

With the fix
https://github.com/user-attachments/assets/5732d4ae-06ec-46dd-8035-16c48a1d04fd

In both cases I pressed esc to lose focus on the second input field

maxandersen:
ah - now i saw video on bigger screen i grok it :)

same issue is also visible in `jbang demos@tamboui form-table-demo`

@melix wdyt ?

PR #245, opened 2026-02-22
feat(toolkit): add native SelectState support to FormFieldElement

KyleKreuter:
## Summary

`FormFieldElement` only accepts `SelectFieldState`, forcing consumers to copy `SelectState` into `SelectFieldState` and breaking two-way binding. `TextInputState` already works as a direct reference — this makes `SelectState` symmetric.

- Add `selectStateRef` field and `SelectState` constructor to `FormFieldElement`
- Prefer `selectStateRef` in `handleSelectFieldKey()` and `renderSelect()`
- Add `Toolkit.formField(String, SelectState)` factory method
- Fully backward-compatible, no existing API changes

## Motivation

When using `FormFieldElement` with a standalone `SelectState` (outside of `FormState`), the only option is to copy into a `SelectFieldState`:

```java
// Copies state — UI changes don't reach the original SelectState
SelectFieldState copy = new SelectFieldState(ss.options(), ss.selectedIndex());
var element = Toolkit.formField("", copy);
```

With this change:

```java
// Direct reference — true two-way binding
var element = Toolkit.formField("", selectState);
```

## Changes

| File | Change |
|---|---|
| `FormFieldElement.java` | New `selectStateRef` field, new constructor, updated `handleSelectFieldKey()` and `renderSelect()` |
| `Toolkit.java` | New `formField(String, SelectState)` factory method |

## Test plan

- [ ] Existing tests remain green
- [ ] Create `FormFieldElement` with `SelectState`, verify `selectNext()`/`selectPrevious()` modifies the original state
- [ ] Create `FormFieldElement` with `SelectFieldState`, verify existing behavior unchanged
- [ ] Verify rendering uses original `SelectState` (no copy in `renderSelect()`)

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
@@ -57,6 +57,7 @@ import dev.tamboui.widgets.form.FormState;
 import dev.tamboui.widgets.form.SelectFieldState;
 import dev.tamboui.widgets.input.TextAreaState;
 import dev.tamboui.widgets.input.TextInputState;
+import dev.tamboui.widgets.select.SelectState;
 import dev.tamboui.widgets.scrollbar.ScrollbarState;
 import dev.tamboui.widgets.spinner.SpinnerStyle;
 import dev.tamboui.widgets.tree.TreeNode;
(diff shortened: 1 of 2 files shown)


melix:
Looks good to me but would be good if you could add some tests, thank you!

Issue #206, opened 2026-02-17
more graceful handling when no terminal available


maxandersen:
i.e. when using `gradle run` on basic app this happens:

```
> Task :run
Feb 17, 2026 5:56:38 PM org.jline.utils.Log logr
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)

<=========--<=========----> 75% EXECUTING [8s]
> :run
^[[?2027;0$y^C⏎      
```

We should be able to know the terminal is not proper and thus not start pusing ansi/escape codes.

the app/rendering should still work though if we are inline mode or similar.

PR #248, opened 2026-02-22
feat: Add support for setting window title

melix:


tamboui-core/src/main/java/dev/tamboui/terminal/AnsiStringBuilder.java
```diff
@@ -117,6 +117,22 @@ public static String hyperlinkEnd() {
...
+     * <p>
+     * Control characters (ESC and BEL) are stripped from the title to prevent
+     * sequence injection or premature termination.
+     *
+     * @param title the window title to set
+     * @return the OSC 2 escape sequence
+     */
+    public static String windowTitle(String title) {
+        String safeTitle = title.replace("\u001b", "").replace("\u0007", "");
+        return ESC + "]2;" + safeTitle + "\u0007";
```
maxandersen, line 133: does this work on all terminals? i.e. windows I assume it would be noise?

maxandersen (approved):
ok as its not forced - but do think its going to mess up windows so #206 and #220 highly important.

maxandersen:
Rebased onto main and pushed a follow-up commit:

- `setWindowTitle` on `TuiRunner`/`InlineTuiRunner`/`ToolkitRunner`/`ToolkitApp` now throws `RuntimeIOException` instead of checked `IOException`, matching the framework's exception strategy (removes the try/catch boilerplate in DemoSelector)
- The original window title is saved on first use (XTWINOPS `CSI 22;2t`) and restored on close (`CSI 23;2t`), so apps no longer permanently overwrite the user's terminal title; terminals without title-stack support ignore the sequences
- Added `AnsiStringBuilderTest` coverage for the title sequences and ESC/BEL injection stripping

Re Windows: OSC 2 works in Windows Terminal, mintty, and ConEmu; conhost with VT processing enabled (which JLine enables) silently ignores it — so no noise.

PR #256, opened 2026-02-23
feat(toolkit): add focusedBorderColor to ListElement and fix unfocused key handling

KyleKreuter:
## Summary

Adds focus-aware border color support to `ListElement` and fixes key event handling when unfocused — matching the existing behavior of `Panel`, `TextAreaElement`, `TextInputElement`, and `FormFieldElement`.

### Changes

**`ListElement.java`**

1. **`focusedBorderColor` support**: New field + setter `focusedBorderColor(Color)`. When set and the element is focused, the border renders in this color instead of the default `borderColor`. Resolution order: programmatic (`focusedBorderColor` / `borderColor`) > CSS — consistent with all other elements.

2. **`renderBorder()` updated**: Border now renders when `focusedBorderColor` is set (in addition to existing `title` / `borderType` triggers). Focus state is queried via `context.isFocused(elementId)`.

3. **`preferredSize()` updated**: Accounts for border dimensions when `focusedBorderColor` is set, preventing layout shifts.

4. **`handleKeyEvent()` fix**: Added `if (!focused) return EventResult.UNHANDLED;` guard — previously, UP/DOWN/PAGE keys were consumed even when the list was not focused. This is the same fix as PR #249 for `TreeElement`.

## Test plan

- [ ] Verify `ListElement` renders border in `focusedBorderColor` when focused
- [ ] Verify border falls back to `borderColor` when unfocused
- [ ] Verify CSS `border-color` applies when no programmatic color is set
- [ ] Verify UP/DOWN keys are ignored when list is not focused
- [ ] Verify `preferredSize()` includes border space when `focusedBorderColor` is configured

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/ListElement.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/ListElement.java
@@ -122,6 +122,7 @@ public final class ListElement<T> extends StyledElement<ListElement<T>> {
     private String title;
     private BorderType borderType;
     private Color borderColor;
+    private Color focusedBorderColor;
     private boolean autoScroll;
     private boolean autoScrollToEnd;
     private boolean stickyScroll;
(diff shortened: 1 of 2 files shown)


melix:
Can you make sure to add tests for this? The tests should use the buffer assertions we have and assert the rendered results. Thank you!

melix:
I think there are a few issues in that PR. First, the behavior is inconsistent with textarea/text input. They will only set a focused border color IF there's a border. I think your PR adds a border unconditionally if focused border color is set.

e.g they do: `if (title != null || borderType != null || effectiveBorderColor != null) {` <-- compares with _effective style_

Same goes for the effective size. The next pb (which is actually a gap easy to miss) is that CSS pseudo-class `:focus` is not handled in that PR. That's not necessarily a pb though, we can add support later.

Issue #243, opened 2026-02-22
Version should be passed to asciidoctor


melix:
The asciidoctor files contain an explicit version as attribute, but it should be passed as a parameter when building.

PR #264, opened 2026-02-24
Comment out misleading attribute which Gradle already sets.

lukasz-kapica:
As per #243 version is already passed as a parameter.

dev.tamboui.docs.gradle.kts line 136

```java
attributes(
        mapOf(
            // ...
            "project-version" to project.version,
            "project-name" to "TamboUI",
            // ...
        )
    )
```

### DIFF
--- a/docs/src/docs/asciidoc/_attributes.adoc
+++ b/docs/src/docs/asciidoc/_attributes.adoc
@@ -1,6 +1,6 @@
 // Shared attributes for TamboUI documentation
 :project-name: TamboUI
-:project-version: 0.1.0-SNAPSHOT
+// :project-version: set by Gradle
 :github-repo: tamboui/tamboui
 :github-url: https://github.com/{github-repo}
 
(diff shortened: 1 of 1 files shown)


maxandersen:
I believe this is so renders show something useful in github without build....

melix (changes requested):
This must not be done, it's a very dirty hack. There are attributes that are injected automatically, including the version. And since I think the version is injected automatically as `version`, I think the simplest and logical fix is to replace `project-version` in the asciidoc files with `version` once for all.

lukasz-kapica:
Right now it renders {project-version}. So it's apparent what it is. Do you want to change it to 'version'?

maxandersen:
Pushed a fix — instead of commenting out the attribute, set it to `LATEST` as a fallback. This way GitHub rendering shows something useful ("use the latest version") while the Gradle build still overrides it with the actual version for published docs. No stale hardcoded values to maintain.

Issue #194, opened 2026-02-13
text input does not seem to accept unicode


maxandersen:
typing in unicode characters does not seem to work. would be nice if it did :)

PR #265, opened 2026-02-24
Fix for #194. Allow typing unicode characters

lukasz-kapica:
This should solve #194.

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
@@ -1290,12 +1290,14 @@ public final class Toolkit {
                 state.moveCursorToEnd();
                 return true;
             case CHAR:
-                // Don't consume characters with Ctrl or Alt modifiers - those are control sequences
-                if (event.modifiers().ctrl() || event.modifiers().alt()) {
+                // AltGr on Windows is reported as Ctrl+Alt — allow it for character input.
+                // Block only standalone Ctrl or Alt, which indicate control sequences.
+                boolean isAltGr = event.modifiers().ctrl() && event.modifiers().alt();
+                if (!isAltGr && (event.modifiers().ctrl() || event.modifiers().alt())) {
                     return false;
                 }
                 char c = event.character();
-                if (c >= 32 && c < 127) {
+                if (!Character.isISOControl(c)) {
                     state.insert(c);
                     return true;
                 }
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
@@ -1290,10 +1290,10 @@ public final class Toolkit {
                 state.moveCursorToEnd();
                 return true;
             case CHAR:
-                // Don't consume characters with Ctrl or Alt modifiers - those are control sequences
-                if (event.modifiers().ctrl() || event.modifiers().alt()) {
-                    return false;
-                }
+//                // Don't consume characters with Ctrl or Alt modifiers - those are control sequences
+//                if (event.modifiers().ctrl() || event.modifiers().alt()) {
+//                    return false;
+//                }
                 char c = event.character();
                 state.insert(c);
                 return true;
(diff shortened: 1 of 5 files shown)


tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
```diff
@@ -4,51 +4,12 @@
...
-import dev.tamboui.toolkit.elements.SpinnerElement;
-import dev.tamboui.toolkit.elements.StackElement;
-import dev.tamboui.toolkit.elements.TableElement;
-import dev.tamboui.toolkit.elements.TabsElement;
-import dev.tamboui.toolkit.elements.TextAreaElement;
-import dev.tamboui.toolkit.elements.TextElement;
-import dev.tamboui.toolkit.elements.TextInputElement;
-import dev.tamboui.toolkit.elements.TreeElement;
-import dev.tamboui.toolkit.elements.WaveTextElement;
+import dev.tamboui.toolkit.elements.*;
```
melix: :x: Please don't collapse imports

melix (commented):
Would be nice to have some tests. I see this is implemented at the toolkit level. Does that mean it works on TUI or with plain backend?

lukasz-kapica:
I noticed that TextInputElement is using this utility method for handling input characters. Other than that only FormFieldElement is using it along with CodingAssistantDemo and some snippet. I will have to look more into this to see how to test it. I'm not even sure why this method is here and not elsewhere.

maxandersen:
@lukasz-kapica looks like a meaningful fix but yeah a unit test in tamboui-toolkit/src/test/java/dev/tamboui/toolkit/elements/TextInputElementTest.java and tamboui-widgets/src/test/java/dev/tamboui/widgets/input/TextInputTest.java seems like a good place for test?

lukasz-kapica:
The KeyCode enum javadoc says on CHAR:
```java
    // Special
    /** Regular printable character - check {@link KeyEvent#character()} */
    CHAR,
```

So it should be already printable in Toolkit.handleTextInputKey CHAR case.

Another problem is that AltGr is handled differently across different OSes.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/Toolkit.java
```diff
@@ -1295,11 +1295,8 @@ public static boolean handleTextInputKey(TextInputState state, KeyEvent event) {
                     return false;
                 }
                 char c = event.character();
-                if (c >= 32 && c < 127) {
```
lukasz-kapica, line 1298: I removed the check because at this point c should be printable.

tamboui-toolkit/src/test/java/dev/tamboui/toolkit/elements/TextInputElementTest.java
```diff
@@ -78,4 +79,38 @@ void attributeSelector_placeholder_affectsBorderColor() {
...
+        Frame frame = Frame.forTesting(buffer);
+
+        input.renderContent(frame, area, context);
+
+        String content = extractBufferContent(buffer, 0, 0, 10, 1);
+
+        assertThat(content).contains(unicodeChars);
+    }
+
+    private String extractBufferContent(Buffer buffer, int startX, int startY, int width, int height) {
```
lukasz-kapica, line 106: This helper function is copied from another module.

lukasz-kapica:
Hi @maxandersen  can you take a look?

maxandersen:
@lukasz-kapica can't seem to make it work - did it work for you? whats app/steps did you test with?

lukasz-kapica:
@maxandersen I have this project https://github.com/lukasz-kapica/tamboui-todo which is using textInput. Given the tambui is published to maven local I run mvn clean package and run the fat jar (without the original part). Then I can type 'ł' using option + l.

maxandersen:
ok, so I dont see that. Also by unicode I was thinking the alt + 00A9 (copyright) or simply paste them in ...those doesn't seem to do much.

lukasz-kapica:
@maxandersen  I commented out ctrl, alt and printable checks for a test, did it solve it for you?

maxandersen:
sorry for delay - traveling this week.  I don't see it work still unfortunately but I'll want to test this when back home.

maxandersen:
afaics you can now type unicode in main so closing this.

PR #268, opened 2026-02-25
Add locale support to Monthly 

lukasz-kapica:
Without this I get 2 tests wrong:

MonthlyTest.renderWithBothHeaders
MonthlyTest.renderWithMonthHeader

Because Monthly.renderMonthHeader uses default locale.

```java
        String monthName = yearMonth.getMonth().getDisplayName(TextStyle.FULL, Locale.getDefault());
```


maxandersen:
good catch - though the fix makes the test pass it hides the issue - we shouldn't rely on the global default, Monhtly should be given a locale to use (which by default should be falling back to default Locale).

Want to make that change in the PR? add Locale to Monthly using default locale, but allow setting locale in builder?

maxandersen:
perfect @lukasz-kapica ! thanks! 

@all-contributors add @lukasz-kapica
 for code

PR #271, opened 2026-02-26
Fix month names in Monthly

lukasz-kapica:
Monthly when created with custom polish locale had two problems with it i.e. for the month of June it rendered "czerwca" instead of "Czerwiec". So wrong declination and wrong letter casing. The former could be solved by changing TextStyle to FULL_STANDALONE and the latter with the help of the new method which capitalizes the first letter of the input String.

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/Column.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/Column.java
@@ -205,14 +205,32 @@ public final class Column extends ContainerElement<Column> {
                 // First try text element special case
                 c = calculateDefaultConstraint(child);
                 if (c == null) {
-                    Size size = child.preferredSize(-1, -1, context);
+                    Size size = child.preferredSize(effectiveArea.width(), -1, context);
                     int preferred = size.height();
                     c = preferred >= 0 ? Constraint.length(preferred) : Constraint.fill();
                 }
             } else if (c instanceof Constraint.Fit) {
-                Size size = child.preferredSize(-1, -1, context);
+                Size size = child.preferredSize(effectiveArea.width(), -1, context);
                 int preferred = size.height();
                 c = preferred >= 0 ? Constraint.length(preferred) : Constraint.fill();
+            } else if (c instanceof Constraint.Min) {
+                // For elements whose preferred height depends on the available width
+                // (e.g., WRAP_CHARACTER / WRAP_WORD text), the Constraint.Min returned by
+                // constraint() is based on the raw newline count and may be far too small.
+                // Recalculate using the actual column width to get the true wrapped height
+                // and raise the floor so fill spacers cannot crowd the element below what it needs.
+                // Constraint.min is used instead of Constraint.length so that the element can
+                // still grow if no fill competitors are present (e.g., single wrapping text with
+                // no spacers gets all remaining space, allowing all wrapped lines to render).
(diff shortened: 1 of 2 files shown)


tamboui-widgets/src/main/java/dev/tamboui/widgets/calendar/Monthly.java
```diff
@@ -222,7 +232,7 @@ private int renderMonthHeader(Buffer buffer, Rect area, int y) {
         }
 
         YearMonth yearMonth = YearMonth.from(displayDate);
-        String monthName = yearMonth.getMonth().getDisplayName(TextStyle.FULL, locale);
```
maxandersen, line 235: FULL to FULL_STANDALONE is the right fix.

But the capitalization seems wrong - that would also be locale specific I guess?

maxandersen: like, I'm not polish so you tell me but it seems polish locale does NOT capitalize months?

```
pl | FULL=czerwca | FULL_STANDALONE=czerwiec
da | FULL=juni | FULL_STANDALONE=juni
tr | FULL=Haziran | FULL_STANDALONE=Haziran
de | FULL=Juni | FULL_STANDALONE=Juni
en | FULL=June | FULL_STANDALONE=June
```

maxandersen: but i guess here in the UI its "starting a sentence" so we need locale correct uppercasing

tamboui-widgets/src/main/java/dev/tamboui/widgets/calendar/Monthly.java
```diff
@@ -330,6 +340,20 @@ private void renderCalendarGrid(Buffer buffer, Rect area, int startY) {
...
+     *
+     * @param str the string to capitalize
+     * @return the string with the first letter converted to uppercase,
+     *         or the original string if it is {@code null} or empty
+     */
+    private static String capitalize(String str) {
+        if (str == null || str.isEmpty()) {
+            return str;
+        }
+        return Character.toUpperCase(str.charAt(0)) + str.substring(1);
```
maxandersen: this should use str.substring(0,1).toUpperCase(locale) to handle language differences.

maxandersen: actually - that isn't enough if codepoint is wider than 1 character.

maxandersen:
```suggestion
    private static String capitalize(String s, Locale locale) {
if (s == null || s.isEmpty()) return s;
int firstCp = s.codePointAt(0);
int firstLen = Character.charCount(firstCp);
String first = new String(Character.toChars(firstCp)).toUpperCase(locale);
return first + s.substring(firstLen);
```

something like that should handle other languages

tamboui-widgets/src/main/java/dev/tamboui/widgets/calendar/Monthly.java
lukasz-kapica: Yes, unlike in English in a middle of a sentence it should be lowercase... But this is not a middle of a sentence but a heading.

maxandersen:
Much appreciated @lukasz-kapica !

PR #274, opened 2026-02-28
feat: add JFR tracing events

maxandersen:
This is an alternative of the tracing in #190 and #169.

Instead of us implementing custom tracing just use the JDK's built jfr system.

Even works on Java 8 using a compile time only polyfill.

Add JFR events for terminal rendering and toolkit event routing.

- Emit dev.tamboui.terminal.draw around Terminal.draw()
- Emit dev.tamboui.toolkit.* routing/candidate/focus/drag/global-handler events
- Add compileOnly jfr-polyfill for Java 8 builds
- Document JFR usage and naming guidelines

with this you can use jfr tools to monitor all the tools and can build our own custom tamboui listener.

draft for now. until get docs in place.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/jfr/TerminalDrawEvent.java
+++ b/tamboui-core/src/main/java/dev/tamboui/jfr/TerminalDrawEvent.java
@@ -10,6 +10,7 @@ import jdk.jfr.Event;
 import jdk.jfr.EventType;
 import jdk.jfr.Label;
 import jdk.jfr.Name;
+import jdk.jfr.Threshold;
 
 /**
  * JFR event measuring time spent in {@code Terminal.draw()}.


maxandersen:
> * can you check that we have the proper module info require clauses?

any specific concerns here? my understanding is jfr is part of base?

maxandersen:
about `--native` I hit this issue https://github.com/tamboui/tamboui/issues/276 but its there even when running on main.

maxandersen:
> > * can you check that we have the proper module info require clauses?
> 
> any specific concerns here? my understanding is jfr is part of base?

those using jfr need to put  `requires jdk.jfr;` - verified manually for now but would be nice our build would catch these so opened #278

maxandersen:
There is a flakey test failing: PilotTest counter app.

I'll merge this and go other build where adding info to build to get build failure info.

maxandersen:
btw. for native-image just need to include jfr in on native image build then it works. If you dont have it then there will be a classloader error  when running native-image as jfr classes not included.

PR #275, opened 2026-02-28
Add locale-based weekday abbreviations to Monthly

lukasz-kapica:
Continuing my quest on localizing the calendar widget. This time the weekday abbreviations which were hardcoded english abbreviations.

### DIFF
--- a/tamboui-widgets/src/main/java/dev/tamboui/widgets/calendar/Monthly.java
+++ b/tamboui-widgets/src/main/java/dev/tamboui/widgets/calendar/Monthly.java
@@ -374,7 +374,7 @@ public final class Monthly implements Widget {
      */
     private static String[] getWeekdayAbbrev(Locale locale) {
       return Arrays.stream(DayOfWeek.values())
-              .map(dayOfWeek -> dayOfWeek.getDisplayName(TextStyle.FULL_STANDALONE, locale))
+              .map(dayOfWeek -> dayOfWeek.getDisplayName(TextStyle.SHORT_STANDALONE, locale))
               .map(dayOfWeek -> capitalize(dayOfWeek, locale))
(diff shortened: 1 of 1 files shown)


tamboui-widgets/src/main/java/dev/tamboui/widgets/calendar/Monthly.java
```diff
@@ -356,6 +358,62 @@ private static String capitalize(String s, Locale locale) {
...
+     * The returned array contains seven elements, one for each day of the week,
+     * ordered from Monday to Sunday as defined by {@link DayOfWeek#values()}.
+     *
+     * @param locale the locale to use for day-of-week display names
+     * @return a seven-element array of two-character weekday abbreviations
+     * @see DayOfWeek#getDisplayName(TextStyle, Locale)
+     */
+    private static String[] getWeekdayAbbrev(Locale locale) {
+      return Arrays.stream(DayOfWeek.values())
+              .map(dayOfWeek -> dayOfWeek.getDisplayName(TextStyle.FULL_STANDALONE, locale))
```
maxandersen: good feature but should we maybe make use of the java locale SHORT or NARROW directly over expliclity 2 chars width ? 

or use NARROW instead of FULL_STANDALONE as basis for 2 char limit?

lukasz-kapica: I wanted to be consistent with the existing hardcoded array. NARROW returns a single character for most locales. For english it will return S for both Saturday and Sunday so it's a bit ambiguous. SHORT_STANDALONE works better, it gives exactly the same array for 739 locales and different for 330 locales. For polish Friday it will return Pt instead of Pi which is better. I will change it for SHORT_STANDALONE.

PR #282, opened 2026-03-05
feat: add synchronized output (Mode 2026) to prevent screen tearing

stalep:
Upgrade aesh-terminal to 3.3 and wrap Terminal.draw() with Mode 2026 BSU/ESU sequences so the terminal buffers all frame output and renders it atomically. This eliminates visible tearing on supported terminals (Kitty, Ghostty, WezTerm, Foot, Contour, iTerm2, Mintty) with no impact on unsupported ones which safely ignore the escape sequences.

ref; https://aeshell.github.io/docs/readline/synchronized-output/

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 jline = "3.25.1"
-aesh-terminal = "3.2"
+aesh-terminal = "3.5"
 junit = "5.14.3"
 assertj = "3.27.7"
 native-build-tools = "0.11.4"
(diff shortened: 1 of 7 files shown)


maxandersen:
@stalep this looks interesting - what did you use to test ? 
for me i'm getting 50% slower framerate on the RGB color demo which I didn't expect

stalep:
Hm, this should be a no-op if the terminal does not support it. Which backend did you test against?

stalep:
@maxandersen there were a bug in the pr, can you please try again?

maxandersen:
it does seem to work better now but I get blank frames/screens 10-20% of the time - that does not happen on main.

tried with tree-demo and filemanager-demo

PR #267, opened 2026-02-24
fix: Column rendering prevents wrapping from working correctly

melix:
This commit fixes how the column interprets min contraints. In case a child has a preferred size, for example for text items that wrap text, the preferred height should be used, otherwise the default constraint which is height 1 will cause collapsing of elements. This is the case when a column has both text and spacers, because spacers will eat up all available space, ignoring the preferred height.

Fixes #261

PR #295, opened 2026-03-20
fix: Fix solver to not fail by default

melix:
This commit fixes a bug introduced in #267, where a column could define multiple min constraints which, when summed up, could exceed the available space. Technically speaking this commit fixes it in 3 complementary ways:

1. a fix in the column code itself, to not exceed the available height. This works, but doesn't prevent the solver from failing in other cases
2. fix the solver, so that min constraints are not using a strong constraint (required). This also fixes the same problem, but doesn't prevent the solver from failing in other cases
3. add a lenient mode to the solver, enabled by default, which soften required constraints whenever constraints cannot be solved.

The lenient mode can be overridden by a system property, when debugging a layout is needed. All 3 fixes combined should make our solver more usable.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/layout/cassowary/Solver.java
+++ b/tamboui-core/src/main/java/dev/tamboui/layout/cassowary/Solver.java
@@ -83,10 +83,31 @@ public final class Solver {
     // Artificial objective for handling infeasibility
     private Row artificial;
 
+    private static final boolean DEFAULT_STRICT = Boolean.getBoolean("tamboui.solver.strict");
+
+    // Strongest non-required strength, used to soften unsatisfiable REQUIRED constraints in lenient mode
+    private static final Strength SOFTENED_REQUIRED = Strength.create(999, 999, 999);
+
+    private final boolean strict;
+
     /**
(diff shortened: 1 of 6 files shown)


tamboui-core/src/main/java/dev/tamboui/layout/cassowary/Solver.java
```diff
@@ -101,10 +122,20 @@ public Solver() {
...
      * @throws DuplicateConstraintException     if the constraint already exists
-     * @throws UnsatisfiableConstraintException if required and unsatisfiable
+     * @throws UnsatisfiableConstraintException if required, unsatisfiable, and solver is in strict mode
      */
     public void addConstraint(CassowaryConstraint constraint) {
-        addConstraintInternal(constraint);
+        if (!tryAddConstraint(constraint) && constraint.isRequired()) {
+            if (strict) {
+                throw new UnsatisfiableConstraintException(constraint);
+            }
```
maxandersen, line 131: could make a jfr event here recording the failing constraint :)

maxandersen (approved):

PR #297, opened 2026-03-22
Add semantic color system with CSS-first theming and fluent Color API

maxandersen:
This is draft-in-progress to explore notion of semantic color variables (which is closely related to Themes) 

This introduce set of variables that gets derived from a base color/theme similar to textual.

All starts from a primaryColor - everything else is derivable from there but possible to be defined in a theme.

This mainly add the color conversions needed + a "theme system" which is nothing but a programmatic typesafe api 
to generate a css that can be used by css system.

Looking for input on it.

Things not yet solved:
- themes/variables not used consistently in widget library today; would need to at least use these names as fallback
- MarkupParser has a styleresolver but we create a new everytime needed; rather than using styles so things dont "listen" to the css applied in componenttree; for now CssMarkupBridge but does not feel rightl
- ansi color interaction; should/could we map ansi index in a terminal theme to semantic names and vice-versa? idea being apps that dont want their own theme will "fit" automatically into  users terminal theme.

details:
Enables theme-independent applications using semantic color names ($primary, $error, etc.) instead of hardcoded values. Applications work with any theme without code changes.

Key features:
- Fluent Color API: color.lighten(0.2f).rotateHue(30).mix(white, 0.3f)
- Theme.generate(): Creates complete palettes from minimal input (inspired by Textual's ColorSystem)
- ColorSpace utilities: HSL/HSV conversion and color manipulation
- CssMarkupBridge: Use CSS variables in markup text [primary]Title[/]
- 7 themes included: Catppuccin, Dracula, Gruvbox, Nord, Tokyo Night, Dark, Light

The fluent API on Color delegates to ColorSpace for implementation, keeping ColorConverter focused on string parsing and ColorSpace on value transformations.

Co-Authored-By: Claude Sonnet 4.5

### DIFF
--- a/tamboui-css/src/main/java/dev/tamboui/css/theme/ThemePropertyNames.java
+++ b/tamboui-css/src/main/java/dev/tamboui/css/theme/ThemePropertyNames.java
@@ -1,0 +1,142 @@
+package dev.tamboui.css.theme;
+
+/**
+ * Documents the standard semantic design tokens that themes should define.
+ * This is documentation only - actual values are defined in .tcss theme files.
+ *
+ * <p>All theme .tcss files should define these variables for consistency.
+ * Applications should use these semantic names instead of hardcoded colors
+ * to ensure theme independence.</p>
+ *
+ * <p>Usage example:</p>
+ * <pre>{@code
+ * // In markup
+ * Text.from("[" + SemanticTokens.PRIMARY + "]Title[/]")
+ *
+ * // In CSS
+ * Panel { border-color: $primary; }
+ * }</pre>
+ */
+public final class ThemePropertyNames {
+
+    // Prevent instantiation - documentation only
+    private ThemePropertyNames() {}
+
+    // ===== Core Brand Colors =====
+
+    /**
+     * Primary brand color.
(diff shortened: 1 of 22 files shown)


tamboui-css/src/main/java/dev/tamboui/css/theme/ThemePropertyNames.java
```diff
@@ -0,0 +1,142 @@
...
+ * This is documentation only - actual values are defined in .tcss theme files.
+ *
+ * <p>All theme .tcss files should define these variables for consistency.
+ * Applications should use these semantic names instead of hardcoded colors
+ * to ensure theme independence.</p>
+ *
+ * <p>Usage example:</p>
+ * <pre>{@code
+ * // In markup
+ * Text.from("[" + SemanticTokens.PRIMARY + "]Title[/]")
```
heuermh, line 14:
```suggestion
 * Text.from("[" + ThemePropertyNames.PRIMARY + "]Title[/]")
```

maxandersen:
closing as came up with a better approach. will push soon.

PR #299, opened 2026-03-24
feat: make TextInput and TextArea elements binding-aware

dwalluck:
TextInputElement and TextAreaElement unconditionally reject all key events with Ctrl/Alt modifiers so configured binding sets are not respected in text editing fields.

This commit checks for modifier keys in the active binding set before rejecting them.

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
@@ -16,6 +16,7 @@ import dev.tamboui.toolkit.element.RenderContext;
 import dev.tamboui.toolkit.element.Size;
 import dev.tamboui.toolkit.element.StyledElement;
 import dev.tamboui.toolkit.event.EventResult;
+import dev.tamboui.tui.bindings.Actions;
 import dev.tamboui.tui.event.KeyEvent;
 import dev.tamboui.widgets.block.Block;
 import dev.tamboui.widgets.block.BorderType;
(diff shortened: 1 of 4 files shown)


tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
```diff
@@ -347,9 +348,9 @@ private static boolean handleTextAreaKey(TextAreaState state, KeyEvent event) {
                 state.insert("    "); // 4 spaces for tab
                 return true;
             case CHAR:
-                // Don't consume characters with Ctrl or Alt modifiers - those are control sequences
+                // Check for modifier keys in the active binding set
                 if (event.modifiers().ctrl() || event.modifiers().alt()) {
-                    return false;
+                    return handleTextAreaAction(state, event);
```
maxandersen, line 353: this does not seem right - the intent afair is that keybindings generated events should be handled *before* they hit this place.

@melix wdyt?

dwalluck: I took the pragmatic approach of handling the keys inside the element's key handler where they're already being handled.

If you're talking about a layer above that in `EventRouter::routeKeyEvent`, then that's interesting. If I understand correctly, that means that most keys would just fall through anyway, and only modifiers get handled there.

For my use case, I want to add more emacs/readline support which currently would hit either `Toolkit` or `Text*State`, so I would like to know the proper pattern before implementing it so that I don't waste time implementing it "incorrectly".

dwalluck: Even if you have global application-level keybindings (and you should), most keybindings still have to fall through to the component-level handler anyway.

Readline features really only apply to a text input component, although some of the keybindings could be shared between text area and text input.

PR #301, opened 2026-03-29
fix: enable mouse support on Windows via JLine3 native tracking

xsShuang:


tamboui-jline3-backend/src/main/java/dev/tamboui/backend/jline3/JLineBackend.java
```diff
@@ -148,6 +159,11 @@ public void enableMouseCapture() throws IOException {
 
     @Override
     public void disableMouseCapture() throws IOException {
+        // Disable JLine3 native mouse tracking (Windows only)
+        if (IS_WINDOWS) {
```
maxandersen, line 163: is there a reason to not just do this on all platforms?

PR #303, opened 2026-04-03
feat: Make text area available for form fields

jtama:
Fixes #292 

Ok, it was a _bit_ more work than I initially thought, but it works.

### DIFF
--- a/demos/form-table-demo/src/main/java/dev/tamboui/demo/layout/FormTableDemo.java
+++ b/demos/form-table-demo/src/main/java/dev/tamboui/demo/layout/FormTableDemo.java
@@ -37,11 +37,16 @@ public final class FormTableDemo {
     private static final AtomicInteger TAB_INDEX = new AtomicInteger(0);
     private static final TableState TABLE_STATE = new TableState();
 
+    private static String BIO = "Augusta Ada King, Countess of Lovelace (née Byron; 10 December 1815 – 27 November 1852), also known as Ada Lovelace, was an English mathematician and writer chiefly known for work on Charles Babbage's proposed mechanical general-purpose computer, the analytical engine. She was the first to recognise the machine had applications beyond pure calculation. Lovelace is often considered the first computer programmer."+
+            System.lineSeparator() +
+            "Lovelace was the only legitimate child of poet Lord Byron and reformer Anne Isabella Milbanke.[2] Lord Byron separated from his wife a month after Ada was born, and died when she was eight. Although often ill in childhood, Lovelace pursued her studies assiduously. She married William King in 1835. King was a Baron, and was created Viscount Ockham and 1st Earl of Lovelace in 1838. The name Lovelace was chosen because Ada was descended from the extinct Baron Lovelaces.[3] The title given to her husband thus made Ada the Countess of Lovelace.";
+
     // Centralized form state for all fields
     private static final FormState FORM = FormState.builder()
             // Profile fields (text)
             .textField("fullName", "Ada Lovelace")
             .textField("email", "ada@analytical.io")
+            .textAreaField("bio", BIO)
             // Preferences fields (boolean + select)
             .selectField("theme", Arrays.asList("Nord", "Dracula", "Solarized", "Monokai"), 0)
             .booleanField("notifications", true)
@@ -55,6 +60,7 @@ public final class FormTableDemo {
     private static final FormState VALIDATION_FORM = FormState.builder()
             .textField("username", "")
             .textField("email", "")
+            .textField("bio", "")
             .maskedField("password", "")  // masked field for password
             .textField("age", "")
             .textField("phone", "")
@@ -177,12 +183,15 @@ public final class FormTableDemo {
                         formField("Email", FORM.textField("email"))
                                 .id("email").labelWidth(LABEL_WIDTH).rounded()
                                 .borderColor(Color.DARK_GRAY).focusedBorderColor(Color.CYAN)
+                                .arrowNavigation(true),
+                        formField("Bio", FORM.textAreaField("bio"), 4)
+                                .id("bio").labelWidth(LABEL_WIDTH).rounded()
+                                .borderColor(Color.DARK_GRAY).focusedBorderColor(Color.CYAN)
                                 .arrowNavigation(true)
                 ))
                 .rounded()
                 .borderColor(Color.CYAN)
                 .fit(),
-
                 row(
                         panel("Preferences (Select + Toggle)", column(
                                 formField("Theme", FORM.selectField("theme"))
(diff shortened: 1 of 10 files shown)


jtama:
Anyone for a review ?

melix (commented):
Just pasting Claude's review for now, I didn't have time to take a look yet or if the comments make sense.

Real bugs                                                                                                                                                                                                                         
                                                                                                                                                                                                                                    
  1. preferredSize regression for all form fields (FormFieldElement.java:596 in PR diff):                                                                                                                                           
  int totalHeight = height + areaHeight;
  return Size.of(width, Math.max(availableHeight, totalHeight));                                                                                                                                                                    
  1. For non-text-area fields areaHeight == 0, but Math.max(availableHeight, totalHeight) now makes every field claim the entire available height. That'll cascade through any column(...) that does sizing — non-text-area fields  
  will stop being 1 or 3 rows tall. The previous return Size.of(width, height) was correct. Most likely this should be Size.of(width, totalHeight).                                                                                 
  2. Unreachable Ctrl+Enter submit branch (FormFieldElement.handleTextAreaFieldKey):                                                                                                                                                
  if (event.isConfirm()) { /* append newline */ return HANDLED; }                                                                                                                                                                   
  if (event.hasCtrl() && event.isConfirm()) { /* validate + onSubmit */ return HANDLED; }                                                                                                                                           
  2. The second branch is dead — Ctrl+Enter still matches isConfirm() by default, so it's caught by the first branch. The two checks are also reversed from the comments' intent.                                                   
  3. Newline inserted at end of buffer, not at cursor (same method): setText(text() + lineSeparator()) then moveCursorToEnd(). If the cursor is in the middle of the text, the newline is appended at the bottom and the cursor is  
  yanked to the end. state.insert('\n') is the right call (and that's what Toolkit.handleTextAreaKey and TextAreaElement do).                                                                                                       
  4. Off-by-one in arrow-navigation escape (same method):                                                                                                                                                                           
  if (event.isDown() && textAreaState.cursorRow() == textAreaState.lineCount()) {                                                                                                                                                   
      return FOCUS_NEXT;
  }                                                                                                                                                                                                                                 
  4. cursorRow ranges over 0..lineCount-1 (verified in TextAreaState.moveCursorDown). cursorRow == lineCount is never true, so Down at the bottom never moves focus to the next field.
  5. Newline-on-Enter requires Shift in Toolkit.handleTextAreaKey (Toolkit.java):                                                                                                                                                   
  if (event.matches(Actions.CONFIRM) && event.hasShift()) {                                                                                                                                                                         
      state.insert('\n');                                                                                                                                                                                                           
      return true;                                                                                                                                                                                                                  
  }               
  5. The comment says "Always insert newline on Enter", but the guard requires Shift. So plain Enter does nothing here. TextAreaElement works around this by intercepting CONFIRM before delegating, but any future caller of       
  handleTextAreaKey will have a non-functional Enter. Either drop the hasShift() check or invert the policy and document it.                                                                                                        
  6. Dead identical branches in Toolkit.handleTextAreaKey:
  if (event.matches(Actions.MOVE_UP)) {                                                                                                                                                                                             
      if (event.hasShift()) state.moveCursorUp();
      else state.moveCursorUp();                                                                                                                                                                                                    
      return true;
  }                                                                                                                                                                                                                                 
  6. Both branches do the same thing. Same for MOVE_DOWN. Comments say "could be used for selection in future". This is speculative scaffolding — drop it.
  7. PAGE_UP / PAGE_DOWN hardcoded to 10 calls:                                                                                                                                                                                     
  for (int i = 0; i < 10; i++) state.moveCursorUp();                                                                                                                                                                                
  7. No relation to actual viewport. Each moveCursorUp re-clamps cursorCol. Either thread the rendered viewport through or skip Page-Up/Down until you can.                                                                         
  8. Demo's validation form uses a single-line textField for "Bio" (FormTableDemo.java): the whole point of the PR is text areas in forms — the validation demo should exercise validation against a textAreaField, not a regular   
  text field with the same name.                                                                                                                                                                                                    
                                                                                                                                                                                                                                    
  Code-quality smells                                                                                                                                                                                                               
                  
  - Fully-qualified names instead of an import: dev.tamboui.tui.event.KeyCode.BACKSPACE, KeyCode.DELETE, KeyCode.TAB, KeyCode.CHAR are written out long-hand all over Toolkit.handleTextAreaKey. The file already imports KeyEvent; 
  just add import dev.tamboui.tui.event.KeyCode;.
  - Param/field name mismatch: constructor parameter maxHeight stored in field areaHeight; FormElement.field(..., int maxHeight, …) flows into FieldConfig.maxHeight. Pick one name.                                                
  - FieldConfig.maxHeight is leaky: it's only meaningful for TEXT_AREA. Two reasonable options: a dedicated textAreaField(name, label, maxHeight, …) builder method on FormElement, or push maxHeight into a small                  
  TEXT_AREA-specific config. As-is, every FieldConfig carries a value that's silently ignored except for one branch.                                                                                                                
  - Stray double blank line in FormState.Builder between textAreaFields declaration and the new textAreaField method, and between field(...) overloads in FormElement. spotless will likely yell.                                   
  - renderTextArea inflates the rect: new Rect(area.x(), area.y(), area.width(), area.height() + areaHeight). Combined with the broken preferredSize, you're double-counting. The renderer should respect the area it's given;      
  sizing belongs in preferredSize.                                                                                                                                                                                                  
  - textValues() collisions: text-area names go into the same flat map as text-field names. No guard against name collisions either at builder time or when iterating. Pre-existing pattern, but the surface area just doubled.     
  - Test coverage is thin: one test (for onSubmit). Nothing for Toolkit.handleTextAreaKey semantics, FormState.textAreaField round-trip, the FOCUS_NEXT/PREVIOUS boundary in FormFieldElement, or preferredSize.                    
                                                                                                                                                                                                                                    
  Could it have been easier?                                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  Yes, in a few places.                                                                                                                                                                                                             
   
  - Don't move handleTextAreaKey to Toolkit yet. It now has two callers (TextAreaElement and FormFieldElement), both inside tamboui-toolkit. A package-private static helper would do, deferring the public-API commitment until    
  there's actually third-party demand. As a Toolkit public method it'll be hard to evolve later (the current shape with shift-gated newline insertion is something you'll want to change).
  - Skip the PAGE_UP/PAGE_DOWN and "future selection" hooks. They don't work properly; the surrounding code reads cleaner without them.                                                                                             
  - Don't split TEXT/TEXT_AREA so aggressively. The diff diverges two new private methods (handleTextFieldKey vs handleTextAreaFieldKey, renderTextInput vs renderTextArea) that are 80% identical. Pulling out a small helper for  
  "resolve effective border color + maybe build a Block" would keep them shorter and reduce the chance of one drifting from the other (which has already happened — the FormField text-area path's Enter/cursor handling is wrong   
  precisely because it diverged from the text path).                                                                                                                                                                                
  - Test coverage is thin: one test (for onSubmit). Nothing for Toolkit.handleTextAreaKey semantics, FormState.textAreaField round-trip, the FOCUS_NEXT/PREVIOUS boundary in FormFieldElement, or preferredSize.

maxandersen:
@jtama do you want to take a stab at updating based on the comment/feedback ? mind you - they range from tiny things to actual bugs - if something is too nitpicky just state so and we can make sure we handle the big ones.

jtama:
yes sorry, didn't had much time since I opened the PR. I will try to rebase, and then start updating.

jtama:
To be more specific, I won't have much time for the few weeks to come. So if someone wants to takeover...

jtama:
Ok so I think much of the comments. If you want to take another look...

PR #310, opened 2026-04-24
fix: let the app have multiple render threads

### LINKED ISSUE #309: multiple TuiRunner capability
currently RenderThread has a check if rendering is on same thread - that makes things problematic for multiple tuirunners in case of web/ssh connections

Ideas:
- be able to turn of the check
- add "threadpool" of allowed threads to handle 50 users with 2 threads
- use virtual threads? 
- use name pattern to bless threads (i.e. "tamboui-render-x")
- use ThreadLocal<Boolean>, if present ok, if not stop

melix:
This commit updates the behavior of the `RenderThread` utility, so that it checks if the current thread is marked as a render thread or not.

Technically speaking, the behavior is similar to what we had before: calling the `run()` method on the TuiRunner automatically sets the current thread as the render thread. The difference is that if there are *multiple* runners, then we won't overwrite the render thread anymore. Therefore, the callers are free to start the event loop on whatever thread they want. The Aesh demo has therefore been modified to use virtual threads for the event loop.

Closes #309

### DIFF
--- a/tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/AbstractElementTest.java
+++ b/tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/AbstractElementTest.java
@@ -4,9 +4,10 @@
  */
 package dev.tamboui.toolkit;
 
-import dev.tamboui.tui.RenderThread;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+
+import dev.tamboui.tui.RenderThread;
 
 public abstract class AbstractElementTest {
 
--- a/tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/AbstractElementTest.java
+++ b/tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/AbstractElementTest.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.toolkit;
+
+import dev.tamboui.tui.RenderThread;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+public abstract class AbstractElementTest {
+
+    @BeforeEach
+    void setUpRenderThread() {
+        RenderThread.markAsRenderThread();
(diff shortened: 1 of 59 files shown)


maxandersen:
AbstractElementTest class is missing?

maxandersen (changes requested):
not compiling as tet baseclass spitting

melix:
PR passing now

tamboui-tui/src/main/java/dev/tamboui/tui/RenderThread.java
```diff
@@ -34,57 +34,49 @@
...
 
     /**
-     * Sets the render thread. Package-private for use by TuiRunner.
-     *
-     * @param thread the thread to set as the render thread
+     * Marks the current thread as a render thread. Package-private for use by TuiRunner.
      */
-    static void setRenderThread(Thread thread) {
-        renderThread.set(thread);
+    public static void markAsRenderThread() {
```
maxandersen: can we really not avoid making this public?

I assume the reason this now need to be public is that we need to be on the thread to set the thread local?

if must be public should clarify the package-private part for use by tuirunner.

tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/AbstractElementTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+import dev.tamboui.tui.RenderThread;
+
+public abstract class AbstractElementTest {
+
+    @BeforeEach
+    void setUpRenderThread() {
+        RenderThread.markAsRenderThread();
```
maxandersen: why is it that we need this now and not before?

maxandersen (commented):
looks simple enough; just not grokking why the tests needs changing for this change?

tamboui-toolkit/src/testFixtures/java/dev/tamboui/toolkit/AbstractElementTest.java
melix: Because before, when no `Thread` was set as the UI thread, we would not perform any check and silently pass. Now we mandate that there is at least one UI thread.

PR #315, opened 2026-04-30
Fix blurry image rendering for native protocols

gnodet:
## Summary

- **Native protocols** (Kitty, iTerm2) now receive the original full-resolution image instead of a pre-scaled copy. The terminal handles a single clean downscale pass via `c=/r=` (Kitty) or `width=/height=` (iTerm2), eliminating the double-scaling blur (app downscale → terminal upscale)
- Added `q=2` (quiet mode) to Kitty protocol escape sequences to suppress terminal response messages
- Upgraded character-based protocol resize interpolation from bilinear to bicubic

## Details

Images rendered via native protocols appeared blurry because `Image.render()` was pre-scaling them to a tiny pixel grid (`cols×8 × rows×16`) before sending them to the terminal. The terminal then scaled this small image back up to fill the actual display area — two scaling passes that degraded quality.

The fix splits the render path:
- **Native protocols**: send the original image, compute a display area in cells respecting the scaling mode (FIT/FILL/STRETCH/NONE) with cell aspect ratio correction
- **Character-based protocols**: unchanged (pre-scale to virtual pixel grid)

## Test plan

- [x] All 71 existing `ImageTest` tests pass
- [ ] Visual comparison in Kitty-compatible terminal (WezTerm, Ghostty, Kitty)

### DIFF
--- a/tamboui-image/src/main/java/dev/tamboui/image/Image.java
+++ b/tamboui-image/src/main/java/dev/tamboui/image/Image.java
@@ -133,11 +133,10 @@ public final class Image implements Widget, RawOutputCapable {
             return;
         }
 
-        if (protocol.requiresRawOutput()) {
-            // Native protocols (Kitty, iTerm2): send original full-resolution image
-            // and let the terminal handle pixel scaling via c=/r= or width=/height=.
-            // This avoids double scaling (app downscale then terminal upscale) which
-            // causes blurry rendering.
+        if (protocol.handlesOwnScaling()) {
+            // Kitty, iTerm2: send original full-resolution image and let the terminal
+            // handle pixel scaling via c=/r= or width=/height=. This avoids double
+            // scaling (app downscale then terminal upscale) which causes blurry rendering.
             Rect displayArea = computeNativeDisplayArea(data, imageArea);
             try {
                 protocol.render(data, displayArea, buffer, rawOutput);
@@ -145,7 +144,7 @@ public final class Image implements Widget, RawOutputCapable {
                 throw new RuntimeIOException("Failed to render image using protocol " + protocol.name(), e);
             }
         } else {
-            // Character-based protocols (half-block, braille): pre-scale to pixel grid
+            // Sixel and character-based protocols (half-block, braille): pre-scale to pixel grid
             ImageProtocol.Resolution res = protocol.resolution();
             int gridWidth = imageArea.width() * res.widthMultiplier();
             int gridHeight = imageArea.height() * res.heightMultiplier();
@@ -183,9 +182,11 @@ public final class Image implements Widget, RawOutputCapable {
                     displayRows = area.height();
                     displayCols = Math.max(1, (int) Math.round(displayRows * cellPxH * imgAspect / cellPxW));
                 }
-                return new Rect(area.x(), area.y(),
-                    Math.min(displayCols, area.width()),
-                    Math.min(displayRows, area.height()));
+                int fitW = Math.min(displayCols, area.width());
+                int fitH = Math.min(displayRows, area.height());
+                int offsetX = (area.width() - fitW) / 2;
+                int offsetY = (area.height() - fitH) / 2;
+                return new Rect(area.x() + offsetX, area.y() + offsetY, fitW, fitH);
             }
             case FILL:
             case STRETCH:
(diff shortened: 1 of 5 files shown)


davsclaus (approved):

maxandersen:
Cool. Got an example with image where this shows ?

Also not sure if good we don't do any bounds checking since the current implementation at least avoids shipping potential massive images across the terminal protocol. 

But maybe that is better the user does that scaling if he really needs it.

gnodet:
> Cool. Got an example with image where this shows ?
> 
> Also not sure if good we don't do any bounds checking since the current implementation at least avoids shipping potential massive images across the terminal protocol.
> 
> But maybe that is better the user does that scaling if he really needs it.

See https://github.com/apache/camel/pull/22856

With stock TamboUI 0.2.0, the image is displayed as:
image

tamboui-image/src/main/java/dev/tamboui/image/protocol/KittyProtocol.java
```diff
@@ -114,7 +114,7 @@ private void sendChunked(OutputStream out, String base64Data, int cols, int rows
                 // c=cols: display width in cells
                 // r=rows: display height in cells
                 // m=0/1: more chunks follow
-                cmd.append(String.format("a=T,f=100,t=d,c=%d,r=%d,m=%d;", cols, rows, more ? 1 : 0));
+                cmd.append(String.format("a=T,f=100,t=d,q=2,c=%d,r=%d,m=%d;", cols, rows, more ? 1 : 0));
```
maxandersen, line 123: is q=2 just  an optimization ? or have some practical reason?

tamboui-image/src/main/java/dev/tamboui/image/ImageData.java
```diff
@@ -273,7 +273,7 @@ public ImageData resize(int newWidth, int newHeight) {
         BufferedImage resized = new BufferedImage(newWidth, newHeight, BufferedImage.TYPE_INT_ARGB);
         Graphics2D g = resized.createGraphics();
         try {
-            g.setRenderingHint(RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BILINEAR);
+            g.setRenderingHint(RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BICUBIC);
```
maxandersen: so that scaling is 2-3x slower than linear - do we really want that default?

tamboui-image/src/main/java/dev/tamboui/image/Image.java
```diff
@@ -133,18 +133,71 @@ public void render(Rect area, Buffer buffer, OutputStream rawOutput) {
             return;
         }
 
-        // Scale image based on scaling mode and protocol resolution
+        if (protocol.requiresRawOutput()) {
```
maxandersen: this is the key fix in here and makes sense to me. the quality difference is very noticable. By not scaling here default is prettier and user can always do prescale if for whatever reason he wants to use less pixles.

tamboui-image/src/main/java/dev/tamboui/image/ImageData.java
maxandersen: i would in this change keep bilinear and then if need differnt one introdue option to select scaling hint.

maxandersen (changes requested):

gnodet:
Thanks for the review! Addressed the feedback:
- Reverted bicubic → bilinear (keeping it out of this PR as suggested)
- Added inline comment explaining `q=2` (suppresses terminal responses on stdin)

Re: bounds checking for large images — agree that the user controls input data and can pre-scale if needed. The point of this fix is to avoid *involuntary* downscaling that causes blur.

gnodet (commented):
Addressed review feedback:

**`q=2` (KittyProtocol)**: Not just an optimization — without it the terminal sends OK/error responses (e.g. `\033_Gi=…;OK\033\\`) back on stdin for every graphics command, which can interfere with input parsing. `q=2` tells the terminal to stay quiet. Added inline comment explaining this.

**Bicubic → reverted to bilinear (ImageData)**: Agreed, keeping that change out of this PR. If needed later, can introduce an option to select the scaling hint.

maxandersen (approved):

PR #316, opened 2026-04-30
Fix image aspect ratio distortion in Kitty and iTerm2 protocols

gnodet:
## Summary
- Fix 2x horizontal stretching of images when using Kitty/iTerm2 native protocols
- Compute actual cell dimensions from the scaled image pixel size instead of using the full display area
- Center images within the available area when they don't fill it completely

## Details
`Image.scaleImage()` pre-scales images to fit within the pixel grid while preserving aspect ratio. For portrait/tall images, the scaled result is narrower than the grid. However, `KittyProtocol` and `ITermProtocol` were passing the **full area** dimensions (`c=area.width(), r=area.height()`) to the terminal, which then stretched the narrower image to fill, distorting the aspect ratio.

## Test plan
- [x] All 71 existing tests pass
- [x] Visual verification with Camel route diagrams in Ghostty (Kitty protocol)
- [ ] Visual verification in iTerm2

### DIFF
--- a/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
+++ b/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
@@ -64,8 +64,22 @@ public final class ITermProtocol implements ImageProtocol {
             return;
         }
 
-        // Move cursor to position
-        String cursorMove = String.format("\033[%d;%dH", area.y() + 1, area.x() + 1);
+        // Compute the actual cell dimensions from the (already scaled) image pixel size.
+        // Same logic as KittyProtocol: the image was pre-scaled to fit within the pixel
+        // grid while preserving aspect ratio, so we must use the actual image cell size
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

maxandersen:
Got example to use for verification?

gnodet:
> Got example to use for verification?

See https://github.com/apache/camel/pull/22856

With stock TamboUI 0.2.0, the image is displayed as:
image

maxandersen (changes requested):
should this not be done for Sixel too?

the compute code is the same - with sixel there would be 3 of them; feels like this should be shared code.

Also; this is overlapping with changes in #315 afaics? maybe merge/clean them up?

gnodet:

Closing in favor of #315, which takes a better approach by moving the native-vs-character protocol split into `Image.java` and avoiding double-scaling entirely. The centering logic and Sixel handling from this PR will be folded into #315.

PR #317, opened 2026-04-30
fix: clean up Kitty graphics images when Image widget is removed

gnodet:
## Summary
- Track areas where `RawOutputCapable` widgets render per frame via `Frame.rawOutputAreas()`
- When transitioning from a frame with raw output to one without:
  - **Kitty**: send delete-all command (`\033_Ga=d,d=a\033\`) to clear the separate graphics layer
  - **iTerm2/Sixel**: overwrite previous image areas with spaces via raw output, since these protocols render images inline in cells that the buffer diff system doesn't track
- Also clean up on terminal resize, since images persist across screen clears

## Context
Raw-output image protocols (Kitty, iTerm2, Sixel) bypass the text buffer and write directly to the terminal. When an `Image` widget is rendered in one frame but absent from the next, the buffer diff detects no change for those cells, so no overwrite is sent — the image lingers on screen.

Kitty is the worst case because its graphics layer is fully independent of the text buffer. iTerm2 and Sixel render inline but the cells aren't tracked by the buffer, producing the same effect.

Braille and HalfBlock protocols are unaffected — they write Unicode characters directly into the buffer, so the diff handles cleanup naturally.

## Test plan
- [ ] Display an `Image` widget using Kitty protocol, then navigate away — image should disappear
- [ ] Same test with iTerm2 protocol
- [ ] Resize the terminal while an image is displayed — image should be cleaned up
- [ ] Verify existing tests pass (`./gradlew :tamboui-core:test` — 527 tests passing)

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
@@ -30,6 +30,7 @@ public final class Frame {
     private Position cursorPosition;
     private boolean cursorVisible;
     private final Deque<String> contextKeyStack = new ArrayDeque<>();
+    private boolean hadRawOutput;
 
     Frame(Buffer buffer, OutputStream rawOutput) {
         this.buffer = buffer;
@@ -98,6 +99,7 @@ public final class Frame {
     public void renderWidget(Widget widget, Rect area) {
         if (widget instanceof RawOutputCapable) {
             ((RawOutputCapable) widget).render(area, buffer, rawOutput);
+            hadRawOutput = true;
         } else {
             widget.render(area, buffer);
         }
@@ -218,6 +220,13 @@ public final class Frame {
         if (!contextKeyStack.isEmpty()) {
             contextKeyStack.pop();
         }
+    }
+
+    /**
+     * Returns whether any {@link RawOutputCapable} widget was rendered in this frame.
+     */
+    boolean hadRawOutput() {
+        return hadRawOutput;
     }
 
     /**
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
@@ -6,7 +6,10 @@ package dev.tamboui.terminal;
 
 import java.io.OutputStream;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Deque;
+import java.util.List;
 import java.util.Optional;
 
 import dev.tamboui.buffer.Buffer;
@@ -31,6 +34,7 @@ public final class Frame {
     private boolean cursorVisible;
     private final Deque<String> contextKeyStack = new ArrayDeque<>();
     private boolean hadRawOutput;
+    private List<Rect> rawOutputAreas;
 
     Frame(Buffer buffer, OutputStream rawOutput) {
         this.buffer = buffer;
@@ -100,6 +104,10 @@ public final class Frame {
         if (widget instanceof RawOutputCapable) {
             ((RawOutputCapable) widget).render(area, buffer, rawOutput);
             hadRawOutput = true;
+            if (rawOutputAreas == null) {
+                rawOutputAreas = new ArrayList<>();
+            }
+            rawOutputAreas.add(area);
         } else {
             widget.render(area, buffer);
         }
@@ -227,6 +235,13 @@ public final class Frame {
      */
     boolean hadRawOutput() {
         return hadRawOutput;
+    }
+
+    /**
+     * Returns the areas where {@link RawOutputCapable} widgets rendered in this frame.
+     */
+    List<Rect> rawOutputAreas() {
+        return rawOutputAreas != null ? rawOutputAreas : Collections.emptyList();
     }
 
     /**
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
@@ -123,6 +123,13 @@ public final class Frame {
      */
     public <S> void renderStatefulWidget(StatefulWidget<S> widget, Rect area, S state) {
         widget.render(area, buffer, state);
+        if (widget instanceof RawOutputCapable) {
+            hadRawOutput = true;
+            if (rawOutputAreas == null) {
+                rawOutputAreas = new ArrayList<>();
+            }
+            rawOutputAreas.add(area);
+        }
     }
 
     /**
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
@@ -6,6 +6,7 @@ package dev.tamboui.terminal;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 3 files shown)


davsclaus (approved):

maxandersen:
Cool. I think similar happens with iterm2 protocol. 

@gnodet you opened the issues in draft mode - was that intended / more to come in them?

gnodet:
> Cool. I think similar happens with iterm2 protocol.
> 
> @gnodet you opened the issues in draft mode - was that intended / more to come in them?

Not really, but I can check iterm2 protocol too.

This is also related to https://github.com/apache/camel/pull/22856.

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
```diff
@@ -133,6 +142,22 @@ public CompletedFrame draw(Consumer<Frame> renderer) {
                 Frame frame = new Frame(currentBuffer, rawOutput);
                 renderer.accept(frame);
 
+                // Clean up images when raw-output widgets are no longer rendered.
+                // Kitty: delete-all removes the separate graphics layer.
+                // iTerm2/Sixel: overwrite previous image areas with spaces.
+                if (previousFrameHadRawOutput && !frame.hadRawOutput()) {
```
maxandersen: what if image has moved between two frames - then it would not be cleared would it?

tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
```diff
@@ -98,6 +103,11 @@ public int height() {
     public void renderWidget(Widget widget, Rect area) {
```
maxandersen, line 103: how about renderStatefulWidget? should it also have similar?

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
```diff
@@ -133,6 +142,22 @@ public CompletedFrame draw(Consumer<Frame> renderer) {
...
+                    rawOutput.write(KITTY_DELETE_ALL);
+                    for (Rect imageArea : previousRawOutputAreas) {
+                        for (int y = imageArea.y(); y < imageArea.y() + imageArea.height(); y++) {
+                            String move = String.format("\033[%d;%dH", y + 1, imageArea.x() + 1);
+                            rawOutput.write(move.getBytes(StandardCharsets.US_ASCII));
+                            byte[] spaces = new byte[imageArea.width()];
```
maxandersen: small nitpick but this spaces could be shared in this loop and only allocate if spaces.length!=imageArea.width and also avoid calling arrays.fill.

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
```diff
@@ -192,6 +220,11 @@ private void resize(Rect area) {
         currentBuffer = Buffer.empty(area);
         previousBuffer = Buffer.empty(area);
         try {
+            if (previousFrameHadRawOutput) {
```
maxandersen: should this not be extracted and also called in close() and clear() so its always deleted the same way?

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
```diff
@@ -192,6 +220,11 @@ private void resize(Rect area) {
...
+                rawOutput.write(KITTY_DELETE_ALL);
```
maxandersen: is kitty_delete_all relevant/same for all image protocols?

maxandersen (changes requested):
it looks like good direction but please check the comments and if possible add tests for these cases if at all possible to assert on it

tamboui-core/src/main/java/dev/tamboui/terminal/Frame.java
gnodet: Added — `renderStatefulWidget()` now checks `instanceof RawOutputCapable` after the stateful render and tracks the area. No current `StatefulWidget` implements `RawOutputCapable`, but this keeps the two render paths consistent for future use.

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
gnodet: Fixed — the spaces array is now allocated once to `maxWidth` across all stale areas, and written with `rawOutput.write(spaces, 0, imageArea.width())` to handle varying widths without reallocation.

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
gnodet: Done — extracted `cleanupRawOutput(List<Rect> currentAreas)` and it's now called from `draw()`, `resize()`, `clear()`, and `close()`.

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
gnodet: Good catch! Fixed — `cleanupRawOutput()` now computes stale areas by diffing previous vs. current raw output areas. When images move, only the old (no longer covered) areas get overwritten with spaces. Kitty delete-all is only sent when ALL images are fully removed (since it can't selectively delete without image IDs, and sending it would also remove the just-rendered current image).

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
gnodet: No — `KITTY_DELETE_ALL` is Kitty-specific. Non-Kitty terminals (iTerm2, Sixel, etc.) safely ignore this APC sequence since they don't understand it. The space overwriting handles cleanup for iTerm2/Sixel (their images occupy regular terminal cells). Added a clarifying comment on the constant.

gnodet:
Addressing the remaining review comments that I couldn't reply to inline (outdated diff hunks):

**Image movement between frames** (@maxandersen [comment](https://github.com/tamboui/tamboui/pull/317#discussion_r3171125605)):
`cleanupRawOutput()` now computes stale areas by diffing previous vs. current raw output areas. When images move, only the old (no longer covered) areas get overwritten with spaces. Kitty delete-all is only sent when ALL images are fully removed (since it can't selectively delete without image IDs, and sending it would also remove the just-rendered current image).

**Extract cleanup method for close()/clear()** (@maxandersen [comment](https://github.com/tamboui/tamboui/pull/317#discussion_r3171144148)):
Extracted `cleanupRawOutput(List<Rect> currentAreas)` — now called from `draw()`, `resize()`, `clear()`, and `close()`.

tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
gnodet: To clarify: we could conditionally send `KITTY_DELETE_ALL` only when the Kitty protocol is in use, but `Terminal` (in tamboui-core) doesn't know which protocol the `RawOutputCapable` widget used — that decision lives in the `Image` widget in tamboui-image. Making it protocol-aware would require threading protocol info back through the Frame or adding a cleanup callback, which adds complexity for very little gain. The sequence is 12 bytes that non-Kitty terminals silently ignore (well-formed APC they don't understand), so sending it unconditionally seems like the right tradeoff. Happy to change if you feel differently.

maxandersen (approved):
looks good.  lets confirm it completes ci run

PR #318, opened 2026-05-01
Fix unicode support when pasting

melix:
This is a followup to our conversation at Devoxx France. This commit reworks the event parser (and backend) to be based on codepoints instead of chars. I tried to keep the API backwards compatible as much as possible, so there is now a deprecated method `.character()`. Look at individual commits for details.


maxandersen:
This looks like the right approach (handling symbols rather than characters).

I see the first commit mention mesh does this natively (yay!) but line and Panama does not ...do I read it right in code that Panama now changed to do it right; but for jline you added manual workaround ?

tamboui-tui/src/main/java/dev/tamboui/tui/event/KeyEvent.java
```diff
@@ -244,11 +308,45 @@ public KeyModifiers modifiers() {
...
+     *
+     * @return a one- or two-{@code char} {@code String} representing the Unicode character,
+     *         or {@code ""} for non-{@link KeyCode#CHAR} events
+     */
+    public String string() {
+        return code == KeyCode.CHAR ? new String(Character.toChars(character)) : "";
+    }
+
+    /**
+     * Returns the Unicode code point for {@link KeyCode#CHAR} events, or {@code 0} otherwise.
```
maxandersen, line 344: what/where is the 0 otherwise referring to ? seems to just always return a character? or is it just hinting at character is not set if non-character?

maxandersen: what I mean I would have expected similar code as found in character() method.

gnodet:
Nice work, Cédric — this is a solid rework. The backwards-compatibility approach (deprecating `character()`, keeping `char` overloads that delegate) is well done, and the new convenience methods (`isChar()`, `string()`, `codePoint()`) clean up the call sites significantly.

A few items I'd like your thoughts on:

**Missing tests for `GraphemeClusters`**
This is the most complex new logic (ZWJ backward traversal, surrogate-aware `prevCpStart`), but it has no dedicated tests. Edge cases worth covering: cursor at position 0, lone surrogates, multi-ZWJ sequences (e.g. 👨‍👩‍👧‍👦), and text with no ZWJ at all. Would you be open to adding a `GraphemeClustersTest`?

**Missing tests for backend code-point decoding**
The new `readCodePoint()` methods in `UnixTerminal`, `JLineBackend`, and `WindowsTerminal` each have non-trivial decoding (UTF-8 multi-byte assembly, surrogate pair stitching, incomplete-sequence fallback). The EventParser tests assume the backend already delivers decoded code points, so the decoding layer itself is untested. Even a few unit tests for the happy path + malformed input would add confidence.

**`GraphemeClusters` only handles ZWJ sequences**
Combining diacritical marks (e.g. `e` + U+0301) and variation selectors are not handled — a backspace on `é` composed as two code points would delete only the accent. Totally fine as a first pass, but worth documenting as a known limitation (either in the class Javadoc or a follow-up issue).

**`findScrollColForCursor` is O(n²) for long lines**
The backward loop calls `CharWidth.of(line.substring(col, end))` per iteration, creating a String each time. Not a practical problem at typical terminal widths, but wanted to flag it in case you'd like to optimize while you're in the area.

**`TextInputState.insert(String)` — newline safety?**
Now that `insert()` receives `event.string()`, is there any concern about a `\n` sneaking into a single-line `TextInputState`? Might be worth a guard or at least a Javadoc note that callers are expected to handle newlines separately.

maxandersen (approved):

PR #319, opened 2026-05-01
feat: Bracketed paste support

melix:
This is built on top of #318 , so there's a single commit to review (see 7811b9c853d866c9609d49d2c7bb3171e716fdd7).


maxandersen:
the unicode stuff looks good - works for me :)

bracketed mode I'm not sure how to test. any tips on which terminals that does it?

also, when enabled - shouldn't text boxes etc. by default "just work" so they don't need to handle the explicit paste event unless they really want/need it ?

melix:
When not enabled (default), pasting "just works" in the sense that events are sent to the terminal codepoint by codepoint. If enabled, then paste events are generated so only elements which use paste events will handle it: at this stage the text field and text area elements.

maxandersen:
i'm ok on merging this but maybe:

a) add demo enabling this?
b) document it?

PR #326, opened 2026-05-03
Allow rendering form field inputs before their labels

ncovercash:
This adds a property `inputBeforeLabel` to `FormFieldElement` which will render a field's input before the label. This is helpful for a left-aligned form; my particular use case has a lot of checkboxes.

### Usage

This is entirely opt-in; **no breaking changes** are included here.

```java
// single field usage
formField("Overdrive", state).inputBeforeLabel(true);
// form usage
form(state)
  .field("foo", "Overdrive", FieldType.CHECKBOX)
  // can specify as many as applicable
  // works before or after fields are specified, too
  .inputBeforeLabelTypes(FieldType.CHECKBOX, FieldType.TOGGLE);
```

### Results

This allows a much more natural look:
```
[x] Option 1
[ ] Option 2 (not as cool)
[x] Option 3
...
```

compared to the more awkward (particularly with long labels)
```
Option 1 [x]
Option 2 (not as cool) [ ]
Option 3 [x]
...
```

### DIFF
--- a/docs/src/snippets/java/dev/tamboui/docs/snippets/FormsSnippets.java
+++ b/docs/src/snippets/java/dev/tamboui/docs/snippets/FormsSnippets.java
@@ -90,13 +90,16 @@ public class FormsSnippets {
 
     void booleanFields() {
         // tag::boolean-fields[]
-        BooleanFieldState subscribeState = new BooleanFieldState(false);
+        BooleanFieldState state = new BooleanFieldState(false);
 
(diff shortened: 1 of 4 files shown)


melix (approved):

maxandersen:
while it builds - shouldn't we actually default to checkbox/toggles to be on the left?

ncovercash:
Up to you — IMO no, since the other fields are all right aligned.

PR #327, opened 2026-05-04
Add new Scrollable element

ncovercash:
This adds a new `ScrollableElement` that functions similarly to a `Panel` or other container element with the added benefit of a built-in `Scrollbar`. This scrollbar will automatically adjust based on the children inside the container and automatically handles up/down keyboard events and scroll up/down mouse events.

This closes #289 and provides a super simple API for most scenarios where there are long lists or lines of text.

Includes:
- Automatic keyboard/mouse event handling;
- Plays nicely with fill-mode elements like `spacer()`;
- Works in DSL and immediate mode;
- Allows customization of the `Scrollbar`
- Works nicely with `VERTICAL_LEFT` and `VERTICAL_RIGHT` scrollbars (no horizontal support is included)
- Allows adding indicators for when content has gone off the screen (`scrollUpIndicator` and `scrollDownIndicator`)

Here's a demo with a long list of checkboxes:

https://github.com/user-attachments/assets/f48075c4-2bc5-444e-b4ab-5f84d7480738

Please let me know what else needs to be added to this PR; I added some notes in the documentation but couldn't find a more appropriate place than that one.

Also, are there any tests I can use as a good reference to write some for this code? This is mostly just a `Panel` with the majority of options removed, however, removing the options doesn't leave me much to test (and the same preferred size calculations, so it feels weird duplicating this test). I can be a bit more redundant but wanted to ask the best approach before I spend significant time on that.

### DIFF
--- a/docs/src/docs/asciidoc/api-levels.adoc
+++ b/docs/src/docs/asciidoc/api-levels.adoc
@@ -402,6 +402,15 @@ ListElement provides methods for navigation:
 include::{snippets-dir}/dev/tamboui/docs/snippets/ApiLevelsSnippets.java[tags=list-element-navigation]
 ----
 
+=== Scrolling
+
+For a convenient vertical scrolling container that works out of the box with keyboard and mouse controls, use `scrollable()`:
+
+[source,java]
+----
+include::{snippets-dir}/dev/tamboui/docs/snippets/ApiLevelsSnippets.java[tags=scrollable-usage]
+----
+
 === Event Handling
 
 Attach handlers to elements:
--- a/docs/src/snippets/java/dev/tamboui/docs/snippets/ApiLevelsSnippets.java
+++ b/docs/src/snippets/java/dev/tamboui/docs/snippets/ApiLevelsSnippets.java
@@ -588,6 +588,14 @@ public class ApiLevelsSnippets {
         // end::list-element-navigation[]
     }
 
+    void scrollableUsage() {
+        // tag::scrollable-usage[]
+        scrollable(children)
+            .scrollUpIndicator(text("[scroll up to see more...]").dim())
+            .scrollDownIndicator(text("[scroll down to see more...]").dim());
+        // end::scrollable-usage[]
+    }
+
     void eventHandling() {
         // tag::event-handling[]
         panel("Interactive")
--- a/docs/video/scrollable-demo.tape
+++ b/docs/video/scrollable-demo.tape
@@ -1,0 +1,14 @@
+Source shared_.tape
+
+# Setup
+Hide
+Type jbang scrollable-demo
+Enter
+Sleep 2
+Show
+
+# Recording
+Type "jjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjj"
+Screenshot screenshots/scrollable-demo.svg
+Type "kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkk"
+Sleep 2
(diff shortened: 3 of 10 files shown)


maxandersen:
@ncovercash looks like a good start.

The feature in #289 I saw more like a ViewPort rather than vertical only scrollable panel - any challenges in making that rather than a vertical only ScrollablePanel?

About widgets - would be be good to add a Demo similar to what exist for almost all other widgets.

ncovercash:
> The feature in #289 I saw more like a ViewPort rather than vertical only scrollable panel - any challenges in making that rather than a vertical only ScrollablePanel?

@maxandersen a dual-axis one would be quite a bit trickier as it would have to know multiple dimensions for each element, which gets very strange very fast if any elements have fill constraints. Horizontally scrolling is also a little weirder since this implementation simply doesn't render elements that don't fit on the page (great for rows of text and the like, plus I didn't see any APIs for rendering just "part" of an element (without doing some extremely contrived logic)), but that means anything except smaller columns probably wouldn't work super well. 

It looks like the example referenced in #289 (both [Flutter's version](https://api.flutter.dev/flutter/widgets/SingleChildScrollView-class.html) and the use case in the ticket) are based on vertical scrolling, anyways.

melix (approved):

maxandersen:
Nice work @ncovercash — this fills a real gap. Having a scrollable container that just works out of the box with keyboard and mouse is exactly the kind of thing that makes the toolkit layer useful. The demo with checkboxes is a great showcase.

A few things to address before merging:

**Naming:**
- `getState()` → `state()` — we avoid the `getXXX` prefix for simple accessors (see AGENTS.md code style guidelines)

**Documentation:**
- The `api-levels.adoc` snippet is a good start, but `ScrollableElement` deserves a section in `widgets.adoc` alongside List, Table, and Tabs — covering the full API (indicators, custom scrollbar, orientation support)
- The class javadoc should document CSS child selectors following the project convention (see `ListElement` or `TableElement` for the pattern)
- Would be good to mention the keyboard/mouse bindings in the docs (up/down, page up/down, scroll wheel)

**Tests:**
- To answer your question about test references: `ListElementTest` and `PanelElementTest` are good starting points. `BufferAssertions` is the main tool for verifying rendered output
- Key things worth testing: scroll position updates on key/mouse events, indicator visibility when at top/bottom/middle, content clipping when items overflow, and the edge case where all content fits (no scrolling needed)

**Minor:**
- The `ScrollableElement(Scrollbar, Element...)` constructor takes a `Scrollbar` widget directly, but the `Toolkit.scrollable(ScrollbarElement, ...)` factory method takes a `ScrollbarElement`. These seem mismatched — should the constructor accept `ScrollbarElement` for consistency with the toolkit layer?

Happy to help with any of these. The core implementation looks solid.

maxandersen:
Rebased on main (clean, no conflicts) and pushed a commit addressing the review feedback:

- **`getState()` → `state()`** per project naming conventions
- **Removed `scrollable(ScrollbarElement, ...)` factory** — it silently treated the scrollbar as a child element due to varargs resolution (the constructor takes `Scrollbar` widget, not `ScrollbarElement`)
- **Javadoc** — documented keyboard/mouse bindings on the class, cleaned up param docs
- **Java 8 compat** — replaced Stream with for-loop in `renderContent`
- **Demo fix** — changed `mouseCapture(false)` → `mouseCapture(true)` so scroll wheel actually works
- **13 tests** — rendering, key/mouse scroll events, indicator visibility, edge cases, preferredSize

Opened #387 for a follow-up: the demo's `formField` checkboxes aren't fully interactive because `ScrollableElement` consumes up/down keys for scrolling before children can use them for focus navigation. Not a blocker for this PR — the scrollable container itself works great.

maxandersen:
wanted to make the demo actually have full widgets in the scrollable area but focus handling is not great - it needs to become focus aware I think for that to work completely. opened #387 to try capture that and we can merge this in as still useful in its current state.

PR #330, opened 2026-05-04
Add new ButtonElement

ncovercash:
This adds a new `ButtonElement` element that wraps an element to add focus styles and built-in key event handling.

It features:
- Automatic focus handling (will apply `reversed` to the inner element by default, or a custom style if specified)
- Built-in handling of confirm keyboard events

---

Please let me know what else needs to be added to this PR; I added basic usage to the documentation (for programmatic submission, now the example actually works too).

### DIFF
--- a/docs/src/snippets/java/dev/tamboui/docs/snippets/FormsSnippets.java
+++ b/docs/src/snippets/java/dev/tamboui/docs/snippets/FormsSnippets.java
@@ -316,7 +316,8 @@ public class FormsSnippets {
         // Render form with submit button
         column(
             loginForm,
-            text(" Login ").bold()  // Note: button() is not yet available
+            button(text(" Login ").bold())
+                .onPress(() -> loginForm.submit())
         );
         // end::programmatic-submit[]
     }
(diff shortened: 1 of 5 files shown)


maxandersen:
Having button definitely makes sense.

Not sure i'm following the idea or renaming GlobalEventHandler to EventHandler and use that directly onPress.

@melix wdyt?

ncovercash:
> Not sure i'm following the idea or renaming GlobalEventHandler to EventHandler and use that directly onPress.

Figured it made more sense than adding a separate class with the exact same API, but I'll gladly create a new one if y'all want

melix (changes requested):
The Javadocs for EventHandler are now wrong. I think we could keep backwards compatibility by having `GlobalEventHandler` implement `EventHandler`.

ncovercash:
> The Javadocs for EventHandler are now wrong. I think we could keep backwards compatibility by having `GlobalEventHandler` implement `EventHandler`.

That makes sense; I'll implement that change Monday

maxandersen:
> > The Javadocs for EventHandler are now wrong. I think we could keep backwards compatibility by having `GlobalEventHandler` implement `EventHandler`.
> 
> That makes sense; I'll implement that change Monday

were you able to make it?

PR #331, opened 2026-05-04
Expose lastFlatEntries from TreeElement

ncovercash:
`lastFlatEntries` is already exposed on `TreeWidget`; the same should be exposed on `TreeElement`. Without it, `selected()` doesn't really provide anything of use (without requiring the external consumer to do its own traversals).


melix (approved):

melix:
@all-contributors add @ncovercash for code

PR #342, opened 2026-05-13
feat: Add alignment to table Cell widget

### LINKED ISSUE #320: Set text-align on table columns
Is it possible to set `text-align` on table `Cell`s, or more broadly on a table column?

I see `alignment(Alignment)` methods on Text, Paragraph, etc. but not on `Cell`.

heuermh:
Fixes #320 

Re: Acceptable Use of LLMs

Claude Code/Opus 4.7 was used to bootstrap this pull request, then code was edited by me for style and to improve docs.

Prompt
```
❯ Start by considering AGENTS.md. Then note text-align support in Paragraph, class
dev.tamboui.widgets.paragraph.Paragraph, implemented via alignment(Alignment)
method. Implement similar functionality in Cell, class dev.tamboui.widgets.table.Cell.
```

### DIFF
--- a/tamboui-widgets/src/main/java/dev/tamboui/widgets/table/Cell.java
+++ b/tamboui-widgets/src/main/java/dev/tamboui/widgets/table/Cell.java
@@ -4,6 +4,9 @@
  */
 package dev.tamboui.widgets.table;
 
+import java.util.Objects;
+
+import dev.tamboui.layout.Alignment;
 import dev.tamboui.style.Style;
 import dev.tamboui.text.Line;
 import dev.tamboui.text.Span;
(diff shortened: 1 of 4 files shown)


melix:
Thanks! I may be missing something but I think it misses support for CSS, so that you can specify the alignment in a CSS file for example.

heuermh:
`Cell` doesn't support CSS style at all; there is no `Builder` with a `styleResolver`.

`Table` in the same package does, but only for `background` and `highlightColor`. There is also `rowStyleResolver(BiFunction<Integer, Integer, Style> resolver)`, would it be worth building out similar functionality for column-wise style?

In any case, I might suggest these be tackled in future pull request(s). Please let me know what you think.

heuermh:
Rebased and force-pushed; could this please be reviewed and merged as-is?

maxandersen:
feature is good but I think we should do this right - table and cells were done before CSS was present so I think we better do this properly and wait after 0.4

PR #345, opened 2026-05-16
Fix #343: replace C0 control chars in Buffer.setString to prevent cursor corruption

### LINKED ISSUE #343: TUI with panels can show garbage when switching if raw printing content that has tab chars
Found by the AI when researching a weird problem in Camel TUI where we have a tab to show logs. And the logs have \t tabs such as when you have stacktraces.

Image
Image

Here is what claude told me

⏺ Found the root cause in the bytecode. The backend skips cursor repositioning for consecutive cells — it relies on the terminal cursor advancing by exactly 1. But stack trace lines contain
  \t (tab) characters, and CharWidth.of('\t') returns 0 in TamboUI, so the tab gets merged into the previous cell's symbol. When the backend writes that cell, the terminal jumps the cursor
  to the next tab stop (column 8, 16, etc.) instead of 1. All subsequent consecutive cells land at wrong positions — and since TamboUI's buffer model doesn't know this happened, the diff
  never corrects those positions. The corruption persists across tab switches.

  The fix: strip \t from log lines in stripAnsi.

davsclaus:
## Problem

Closes #343.

C0 control characters (U+0000–U+001F) and DEL (U+007F) written into a `Buffer` via `setString()` are stored verbatim as `Cell` symbols. `CharWidth.of('\t')` returns 1, so `Buffer.setString()` stores `"\t"` as a 1-wide cell. In `AbstractBackend.draw()`, after writing that cell, `cursorX = x + 1` is assumed — but the terminal actually moves the cursor to the **next tab stop** (column 8, 16, …), not `x+1`.

The adjacency optimisation then skips emitting `\e[y;xH` for the next cell because `cursorX` appears to already match. That cell lands at the tab-stop column instead of `x+1`, corrupting the rest of the row. The corruption persists across frames because `diff()` only catches changed cells — not positions that were misrendered due to cursor state drift.

The same issue applies to `\r` (resets to column 0), `\n` (moves to next line), and all other C0 control characters.

## Fix

Sanitise in `Buffer.setString()` by replacing any C0/DEL code point with a space before it reaches the cell model. This is the correct abstraction boundary: the buffer represents displayable terminal state, and control characters have no valid display representation there. Fixing here also protects `toAnsiString()`, `diff()`, and all other buffer consumers.

```java
// Before
String symbol = codePoint < 128 ? ASCII_STRINGS[codePoint] : new String(Character.toChars(codePoint));

// After
String symbol = (codePoint < 0x20 || codePoint == 0x7F)
        ? " "
        : codePoint < 128 ? ASCII_STRINGS[codePoint] : new String(Character.toChars(codePoint));
```

## Test

Added `setStringControlCharsReplacedWithSpace()` to `BufferTest` covering `\t`, `\r`, `\n`, NUL+1, and DEL.

---
_Discovered while using TamboUI in [Apache Camel JBang TUI](https://github.com/apache/camel). Stack traces in log output contain tab-indented lines which triggered this bug, causing garbage rendering on all other TUI tabs._

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/buffer/Buffer.java
+++ b/tamboui-core/src/main/java/dev/tamboui/buffer/Buffer.java
@@ -290,7 +290,13 @@ public final class Buffer {
             }
             appendToLast = false;
 
-            String symbol = codePoint < 128 ? ASCII_STRINGS[codePoint] : new String(Character.toChars(codePoint));
+            // C0 control chars (U+0000–U+001F) and DEL (U+007F) are not displayable glyphs.
+            // Writing them to the terminal moves the cursor unpredictably (e.g. \t jumps to
+            // the next tab stop, \r resets to column 0), which breaks the cursor-adjacency
+            // optimisation in AbstractBackend and corrupts all subsequent cells in the row.
+            String symbol = (codePoint < 0x20 || codePoint == 0x7F)
+                    ? " "
+                    : codePoint < 128 ? ASCII_STRINGS[codePoint] : new String(Character.toChars(codePoint));
 
             if (charWidth == 2 && col + 1 >= area.right()) {
                 // Wide char at rightmost column: no room for continuation, replace with space
--- a/tamboui-core/src/test/java/dev/tamboui/buffer/BufferTest.java
+++ b/tamboui-core/src/test/java/dev/tamboui/buffer/BufferTest.java
@@ -64,6 +64,27 @@ class BufferTest {
     }
 
     @Test
+    @DisplayName("Buffer setString replaces C0 control characters with space")
+    void setStringControlCharsReplacedWithSpace() {
+        Buffer buffer = Buffer.empty(new Rect(0, 0, 10, 1));
+        // Tab (\t) is the most common culprit: terminals jump to the next tab stop
+        // rather than advancing one column, breaking cursor-adjacency tracking.
+        buffer.setString(0, 0, "a\tb", Style.EMPTY);
+
+        assertThat(buffer.get(0, 0).symbol()).isEqualTo("a");
+        assertThat(buffer.get(1, 0).symbol()).isEqualTo(" "); // \t replaced with space
+        assertThat(buffer.get(2, 0).symbol()).isEqualTo("b");
+
+        // Other C0 control chars and DEL are also replaced with space
+        Buffer buffer2 = Buffer.empty(new Rect(0, 0, 10, 1));
+        buffer2.setString(0, 0, "\r\n\u0001\u007f", Style.EMPTY);
+        assertThat(buffer2.get(0, 0).symbol()).isEqualTo(" "); // \r replaced
+        assertThat(buffer2.get(1, 0).symbol()).isEqualTo(" "); // \n replaced
+        assertThat(buffer2.get(2, 0).symbol()).isEqualTo(" "); // NUL+1 replaced
+        assertThat(buffer2.get(3, 0).symbol()).isEqualTo(" "); // DEL replaced
+    }
+
+    @Test
     @DisplayName("Buffer setString with style")
     void setStringWithStyle() {
         Buffer buffer = Buffer.empty(new Rect(0, 0, 10, 1));


davsclaus:
Here are some small animations with the bug visible and then after with this PR that fixes it
not-fixed

davsclaus:
and here is fixed
fixed

maxandersen:
i'm trying to grok what is a reproducer for this? any output to buffer with a \t in it ?

davsclaus:
> i'm trying to grok what is a reproducer for this? any output to buffer with a \t in it ?

any kind of java application that throws an exception with stacktraces that is logged

davsclaus:
Something like
```
2026-05-18 07:22:00.309  INFO 65988 --- [r-cron_Worker-1] org.quartz.core.JobRunShell              : Job Camel_cron.trigger threw a JobExecutionException:
org.quartz.JobExecutionException: groovy.lang.MissingPropertyException: No such property: bindArr for class: Script_f6748a166fdafb3ae098cda0cdb0fef6
Possible solutions: binding [See nested exception: groovy.lang.MissingPropertyException: No such property: bindArr for class: Script_f6748a166fdafb3ae098cda0cdb0fef6
Possible solutions: binding]
	at org.apache.camel.component.quartz.CamelJob.execute(CamelJob.java:89)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:203)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:566)
Caused by: groovy.lang.MissingPropertyException: No such property: bindArr for class: Script_f6748a166fdafb3ae098cda0cdb0fef6
Possible solutions: binding
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:69)
	at org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.unwrap(IndyGuardsFiltersAndSignatures.java:163)
	at Script_f6748a166fdafb3ae098cda0cdb0fef6.run(Script_f6748a166fdafb3ae098cda0cdb0fef6.groovy:6)
	at org.apache.camel.language.groovy.GroovyExpression.evaluate(GroovyExpression.java:62)
	at org.apache.camel.processor.SetBodyProcessor.process(SetBodyProcessor.java:45)
	at org.apache.camel.processor.errorhandler.RedeliveryErrorHandler$SimpleTask.handleFirst(RedeliveryErrorHandler.java:445)
	at org.apache.camel.processor.errorhandler.RedeliveryErrorHandler$SimpleTask.run(RedeliveryErrorHandler.java:421)
	at org.apache.camel.impl.engine.DefaultReactiveExecutor$Worker.doRun(DefaultReactiveExecutor.java:202)
	at org.apache.camel.impl.engine.DefaultReactiveExecutor$Worker.executeReactiveWork(DefaultReactiveExecutor.java:192)
	at org.apache.camel.impl.engine.DefaultReactiveExecutor$Worker.tryExecuteReactiveWork(DefaultReactiveExecutor.java:169)
	at org.apache.camel.impl.engine.DefaultReactiveExecutor$Worker.schedule(DefaultReactiveExecutor.java:143)
	at org.apache.camel.impl.engine.DefaultReactiveExecutor.scheduleMain(DefaultReactiveExecutor.java:59)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:162)
	at org.apache.camel.impl.engine.CamelInternalProcessor.processNonTransacted(CamelInternalProcessor.java:385)
	at org.apache.camel.impl.engine.CamelInternalProcessor.process(CamelInternalProcessor.java:361)
	at org.apache.camel.impl.engine.DefaultAsyncProcessorAwaitManager.process(DefaultAsyncProcessorAwaitManager.java:81)
	at org.apache.camel.support.AsyncProcessorSupport.process(AsyncProcessorSupport.java:32)
	at org.apache.camel.component.quartz.CamelJob.execute(CamelJob.java:78)
	... 2 more
```

davsclaus:
where all the stacktrace lines are prefixed with a tab char in the dump which happens by default in logging libraries and the JDK itself

davsclaus:
The unit test in `BufferTest.setStringControlCharsReplacedWithSpace()` is the minimal code-level reproducer:

```java
Buffer buffer = Buffer.empty(new Rect(0, 0, 10, 1));
buffer.setString(0, 0, "a\tb", Style.EMPTY);

assertThat(buffer.get(1, 0).symbol()).isEqualTo(" "); // \t replaced with space
```

So yes — any string passed to `setString()` that contains a `\t` (or `\r`, `\n`, NUL, DEL, or any C0 code point) will trigger it. In practice the most common source is Java stack traces, where every `\tat org.example...` line is prefixed with a literal tab by the JDK's `Throwable.printStackTrace()`.

maxandersen (approved):

maxandersen:
Confirmed the bug visually — it manifests when resizing the terminal small enough that the diff's cursor-adjacency optimization kicks in on rows with tab characters. On the first full draw every cell gets an explicit cursor reposition so it looks fine, but on subsequent partial redraws the hidden `\t` in the cell symbol jumps the cursor to the next tab stop, corrupting all adjacent cells.

For reference, here's how other TUI frameworks handle this:

| Framework | Approach |
|-----------|----------|
| **Ratatui** (Rust) | Filters out any grapheme containing a control char entirely — `.filter(\|symbol\| !symbol.contains(char::is_control))` |
| **tcell** (Go) | No sanitization at cell level, but repositions cursor via `TGoto` for every dirty cell so corruption is less likely |
| **Rich/Textual** (Python) | Two layers: `strip_control_codes()` removes bell/backspace/vertical tab/form feed/CR on Text creation, then `expand_tabs()` converts `\t` to column-aligned spaces during rendering |

Your approach of replacing with space at the buffer level is the right safety net. Opened #369 for a follow-up to also handle tabs more intelligently at the widget level (column-aligned expansion like Rich does).

Fix looks good, merging.

PR #346, opened 2026-05-17
feat: add DualSparkline widget with multi-row Sparkline improvements

davsclaus:
## Summary

Adds `MirroredSparkline` — a dual-series chart widget where bars grow upward (top series) and downward (bottom series) from a shared centre axis. Inspired by macOS Activity Monitor's network/disk graphs and the existing `Sparkline` widget.

- Two data series (top/bottom) rendered symmetrically from a centre separator row (`─`)
- Sub-pixel resolution via Unicode block characters (`▁▂▃▄▅▆▇█`) using `Sparkline.BarSet`
- Optional Y-axis labels (max / 0 / max) on the left
- Optional X-axis label row below the chart body
- Explicit `.max()` or auto-scaling across both datasets
- `RenderDirection.LEFT_TO_RIGHT` (default) and `RIGHT_TO_LEFT` support with mirrored x-axis labels
- `from(long[], long[])` and `from(List<Long>, List<Long>)` factory methods (matching `Sparkline` API)
- CSS feature parity: `top-color` and `bottom-color` properties via `StylePropertyResolver`, with programmatic `topForeground(Color)` / `bottomForeground(Color)` shortcuts that take precedence
- `CharWidth` used for x-axis label width calculations
- Documentation entry in `widgets.adoc` with code snippet
- 30 unit tests covering: basic rendering, scaling, truncation, bar sets, y-axis, x-axis, block border, empty/zero data, direction, factory methods, and CSS resolution

## Test plan

- [x] All 30 `MirroredSparklineTest` tests pass
- [x] Docs snippet compiles (`./gradlew :tamboui-docs:compileSnippets`)
- [x] Javadoc clean (`./gradlew :tamboui-widgets:javadoc`)

_Claude Code on behalf of Claus Ibsen_

### DIFF
--- a/tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
+++ b/tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
@@ -1,0 +1,604 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.widgets.sparkline;
+
+import java.util.List;
+
+import dev.tamboui.buffer.Buffer;
+import dev.tamboui.layout.Rect;
+import dev.tamboui.style.Color;
+import dev.tamboui.style.ColorConverter;
+import dev.tamboui.style.PropertyDefinition;
+import dev.tamboui.style.Style;
+import dev.tamboui.style.StylePropertyResolver;
+import dev.tamboui.text.CharWidth;
+import dev.tamboui.widget.Widget;
+import dev.tamboui.widgets.block.Block;
+
+/**
+ * A dual sparkline widget that displays two time-series datasets as vertical bars growing in opposite directions
+ * from a shared centre axis.
+ * <p>
+ * The top series renders as bars growing <em>upward</em> from the centre; the bottom series renders as bars growing
+ * <em>downward</em> from the centre. Sub-pixel resolution is achieved using Unicode block characters (▁▂▃▄▅▆▇█), giving
+ * smooth visual gradation within a single character row. This layout matches the style of macOS Activity Monitor's
+ * network and disk activity graphs.
+ * <p>
+ * Example usage:
+ *
+ * <pre>{@code
+ * DualSparkline chart = DualSparkline.builder()
+ *         .topData(inRates)
+ *         .bottomData(outRates)
+ *         .topStyle(Style.EMPTY.fg(Color.GREEN))
+ *         .bottomStyle(Style.EMPTY.fg(Color.BLUE))
+ *         .xLabels("-60s", "-45s", "-30s", "-15s", "now")
+ *         .block(Block.builder().borderType(BorderType.ROUNDED)
+ *                 .title(Title.from("In / Out  msg/s")).build())
+ *         .build();
+ * }</pre>
+ *
+ * <h2>Differences from {@link Sparkline}</h2>
+ * <table>
+ * <caption>Feature comparison between Sparkline and DualSparkline</caption>
+ * <tr>
+ * <th></th>
+ * <th>{@code Sparkline}</th>
+ * <th>{@code DualSparkline}</th>
+ * </tr>
+ * <tr>
+ * <td>Series</td>
+ * <td>1</td>
+ * <td>2 (top + bottom)</td>
+ * </tr>
+ * <tr>
+ * <td>Growth direction</td>
+ * <td>always upward from bottom row</td>
+ * <td>top grows up, bottom grows down, from a shared centre separator row</td>
+ * </tr>
+ * <tr>
+ * <td>Height</td>
+ * <td>1 row (fixed at bottom of area)</td>
+ * <td>fills the full area height</td>
+ * </tr>
+ * <tr>
+ * <td>Y-axis labels</td>
+ * <td>optional: max at the bar row</td>
+ * <td>optional: max / 0 / max at top, centre, and bottom rows</td>
+ * </tr>
+ * <tr>
+ * <td>X-axis labels</td>
+ * <td>optional label row below the bar row</td>
+ * <td>optional label row rendered below the chart body</td>
+ * </tr>
+ * <tr>
+ * <td>BarSet</td>
+ * <td>yes ({@link Sparkline.BarSet})</td>
+ * <td>yes (reuses {@link Sparkline.BarSet})</td>
+ * </tr>
+ * </table>
+ */
+public final class DualSparkline implements Widget {
+
+    /**
+     * CSS property for the top series bar colour ({@code top-color}).
+     * <p>
+     * Resolved by {@link Builder#styleResolver(StylePropertyResolver)}; a programmatic
+     * value set via {@link Builder#topForeground(Color)} takes precedence.
+     */
+    public static final PropertyDefinition<Color> TOP_COLOR =
+            PropertyDefinition.of("top-color", ColorConverter.INSTANCE);
+
+    /**
+     * CSS property for the bottom series bar colour ({@code bottom-color}).
+     * <p>
+     * Resolved by {@link Builder#styleResolver(StylePropertyResolver)}; a programmatic
+     * value set via {@link Builder#bottomForeground(Color)} takes precedence.
+     */
+    public static final PropertyDefinition<Color> BOTTOM_COLOR =
+            PropertyDefinition.of("bottom-color", ColorConverter.INSTANCE);
+
+    private static final int Y_LABEL_WIDTH = 4;
+    private static final Style DIM = Style.EMPTY.dim();
+    private static final String CENTRE_SEPARATOR = "─";
+
+    private final long[] topData;
+    private final long[] bottomData;
+    private final Style topStyle;
+    private final Style bottomStyle;
+    private final Long max;
+    private final Block block;
+    private final Sparkline.BarSet barSet;
+    private final Sparkline.BarSet reversedBarSet;
+    private final Sparkline.RenderDirection direction;
+    private final boolean showYAxis;
+    private final String[] xLabels;
+
+    private DualSparkline(Builder builder) {
+        this.topData = builder.topData;
+        this.bottomData = builder.bottomData;
+        this.max = builder.max;
+        this.block = builder.block;
+        this.barSet = builder.barSet;
+        this.reversedBarSet = builder.barSet.reversed();
+        this.direction = builder.direction;
+        this.showYAxis = builder.showYAxis;
+        this.xLabels = builder.xLabels;
+
+        Color resolvedTopFg = builder.resolveTopColor();
+        Style baseTopStyle = builder.topStyle;
+        this.topStyle = resolvedTopFg != null ? baseTopStyle.fg(resolvedTopFg) : baseTopStyle;
+
+        Color resolvedBottomFg = builder.resolveBottomColor();
+        Style baseBottomStyle = builder.bottomStyle;
+        this.bottomStyle = resolvedBottomFg != null ? baseBottomStyle.fg(resolvedBottomFg) : baseBottomStyle;
+    }
+
+    /**
+     * Creates a new builder.
+     *
+     * @return a new Builder
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Creates a dual sparkline with the given top and bottom data using default settings.
+     *
+     * @param topData    the top series data (bars grow upward from centre)
+     * @param bottomData the bottom series data (bars grow downward from centre)
+     * @return a new DualSparkline
+     */
+    public static DualSparkline from(long[] topData, long[] bottomData) {
+        return builder().topData(topData).bottomData(bottomData).build();
+    }
+
+    /**
+     * Creates a dual sparkline with the given top and bottom data using default settings.
+     *
+     * @param topData    the top series data (bars grow upward from centre)
+     * @param bottomData the bottom series data (bars grow downward from centre)
+     * @return a new DualSparkline
+     */
+    public static DualSparkline from(List<Long> topData, List<Long> bottomData) {
+        return builder().topData(topData).bottomData(bottomData).build();
+    }
+
+    @Override
+    public void render(Rect area, Buffer buffer) {
+        if (area.isEmpty()) {
+            return;
+        }
+
+        Rect inner = area;
+        if (block != null) {
+            block.render(area, buffer);
+            inner = block.inner(area);
+        }
+
+        if (inner.isEmpty()) {
+            return;
+        }
+
+        int innerH = inner.height();
+        int innerW = inner.width();
+
+        boolean hasXAxis = xLabels != null && xLabels.length > 0;
+        // Reserve one row at the bottom for x-axis labels when configured
+        int chartBodyRows = hasXAxis ? Math.max(1, innerH - 1) : innerH;
+        // Layout: top half, optional centre separator, bottom half.
+        // At height >= 3 the centre separator is shown; at height <= 2 it is
+        // dropped so both series still get a data row.
+        //   5 rows → top=2, centre=1, bottom=2
+        //   4 rows → top=1, centre=1, bottom=1, spare=1 (empty)
+        //   3 rows → top=1, centre=1, bottom=1
+        //   2 rows → top=1, bottom=1 (no centre separator)
+        //   1 row  → top=1 only
+        boolean hasCentre = chartBodyRows >= 3;
+        int topHalfH;
+        int bottomHalfH;
+        int centerRow;
+        if (hasCentre) {
+            // Both halves get equal rows; any spare row stays empty
+            int dataRows = chartBodyRows - 1; // exclude centre
+            topHalfH = dataRows / 2;
+            bottomHalfH = dataRows / 2;
+            centerRow = topHalfH;
+        } else if (chartBodyRows >= 2) {
+            // No centre separator — both series get 1 row
+            topHalfH = 1;
+            bottomHalfH = 1;
+            centerRow = -1;
+        } else {
+            // Single row — top only
+            topHalfH = 1;
+            bottomHalfH = 0;
+            centerRow = -1;
+        }
+
+        int yLabelW = showYAxis ? Y_LABEL_WIDTH : 0;
+        int chartW = Math.max(1, innerW - yLabelW);
+
+        int dataLen = Math.max(topData.length, bottomData.length);
+        int ticks = Math.min(dataLen, chartW);
+
+        long effectiveMax = computeMax();
+
+        // --- Bar rows ---
+        for (int r = 0; r < chartBodyRows; r++) {
+            int y = inner.y() + r;
+
+            if (showYAxis) {
+                String maxLabel = effectiveMax > 9999 ? "999+" : String.format("%4d", effectiveMax);
+                String label;
+                if (r == 0) {
+                    label = maxLabel;
+                } else if (hasCentre && r == centerRow) {
+                    label = "   0";
+                } else if (bottomHalfH > 0 && r == chartBodyRows - 1) {
+                    label = maxLabel;
+                } else {
+                    label = "    ";
+                }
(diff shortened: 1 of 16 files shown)


davsclaus:
I needed a graph that has dual data points to show IN vs OUT messages with Apache Camel. Tamboui sparkline is only 1 data set. So I added a mirrored variant that has 2 dataset and the mirror is drawn upside down. See this screenshot
Screenshot 2026-05-17 at 16 35 52

davsclaus:
During a deep comparison of `MirroredSparkline` against `Sparkline` two design questions emerged that maintainers may want to weigh in on. Neither blocks this PR — flagging for discussion.

---

### 1. `Sparkline.RenderDirection` coupling

To set direction on `MirroredSparkline`, callers must reference `Sparkline`:

```java
MirroredSparkline.builder()
    .direction(Sparkline.RenderDirection.RIGHT_TO_LEFT)
    .build();
```

A user who never uses `Sparkline` at all still has to import it. `Sparkline` itself writes `RenderDirection.RIGHT_TO_LEFT` (unqualified, it's a local type). A few options to reduce the coupling:

- Move `RenderDirection` and `BarSet` to a shared `SparklineSupport` or `dev.tamboui.widgets.sparkline` package-level class that both widgets reference.
- Add forwarding constants to `MirroredSparkline`: `public static final Sparkline.RenderDirection LEFT_TO_RIGHT = Sparkline.RenderDirection.LEFT_TO_RIGHT`.
- Leave as-is if the coupling is intentional (both widgets are in the same package and the shared type is a deliberate reuse signal).

---

### 2. Block rendering with empty data

`Sparkline.render()` exits before rendering the block when data is empty:

```java
if (area.isEmpty() || data.length == 0) {
    return;  // block NOT rendered — caller sees a blank area
}
```

`MirroredSparkline` renders the block even when both data arrays are empty (ticks = 0, so no bars appear inside):

```java
if (block != null) {
    block.render(area, buffer);  // border and title visible even with no data
}
// ticks = 0 → body loop skipped
```

The `MirroredSparkline` behaviour seems more useful for live dashboards — you still see the titled border while data is loading. But if consistency with `Sparkline` is preferred, the fix is a single early-return check. Worth deciding which is the intended contract before 1.0.

---

_Claude Code on behalf of Claus Ibsen_

davsclaus:
### Feature suggestion: bring Y-axis and X-axis labels to `Sparkline`

While implementing `MirroredSparkline` two axis features were added that don't exist on the standard `Sparkline` widget but would be equally useful there:

**Y-axis labels** — a 4-character label column on the left showing the data maximum (and optionally zero). Useful any time you want a quick scale reference without wrapping the chart in a `Block` with a title.

```
 100████▆▄
  75█▆▄▂▁▁
```

**X-axis labels** — a single label row below the chart body. Labels are distributed evenly across the data range; the rightmost label is right-aligned so it never overflows. Works with both `LEFT_TO_RIGHT` and `RIGHT_TO_LEFT` render directions.

```
█▆▄▂▁▁▂▄▆█
-60s      now
```

The implementation in `MirroredSparkline` is self-contained and straightforward to port:

- `xLabels(String... labels)` builder method — stores label array, reserves one row at the bottom of the area, distributes labels by fraction across the tick range using `CharWidth` for correct display-width calculations.
- `showYAxis(boolean show)` builder method — reserves a fixed 4-character column on the left, writes `%4d`-formatted max and `   0` at the appropriate rows.

Both features are additive (default off for `xLabels`, the existing `Sparkline` is single-row so `showYAxis` would need a height ≥ 2 to be useful) and have no impact on existing callers.

Happy to extract those into a follow-up PR against `Sparkline` if the maintainers think it fits the roadmap.

_Claude Code on behalf of Claus Ibsen_

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
```diff
@@ -0,0 +1,564 @@
...
+import dev.tamboui.style.Color;
+import dev.tamboui.style.ColorConverter;
+import dev.tamboui.style.PropertyDefinition;
+import dev.tamboui.style.Style;
+import dev.tamboui.style.StylePropertyResolver;
+import dev.tamboui.text.CharWidth;
+import dev.tamboui.widget.Widget;
+import dev.tamboui.widgets.block.Block;
+
+/**
```
maxandersen, line 20: is there a good reason why the mirroring isn't just a direction attribute? i.e. what warrants it to be separate widget type?

tamboui-widgets/src/test/java/dev/tamboui/widgets/sparkline/MirroredSparklineTest.java
```diff
@@ -0,0 +1,617 @@
...
+import dev.tamboui.layout.Rect;
+import dev.tamboui.style.Color;
+import dev.tamboui.style.Style;
+import dev.tamboui.style.TestStylePropertyResolver;
+import dev.tamboui.widgets.block.Block;
+
+import static dev.tamboui.assertj.BufferAssertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+class MirroredSparklineTest {
```
maxandersen: would be good with demo added for the widget

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
davsclaus: Good question. The core distinction is that `MirroredSparkline` manages **two independent data series** — top and bottom — whereas `Sparkline` is built around a single dataset. A direction attribute alone wouldn't capture that: you'd also need separate arrays, independent colour/scale control per series, and the centre separator row logic. The rendering is fundamentally different (bars grow both up *and* down from a shared midpoint rather than all upward from a bottom row), so a new type felt cleaner than overloading `Sparkline` with two optional data arrays and conditional rendering paths.

tamboui-widgets/src/test/java/dev/tamboui/widgets/sparkline/MirroredSparklineTest.java
davsclaus: Done — added `mirrored-sparkline-demo` under `tamboui-widgets/demos/mirrored-sparkline-demo/` and registered it in `jbang-catalog.json`. The demo shows two animated chart panels (Network IN/OUT in green/blue, Disk READ/WRITE in yellow/magenta) with Y-axis labels and X-axis timestamps, matching the style of the existing `sparkline-demo`.

davsclaus:
Screenshot 2026-05-18 at 07 43 45

davsclaus:
the position in the catalog is required to be accordingly
```
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -163,15 +163,15 @@
     "list-demo": {
       "script-ref": "tamboui-widgets/demos/list-demo/src/main/java/dev/tamboui/demo/ListDemo.java"
     },
+    "mirrored-sparkline-demo": {
+      "script-ref": "tamboui-widgets/demos/mirrored-sparkline-demo/src/main/java/dev/tamboui/demo/MirroredSparklineDemo.java"
+    },
     "paragraph-demo": {
       "script-ref": "tamboui-widgets/demos/paragraph-demo/src/main/java/dev/tamboui/demo/ParagraphDemo.java"
     },
     "scrollbar-demo": {
       "script-ref": "tamboui-widgets/demos/scrollbar-demo/src/main/java/dev/tamboui/demo/ScrollbarDemo.java"
     },
-    "mirrored-sparkline-demo": {
-      "script-ref": "tamboui-widgets/demos/mirrored-sparkline-demo/src/main/java/dev/tamboui/demo/MirroredSparklineDemo.java"
-    },
     "sparkline-demo": {
       "script-ref": "tamboui-widgets/demos/sparkline-demo/src/main/java/dev/tamboui/demo/SparklineDemo.java"
     },
```

davsclaus:
> > Task :tamboui-widgets:demos:mirrored-sparkline-demo:validateDemo FAILED

davsclaus:
LGTM would be good to get this into tamboui

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
maxandersen: So it's more like a dual sparkline/graph?

Is mirrored sparkling a common name for these ? Ie. They aren't mirrors, not symmetrical...so why named mirrored ? :)

maxandersen:
I'll give it a try - the more links to examples the better :)

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
davsclaus: Good point — the data sets aren't symmetric, so "mirrored" describes the **layout** rather than the data: one series grows upward and the other downward from a shared midpoint, mirroring each other's direction. This pattern is sometimes called a "mirrored bar chart" or "butterfly chart" in data-viz literature.

That said, `DualSparkline` would also work well and might be more intuitive. Happy to rename if you prefer — what do you think?

maxandersen: +1 on DualSparkline thats more telling.

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/MirroredSparkline.java
```diff
@@ -0,0 +1,564 @@
...
+    public static final class Builder {
+        private long[] topData = new long[0];
+        private long[] bottomData = new long[0];
+        private Style topStyle = Style.EMPTY;
+        private Style bottomStyle = Style.EMPTY;
+        private Long max;
+        private Block block;
+        private Sparkline.BarSet barSet = Sparkline.BarSet.NINE_LEVELS;
+        private Sparkline.RenderDirection direction = Sparkline.RenderDirection.LEFT_TO_RIGHT;
+        private boolean showYAxis = true;
```
maxandersen: normal sparkline doesn't have the YAxis - should we default this to false so the two are "aligned" ? if add YAxis to sparkline then default true would be ok.

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
```diff
@@ -0,0 +1,564 @@
...
+
+    @Override
+    public void render(Rect area, Buffer buffer) {
+        if (area.isEmpty()) {
+            return;
+        }
+
+        Rect inner = area;
+        if (block != null) {
+            block.render(area, buffer);
```
maxandersen, line 178: this is slightly different than how sparkline works - it does not render anything on empty data...I kinda like what happens here - the grid/axis printed so its "ready" for data.

could you align SparkLine so we get consistent behavior?

maxandersen:
| Feature suggestion: bring Y-axis and X-axis labels to Sparkline

yes, makes sense!

maxandersen (changes requested):
added some comments and realize matches your same findings. let me know if you can make them otherwise I'll make adjustmenets and merge to get a 0.4 out ;)

davsclaus:
> added some comments and realize matches your same findings. let me know if you can make them otherwise I'll make adjustmenets and merge to get a 0.4 out ;)

Yes you are very welcome to make the adjustments as needed. Looking forward to a release with this out of the box.
And I also think x and y axis on both the single and dual sparkline makes sense

maxandersen:
> Screenshot 2026-05-18 at 07 43 45

just realized the lower bar rendering is reverse on each line...fixing so it actually  is "flipped" correctly

maxandersen:
**Rebased, squashed, and extended the PR. Force-pushed 5 clean commits (was 9).**

### Commits

1. **`fix: render Sparkline block even with empty data`** — Block border/title stays visible while data loads
2. **`feat: add DualSparkline widget with demo`** — Original MirroredSparkline work squashed and renamed. Includes widget (568 lines), 630 lines of tests, demo app, docs snippet, VHS tape, JBang catalog entry. `showYAxis` defaults to `false`.
3. **`refactor: extract RenderDirection to top-level type`** — Shared enum for both Sparkline and DualSparkline. `Sparkline.RenderDirection` kept as `@Deprecated` with bridge method.
4. **`feat: add multi-row rendering and Y/X axis labels to Sparkline`** — Sparkline now fills full area height with sub-pixel resolution (was single-row). Added `showYAxis()` and `xLabels()` to both Sparkline and SparklineElement. Updated demos with compact/expanded layout and height controls.
5. **`feat: add DualSparklineElement to toolkit DSL`** — Toolkit wrapper with `topColor()`/`bottomColor()`, CSS `top-color`/`bottom-color` support via `styleResolver`. Factory methods: `dualSparkline()`, `dualSparkline(long[], long[])`, `dualSparkline(int[], int[])`. Updated AGENTS.md.

### Changes from original PR
- Renamed `MirroredSparkline` → `DualSparkline` (data isn't mirrored, layout is)
- `showYAxis` defaults to `false` (aligned with Sparkline defaults)
- Fixed bottom-half rendering to use upper block characters (`▔▀█`)
- Ported `showYAxis()`/`xLabels()` to Sparkline (was DualSparkline-only)
- Added multi-row rendering to Sparkline (was single-row)
- Added `DualSparklineElement` toolkit DSL element (new)
- Sparkline snippet updated to showcase new features

### Not in scope (filed as #380)
- Gradient/heatmap per-bar coloring (`colorMapper`, `ColorGradient`)
- Extracting `TFxColorSpace.lerp()` to `tamboui-core`
- CSS gradient properties

maxandersen:
Force-pushed — removed the top-level `RenderDirection` enum. Kept as `Sparkline.RenderDirection` (inner enum), consistent with `Sparkline.BarSet`. Both are sparkline-specific concepts, not general-purpose — no reason to promote them to package level. DualSparkline references it as `Sparkline.RenderDirection`.

**4 commits (down from original 9):**

1. **`fix: render Sparkline block even with empty data`** — Block border/title stays visible while data loads
2. **`feat: add DualSparkline widget with demo`** — Original MirroredSparkline squashed and renamed. Widget (568 lines), 630 lines of tests, demo app, docs snippet, VHS tape, JBang catalog entry. `showYAxis` defaults to `false`.
3. **`feat: add multi-row rendering and Y/X axis labels to Sparkline`** — Sparkline now fills full area height with sub-pixel resolution (was single-row). Added `showYAxis()` and `xLabels()` to both Sparkline and SparklineElement. Updated demos with compact/expanded layout and height controls.
4. **`feat: add DualSparklineElement to toolkit DSL`** — Toolkit wrapper with `topColor()`/`bottomColor()`, CSS `top-color`/`bottom-color` support via `styleResolver`. Factory methods in `Toolkit`: `dualSparkline()`, `dualSparkline(long[], long[])`, `dualSparkline(int[], int[])`. Updated AGENTS.md.

### Changes from original PR
- Renamed `MirroredSparkline` → `DualSparkline` (data isn't mirrored, layout is)
- `showYAxis` defaults to `false` (aligned with Sparkline defaults)
- Fixed bottom-half rendering to use upper block characters (`▔▀█`)
- Ported `showYAxis()`/`xLabels()` to Sparkline (was DualSparkline-only)
- Added multi-row rendering to Sparkline (was single-row)
- Added `DualSparklineElement` toolkit DSL element (new)
- Sparkline snippet updated to showcase new features
- Kept `RenderDirection` as `Sparkline.RenderDirection` inner enum — it's sparkline-specific, same as `BarSet`

### Not in scope (filed as #380)
- Gradient/heatmap per-bar coloring (`colorMapper`, `ColorGradient`)
- Extracting `TFxColorSpace.lerp()` to `tamboui-core`
- CSS gradient properties

maxandersen (approved):

maxandersen:
@melix @davsclaus I've done the proposed rename to DualSparkline + fixes to sparkline apis I found while adjusting on dualsparkline so they are more aligned in behavior (no axis/labels by default, ability to render higher than just 1-2 characters, update demos to illustrate the new features,e tc)

would be great to have a set of extra eyeballs before merging.

tamboui-widgets/src/main/java/dev/tamboui/widgets/sparkline/DualSparkline.java
```diff
@@ -0,0 +1,568 @@
...
+            PropertyDefinition.of("top-color", ColorConverter.INSTANCE);
+
+    /**
+     * CSS property for the bottom series bar colour ({@code bottom-color}).
+     * <p>
+     * Resolved by {@link Builder#styleResolver(StylePropertyResolver)}; a programmatic
+     * value set via {@link Builder#bottomForeground(Color)} takes precedence.
+     */
+    public static final PropertyDefinition<Color> BOTTOM_COLOR =
+            PropertyDefinition.of("bottom-color", ColorConverter.INSTANCE);
```
davsclaus, line 101: Minor: with an even `innerH` (e.g. 4), `halfH = (4-1)/2 = 1` gives the top series 1 row and the bottom series 2 rows. If this asymmetry is intentional, a brief javadoc note on `DualSparkline` or this calculation would help future readers understand the choice.

davsclaus (commented):
## Review Summary

Nice work — the `DualSparkline` widget is well-structured, the test coverage is thorough (30 tests), and the Sparkline multi-row / axis-label enhancements are a solid improvement. A few observations below.

**Note:** This review covers project conventions and scope. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

### Findings

**1. (Medium) PR title/body say "MirroredSparkline" but code uses "DualSparkline"**

The title is `feat: add MirroredSparkline widget` and the body opens with "Adds MirroredSparkline", but every file in the diff uses `DualSparkline`. The title and body should be updated to match the actual class name, and ideally mention the Sparkline enhancements (multi-row rendering, Y-axis labels, X-axis labels) since those are a meaningful part of this PR too.

**2. (Low) Asymmetric height allocation in DualSparkline for even heights**

With an even `innerH` (e.g. 4 rows), `halfH = (4-1)/2 = 1` gives the top series 1 row and the bottom series 2 rows. If this is by design, a brief javadoc note would help future readers.

**3. (Low) Inconsistent test assertion style in SparklineTest**

New tests use `BufferAssertions.assertThat(buffer).hasContent(...)` while existing tests use `assertThat(buffer.get(...).symbol()).isEqualTo(...)`. Minor style inconsistency within the same file — not blocking.

---

maxandersen:
Addressed review feedback:

**1. PR title** — Updated to "feat: add DualSparkline widget with multi-row Sparkline improvements". No more MirroredSparkline references.

**2. Asymmetric height allocation** — Fixed. Was a bug, not a design choice. With even `chartBodyRows` (e.g. 4), both halves were scaled to `halfH=1` but the bottom got 2 rows of space — the extra row was always empty. Now `topHalfH` and `bottomHalfH` are computed independently, extra row goes to the top half. Added inline comment explaining the split.

```
4 rows → top=2, centre=1, bottom=1
5 rows → top=2, centre=1, bottom=2
```

**3. Test assertion style** — Acknowledged, low priority. Existing tests use raw `assertThat`, new ones use `BufferAssertions`. Not mixing within individual tests, just across old vs new tests in the same file. Can unify in a follow-up.

Force-pushed, still 4 commits.

maxandersen:
### DualSparkline graceful height degradation

The widget adapts automatically as height shrinks — no separator at ≤2 inner rows, both halves always get equal space.

**2 rows — ultra compact, no border, no separator** (overview dashboards):



**4 rows — with border, inner=2, separator dropped**:



**7 rows — separator visible**:



**9 rows — with Y-axis labels**:



**11 rows — full featured with Y-axis + X-axis labels**:

maxandersen (approved):
LGTM - I think this makes for a super useful component and sparkline in general better off. The automatic fallback to compact rendering so you can in 2 character blocks height show in/out throughput info is 👨🏻‍🍳😘 

Ultimately DualSparkline is actually a superset of Sparkline this way so could possibly make even more code shared and just provide two different exposed API's so content assist is limited to what you need...for now kept as separate classes.

PR #351, opened 2026-05-23
feat: Allow JLineBackend to accept an existing Terminal

gnodet:
## Problem

When a TamboUI application runs inside a host that already owns the system terminal (e.g., a JLine-based interactive shell), `JLineBackend()` tries to create a second system terminal via `TerminalBuilder.builder().system(true).build()`. This fails with a `Stream Closed` IOException because stdin is already consumed by the host shell's terminal.

## Solution

Add a `JLineBackend(Terminal terminal)` constructor that accepts an existing JLine `Terminal` instance:

- The caller retains ownership — the backend **will not close** the terminal on `close()`
- The existing no-arg constructor is unchanged (creates and owns a new system terminal)
- A private `JLineBackend(Terminal, boolean ownTerminal)` delegates both constructors

Combined with the existing `TuiConfig.builder().backend(backend).build()` path, host applications can now pass their terminal through:

```java
Terminal shellTerminal = getActiveTerminal();
Backend backend = new JLineBackend(shellTerminal);
try (var tui = TuiRunner.create(TuiConfig.builder().backend(backend).build())) {
    tui.run(handler, renderer);
}
```

## Use case

Apache Camel JBang has an interactive shell (`camel shell`) built on JLine. Subcommands like `tui` (a TamboUI dashboard) need to reuse the shell's terminal rather than creating their own. Without this change, running `tui` from inside `camel shell` crashes immediately.

### DIFF
--- a/tamboui-aesh-backend/src/main/java/dev/tamboui/backend/aesh/AeshBackend.java
+++ b/tamboui-aesh-backend/src/main/java/dev/tamboui/backend/aesh/AeshBackend.java
@@ -26,6 +26,15 @@ import dev.tamboui.terminal.Mode2027Support;
  * <p>
  * This backend uses the aesh-readline library's TerminalConnection abstraction
  * for terminal I/O operations.
+ * <p>
+ * Two construction modes are supported:
+ * <ul>
+ *   <li>{@link #AeshBackend()} — creates and owns a system terminal connection;
+ *       {@link #close()} will close it.</li>
+ *   <li>{@link #AeshBackend(Connection)} — uses an externally provided connection
+ *       (e.g., from SSH or HTTP); the caller retains ownership and {@link #close()}
+ *       will <em>not</em> close the connection.</li>
+ * </ul>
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

davsclaus:
This would make Camel TUI more awesome so appreciate if this can be merged.

maxandersen:
Thanks @gnodet — this is the right approach for embedding TamboUI in a host that already owns the terminal.

I pushed a commit on top with a few additions:

- Added `Objects.requireNonNull` on the `Terminal` parameter
- Added class-level, constructor, and `close()` javadoc documenting ownership semantics
- Applied the same ownership pattern to `AeshBackend`, which already accepted an external `Connection` but was missing clear ownership semantics — it was always closing the connection, causing a double-close in our SSH/HTTP demo

Both backends now follow the same contract: no-arg constructor owns and closes, external instance constructor leaves ownership with the caller.

PR #355, opened 2026-05-24
Fix JLine terminal close hanging on macOS

### LINKED ISSUE #354: JLine terminal close hangs on macOS due to stdin FD deadlock
## Description

When the TUI exits (e.g., pressing `q`), `TuiRunner.close()` hangs indefinitely on macOS. The process requires `Ctrl+C` or `kill` to terminate.

## Root Cause

Classic JLine/macOS deadlock between two threads operating on the same stdin file descriptor:

- **Main thread** calls `JLineBackend.close()` → `AbstractTerminal.close()` → `NonBlockingInputStreamImpl.close()` → `FileInputStream.close()` → `FileDescriptor.close0()` (native) — **blocks**
- **JLine reader thread** is stuck in `FileInputStream.read0()` (native blocking read) on the same stdin FD

On macOS, `close()` on an FD that another thread is `read()`ing from blocks indefinitely.

## Thread Dump (relevant threads)

```
"main" #1 prio=5 - RUNNABLE
  at java.io.FileDescriptor.close0(Native Method)
  - locked <0x000000040033de20> (a java.io.FileDescriptor)
  at java.io.FileDescriptor.close(FileDescriptor.java:304)
  at java.io.FileInputStream$1.close(FileInputStream.java:526)
  at java.io.FileDescriptor.closeAll(FileDescriptor.java:362)
  at java.io.FileInputStream.close(FileInputStream.java:524)
  at org.jline.utils.NonBlockingInputStreamImpl.close(NonBlockingInputStreamImpl.java:79)
  at org.jline.utils.NonBlocking$NonBlockingInputStreamReader.close(NonBlocking.java:264)
  at org.jline.terminal.impl.AbstractUnixSysTerminal.doClose(AbstractUnixSysTerminal.java:236)
  at org.jline.terminal.impl.AbstractTerminal.close(AbstractTerminal.java:180)
  at dev.tamboui.backend.jline3.JLineBackend.close(JLineBackend.java:296)
  at dev.tamboui.tui.TuiRunner.cleanup(TuiRunner.java:760)
  at dev.tamboui.tui.TuiRunner.close(TuiRunner.java:729)

"JLine terminal non blocking reader thread" #42 daemon - RUNNABLE
  at java.io.FileInputStream.read0(Native Method)
  at java.io.FileInputStream.read(FileInputStream.java:231)
  at org.jline.utils.NonBlockingInputStreamImpl.run(NonBlockingInputStreamImpl.java:206)
```

## Suggested Fix

In `JLineBackend.close()`, interrupt the JLine reader thread or call `NonBlockingInputStreamImpl.shutdown()` before closing the terminal. This would unblock the reader thread's native `read()` call, allowing `close()` to proceed.

## Environment

- macOS (Darwin 24.6.0)
- JDK 21.0.5+11-LTS
- JLine (via tamboui-backend-jline3)

davsclaus:
Fixes #354

## Summary

- Run `terminal.close()` in a daemon thread with a 2-second timeout in `JLineBackend.close()` to work around a macOS deadlock where `FileDescriptor.close0()` blocks indefinitely when JLine's internal reader thread is in a native `read0()` on stdin
- Interrupt the input reader thread in `TerminalInputReader.stop()` so it exits its JLine `read()` wait faster

## Root Cause

On macOS, the `close()` syscall on a file descriptor blocks if another thread is in a `read()` syscall on the same FD. JLine's internal "non blocking reader thread" sits in `FileInputStream.read0()` on stdin, and `Thread.interrupt()` has no effect on native I/O. When `terminal.close()` calls `FileDescriptor.close0()`, it deadlocks.

## Approach

All terminal state (raw mode, alternate screen, cursor, mouse capture) is already restored before `terminal.close()` is called. The close just releases the underlying streams. By running it in a daemon thread with a timeout, if it hangs (macOS), the JVM exits cleanly. If it works (Linux), it completes within the timeout.

## Thread dump showing the deadlock

```
"main" RUNNABLE
  at java.io.FileDescriptor.close0(Native Method)  ← blocks here
  at java.io.FileInputStream.close(...)
  at org.jline.utils.NonBlockingInputStreamImpl.close(...)
  at dev.tamboui.backend.jline3.JLineBackend.close(...)

"JLine terminal non blocking reader thread" RUNNABLE
  at java.io.FileInputStream.read0(Native Method)   ← stuck here
  at org.jline.utils.NonBlockingInputStreamImpl.run(...)
```

## Test plan

- [ ] Start any TamboUI TUI app on macOS, press `q` to quit — should exit cleanly without hanging
- [ ] Same test on Linux — should still exit cleanly (close completes within timeout)

### DIFF
--- a/tamboui-jline3-backend/src/main/java/dev/tamboui/backend/jline3/JLineBackend.java
+++ b/tamboui-jline3-backend/src/main/java/dev/tamboui/backend/jline3/JLineBackend.java
@@ -293,7 +293,27 @@ public class JLineBackend extends AbstractBackend {
         disableRawMode();
 
         writer.flush();
-        terminal.close();
+
+        // Close terminal with a timeout to work around a macOS deadlock:
+        // JLine's internal reader thread blocks in native FileInputStream.read0()
+        // on stdin, and FileDescriptor.close0() blocks until that read completes.
+        // On macOS, close() never unblocks a concurrent read() on the same FD.
+        // All terminal state is already restored above, so skipping close on
+        // timeout is safe — JVM exit handles FD cleanup.
+        Thread closeThread = new Thread(() -> {
+            try {
+                terminal.close();
+            } catch (IOException e) {
+                // best effort
+            }
+        }, "jline-terminal-close");
+        closeThread.setDaemon(true);
+        closeThread.start();
+        try {
+            closeThread.join(2000);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
     }
 
     /**
--- a/tamboui-tui/src/main/java/dev/tamboui/tui/TerminalInputReader.java
+++ b/tamboui-tui/src/main/java/dev/tamboui/tui/TerminalInputReader.java
@@ -74,6 +74,7 @@ public final class TerminalInputReader implements Runnable {
     public void stop(long timeoutMs) {
         Thread t = thread;
         if (t != null && t.isAlive()) {
+            t.interrupt();
             try {
                 t.join(timeoutMs);
             } catch (InterruptedException e) {


davsclaus:
We have a temporary workaround in Camel TUI but would like to have this fixed out of the box in tamboui

tamboui-jline3-backend/src/main/java/dev/tamboui/backend/jline3/JLineBackend.java
```diff
@@ -293,7 +293,27 @@ public void close() throws IOException {
         disableRawMode();
 
         writer.flush();
-        terminal.close();
+
+        // Close terminal with a timeout to work around a macOS deadlock:
```
maxandersen, line 297: this is hopefully not the *right* fix? this seems very wrong we have to go through creating thread to  close it?

davsclaus: Without this you have a deadlock, its the JDK HTTP server that creates a non daemon thread that hangs

davsclaus: Sorry, my earlier reply was imprecise — the deadlock is specifically between JLine's internal "non blocking reader thread" (stuck in native `FileInputStream.read0()` on stdin) and the main thread calling `FileDescriptor.close0()` during `terminal.close()`.

This is a known macOS kernel behavior: `close()` on a file descriptor blocks indefinitely if another thread is in a `read()` on the same FD. On Linux, `close()` unblocks the concurrent `read()`, so it works fine there. Since the blocking happens in native code, `Thread.interrupt()` has no effect.

The daemon-thread-with-timeout approach is the standard JVM workaround for this — it's what other terminal libraries do as well. The key safety argument is that all terminal state (raw mode, alternate screen, cursor, mouse capture) is already fully restored *before* `terminal.close()` is called, so the close only releases the underlying streams. If it hangs on macOS, the JVM exit handles FD cleanup. If it completes (Linux), it finishes well within the 2-second timeout.

So no, it's not the "ideal" fix — the real fix would need to come from JLine or the JDK — but it is the correct workaround at this layer.

davsclaus: To be more precise about precedent — this is a well-known issue in JLine itself, documented across several issues:

- [jline/jline3#167](https://github.com/jline/jline3/issues/167) — exact same deadlock: reader thread holds lock in `FileInputStream.readBytes()`, shutdown hook blocked on `InputStreamReader.close()`
- [jline/jline3#31](https://github.com/jline/jline3/issues/31) — suggested fix was to make the pump thread a daemon thread so the JVM can exit
- [jline/jline3#128](https://github.com/jline/jline3/issues/128) — workaround was to skip closing `terminal.reader()` entirely
- [jline/jline3#1361](https://github.com/jline/jline3/issues/1361) — macOS-specific hang in `FileDescriptor.close0()`, fixed in PR #1382 by removing PTY terminal usage on macOS for pipe operations

Our approach combines the ideas from #31 (daemon thread so JVM can exit) and #128 (safe to skip the close since terminal state is already restored). I shouldn't have said "standard workaround" — it's more that the building blocks (daemon threads, skipping close) are the same ones JLine's own maintainers have used.

davsclaus: Would be good with feedback from @gnodet as maybe jline itself can have a fix if possible

davsclaus:
closed, see comment in ticket

PR #357, opened 2026-05-25
fix: Ctrl+letter and Shift+letter keys dropped during tape replay

davsclaus:
## Summary

- Fix `InteractionPlayer.enqueueKey()` where `Ctrl+letter` key combinations (e.g., `Ctrl+t`, `Ctrl+d`) were silently dropped during tape replay. The modifier prefix was stripped from the lowercase copy but `keySpec` still held the full string — `keySpec.length() == 1` failed for `"ctrl+t"` (length 6), so no byte was ever emitted.
- Also fix `parseVhsCommand()` to accept `Shift+key` patterns (previously only `Ctrl+` was allowed through).
- Add `InteractionPlayerTest` with 4 test cases covering Ctrl+letter, Shift+letter, typed text, and mixed tape sequences.

## Test plan

- [x] All 4 new tests pass
- [ ] Verify `Ctrl+t` in a tape file produces byte 0x14 during replay
- [ ] Verify `Shift+D` in a tape file produces uppercase 'D' during replay

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/internal/record/InteractionPlayer.java
+++ b/tamboui-core/src/main/java/dev/tamboui/internal/record/InteractionPlayer.java
@@ -272,8 +272,8 @@ final class InteractionPlayer {
                 break;
 
             default:
-                // Check for Ctrl+key patterns
-                if (command.startsWith("ctrl+")) {
+                // Check for modifier+key patterns (Ctrl+x, Shift+x)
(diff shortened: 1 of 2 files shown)


davsclaus:
I have verified this in a tape recording that ctrl + key and shift + key now works. 
We use this feature in Camel TUI

maxandersen:
makes sense! thanks!

PR #360, opened 2026-05-28
Fix arrow key handling on Windows terminals (Panama backend)

### LINKED ISSUE #359: Arrow keys are parsed as UNKNOWN on Windows Panama backend
### Summary
Arrow key input does not work on Windows when using the Panama backend.

Pressing Up, Down, Left, or Right does not produce the expected KeyCode.UP, KeyCode.DOWN, KeyCode.LEFT, or KeyCode.RIGHT events. Instead, the input is parsed as UNKNOWN or ignored by the application.

### Environment

* OS: Windows
* Backend: tamboui-panama-backend
* Runtime: GraalVM native image

### Expected behavior

Arrow keys should be parsed as navigation key events:

* Up -> KeyCode.UP
* Down -> KeyCode.DOWN
* Left -> KeyCode.LEFT
* Right -> KeyCode.RIGHT

### Actual behavior

Arrow keys are not recognized correctly. In TUI applications, this makes list navigation and selection movement unusable on Windows.

acltabontabon:
## Quick Gist
Arrow key navigation (Up, Down, Left, Right) did not work in TUI applications running on Windows as a native image. Pressing an arrow key produced no response or triggered an UNKNOWN key event instead of KeyCode.UP / KeyCode.DOWN / etc.

## Bug
`WindowsTerminal.peek()` did **not** return the next character. It only checked
whether any input events were queued and returned a flag:

```java
// BEFORE (broken)
@Override
public int peek(int timeoutMs) throws IOException {
    if (Kernel32.getNumberOfConsoleInputEvents(inputHandle, intBuffer) == 0) {
        throw new RuntimeIOException("...");
    }
    return intBuffer.get(ValueLayout.JAVA_INT, 0) > 0 ? 0 : -2;
    //                                                   ^
    //                           always 0 when events exist, never the actual char
}
```

## Workaround
https://github.com/acltabontabon/mod-scope/commit/eeeb89dd1b090de49605ca1ea821f230ce169721

---

Fixes #359

### DIFF
--- a/tamboui-panama-backend/src/main/java/dev/tamboui/backend/panama/windows/WindowsTerminal.java
+++ b/tamboui-panama-backend/src/main/java/dev/tamboui/backend/panama/windows/WindowsTerminal.java
@@ -4,16 +4,15 @@
  */
 package dev.tamboui.backend.panama.windows;
 
+import dev.tamboui.backend.panama.PlatformTerminal;
+import dev.tamboui.error.RuntimeIOException;
+import dev.tamboui.layout.Size;
 import java.io.IOException;
 import java.lang.foreign.Arena;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.ValueLayout;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-
-import dev.tamboui.backend.panama.PlatformTerminal;
-import dev.tamboui.error.RuntimeIOException;
-import dev.tamboui.layout.Size;
 
 /**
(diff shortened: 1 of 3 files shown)


maxandersen:
Good fix — this is a real bug that's still present on main. The `peek()` contract requires returning the actual next character, and the old implementation just returned `0` when events were available, which broke escape sequence detection for arrow keys.

I rebased your branch onto current main to resolve the test conflicts (the F-key tests from #372 and your arrow key tests now coexist). Force-pushed since it was a rebase. All tests pass.

PR #362, opened 2026-05-31
Fix MarkdownView to render partial code blocks

### LINKED ISSUE #361: MarkdownView skips code blocks that don't fully fit in the viewport
## Problem

`MarkdownView` completely skips rendering fenced code blocks (and other `WidgetChunk`-based elements like tables) when they don't fully fit within the remaining viewport height. Instead of showing a partial code block clipped at the bottom, the space is left blank — the user sees nothing and has no indication a code block exists there.

## Root Cause

In `MarkdownView.render()` (line 268), the rendering guard requires both conditions to be true:

```java
if (chunkSkip == 0 && visibleHeight == chunkHeight) {
    Rect target = new Rect(contentArea.left(), y, contentArea.width(), chunkHeight);
    chunk.render(target, buffer);
}
```

The `visibleHeight == chunkHeight` condition means the **entire** code block must fit in the remaining screen space. If a 20-line code block appears near the bottom of the viewport with only 10 lines of space left, it is completely invisible.

## Expected Behavior

A code block near the bottom of the viewport should render partially — showing as many lines as fit, with the Block border clipped naturally by the buffer bounds. The user can then scroll down to see the rest.

## Suggested Fix

Remove the `visibleHeight == chunkHeight` guard and pass `visibleHeight` instead of `chunkHeight` to the render target:

```java
if (chunkSkip == 0) {
    Rect target = new Rect(contentArea.left(), y, contentArea.width(), visibleHeight);
    chunk.render(target, buffer);
}
```

The `chunkSkip == 0` guard should remain since scrolling past the top of a bordered widget is harder to handle (the Block draws borders from the top). But the height restriction is unnecessary — `Paragraph` already clips lines that don't fit, and the `Block` bottom border gets clipped naturally.

## How to Reproduce

Use `MarkdownView` with a markdown string containing a fenced code block tall enough that it doesn't fit at the bottom of the viewport. Scroll so the code block is partially visible — it will disappear entirely instead of showing the top portion.

davsclaus:
## Summary

- Fix `MarkdownView` to render fenced code blocks and tables even when they don't fully fit in the remaining viewport height
- Previously, `WidgetChunk`-based elements were completely skipped if `visibleHeight != chunkHeight`, leaving blank space
- Now passes the available `visibleHeight` to the widget's render target, allowing `Block` and `Paragraph` to clip naturally

## Changes

Removed the `visibleHeight == chunkHeight` guard in `MarkdownView.render()` and pass `visibleHeight` instead of `chunkHeight` to the `Rect` target. The `chunkSkip == 0` guard is preserved since scrolling past the top of a bordered widget requires more work.

## Tests

- Added `rendersPartialCodeBlock` — verifies a 5-row code block renders correctly when only 3 rows are available (top border + 1 content line + bottom border)
- Added `rendersPartialCodeBlockAfterText` — verifies a code block preceded by a paragraph renders its top border even when only 2 of 3 rows fit

Closes #361

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of Claus Ibsen

### DIFF
--- a/tamboui-markdown/src/main/java/dev/tamboui/markdown/MarkdownView.java
+++ b/tamboui-markdown/src/main/java/dev/tamboui/markdown/MarkdownView.java
@@ -261,12 +261,11 @@ public final class MarkdownView implements Widget {
                 continue;
             }
             if (chunk instanceof WidgetChunk) {
-                // Block-bordered widgets (code blocks, tables) need their full
-                // geometry to render correctly — a partial slice paints only
-                // the top border with no closure. Skip when we can't fit it
-                // whole; the user scrolls to align with its top.
-                if (chunkSkip == 0 && visibleHeight == chunkHeight) {
-                    Rect target = new Rect(contentArea.left(), y, contentArea.width(), chunkHeight);
+                // Render block-bordered widgets (code blocks, tables) when
+                // their top is visible. Paragraph and Block clip naturally
+                // when the available height is less than the full chunk.
+                if (chunkSkip == 0) {
+                    Rect target = new Rect(contentArea.left(), y, contentArea.width(), visibleHeight);
                     chunk.render(target, buffer);
                 }
             } else if (chunk instanceof LinesChunk) {
--- a/tamboui-markdown/src/main/java/dev/tamboui/markdown/MarkdownView.java
+++ b/tamboui-markdown/src/main/java/dev/tamboui/markdown/MarkdownView.java
@@ -261,12 +261,22 @@ public final class MarkdownView implements Widget {
                 continue;
             }
             if (chunk instanceof WidgetChunk) {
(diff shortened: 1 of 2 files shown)


maxandersen:
Thanks for the fix! I extended it to also handle the scroll-up case — when a code block's top is scrolled above the viewport, it was previously hidden entirely (the `chunkSkip == 0` guard). 

The new commit renders the widget to a scratch buffer at full height and copies only the visible rows, so the bottom portion of a code block stays visible as you scroll past it. Added two tests for the scroll-off-top cases.

I pushed directly to your branch — hope that's OK.

davsclaus:
Absolutely you are welcome. Use as much or as little you can - at the end if we can make tamboui better.

maxandersen:
merging as this fairly local fix and not worse than what was there before :)

PR #363, opened 2026-06-04
Fix Memory Leak in a Terminal When Showing an Image

cdhermann:
## Problem

Displaying an image with the native protocols (Kitty, iTerm2, Sixel) in a render
loop transmitted the full payload **every frame**. On iTerm2 — which retains every
inline image it receives — this grew the terminal process memory without bound
(observed: a viewer climbing to ~150 GB before the terminal had to be killed).
Large images additionally froze the UI for many seconds and left stacked, never-
cleared copies on screen.

## Changes

**Memory leak**
- `NativeImageCache`: skip re-transmission when the same image is already shown at
  the same position; cache encoded payloads weakly; assign stable Kitty image/
  placement ids so repeats replace instead of accumulating.
- `KITTY_DELETE_ALL`: `d=a` → `d=A` so image data is actually freed.
- New `RawOutputContext` screen *generation*, bumped by `Terminal` on `clear()`/
  `resize()`, so images transparently survive a screen clear.

**Large-image robustness**
- Cap the resolution transmitted to self-scaling protocols (Kitty/iTerm2) to the
  display footprint × a supersampling factor — kills multi-second freezes and the
  "a file is being downloaded" stuck state, while staying crisp.
- Clear stale overlapping footprints on emit (iTerm2/Sixel: spaces; Kitty: delete
  old image id) — no more stacked images on scaling/position changes.

## Tests

- New `NativeImageProtocolSkipTest` (16 cases): skip, content/position/generation
  changes, overwrite-return regression — across all three protocols.
- `RawOutputCleanupTest` updated for `d=A`.
- Full core + image suites green.

## Verified

Manually in iTerm2: memory grows slowly, on Ghostty it stays flat, switching scaling/protocols is instant,
no artifacts.

## Known / out of scope

- One-time PNG-encode cost on the first frame per image (cached afterwards).

### DIFF
--- a/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
+++ b/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
@@ -7,7 +7,6 @@ package dev.tamboui.image.protocol;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.Base64;
 import java.util.List;
 
 import dev.tamboui.buffer.Buffer;
@@ -86,7 +85,7 @@ public final class ITermProtocol implements ImageProtocol {
 
         // Encode image as PNG then base64. The payload depends only on the pixels, so it is
         // cached per image to avoid re-encoding on every frame of the render loop.
-        String base64Data = cache.payload(image, () -> encodeBase64(image));
+        String base64Data = cache.payload(image, () -> NativeImageCache.encodeBase64(image));
 
         // Build the iTerm2 escape sequence
         StringBuilder cmd = new StringBuilder();
@@ -108,18 +107,6 @@ public final class ITermProtocol implements ImageProtocol {
 
         rawOutput.write(cmd.toString().getBytes(StandardCharsets.US_ASCII));
         rawOutput.flush();
-    }
-
-    /**
-     * Encodes the image as base64-encoded PNG, wrapping the checked IO exception so it
-     * can be used from a cache supplier.
-     */
-    private static String encodeBase64(ImageData image) {
-        try {
-            return Base64.getEncoder().encodeToString(image.toPng());
-        } catch (IOException e) {
-            throw new RuntimeIOException("Failed to encode image as PNG for iTerm2 protocol", e);
-        }
     }
 
     @Override
--- a/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
+++ b/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
+import java.util.List;
 
 import dev.tamboui.buffer.Buffer;
 import dev.tamboui.error.RuntimeIOException;
@@ -71,9 +72,13 @@ public final class ITermProtocol implements ImageProtocol {
         // process memory without bound. The image stays on screen because the diff-based
         // renderer leaves the image cells untouched between frames. A change in screen
         // generation (clear/resize) forces a redraw.
-        if (!cache.needsEmit(image, area, NativeImageCache.generationOf(rawOutput))) {
+        List<Rect> stale = cache.staleAreasToClear(image, area, NativeImageCache.generationOf(rawOutput));
+        if (stale == null) {
             return;
         }
+        // Wipe any previously shown image whose footprint this one does not fully cover, so a
+        // shrinking image (e.g. FILL -> FIT) does not leave the larger one behind.
+        NativeImageCache.clearAreas(stale, rawOutput);
 
         // Move cursor to position
         String cursorMove = String.format("\033[%d;%dH", area.y() + 1, area.x() + 1);
--- a/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
+++ b/tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
@@ -37,6 +37,7 @@ public final class ITermProtocol implements ImageProtocol {
     private static final String ST = "\033\\";  // Alternative terminator
 
     private final boolean useStTerminator;
+    private final NativeImageCache cache = new NativeImageCache();
 
     /**
      * Creates an iTerm2 protocol instance using BEL terminator.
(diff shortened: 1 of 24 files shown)


cdhermann:
Even iTerm2 should support the kitty protocol, it still doesn't work flawlessly because `00` a the end sneak through and iTerm2 treats it as keyboard input.

maxandersen:
thanks - I tried making a demo app to stress test, I merged that to main in #364 

What's weird for me though is the memory leak shows up in Ghostty for me, but not in iterm2 - but in term2 input stops working so something is definitely not right ;)

maxandersen:
ok so with stress test that is on main this PR definitely makes the memory use issue go away.

BUT changing image every time does not work in Ghostty iterm2 protocol (it just stays stuck on first image), while in iterm2 something is off because I end up with this:

Screenshot 2026-06-06 at 11 21 01

something is not fully right here. I'll keep looking but cool if @cdhermann could confirm same findings and if you have idea/improvements in the PR to make it catch the issues you are very welcome ;)

tamboui-image/src/main/java/dev/tamboui/image/protocol/NativeImageCache.java
```diff
@@ -0,0 +1,220 @@
...
+ * content changes (e.g. flipping PDF pages) — reuses the same id, so the terminal
+ * <em>replaces</em> the stored image instead of piling up copies. This bounds terminal
+ * memory to O(number of distinct image positions on screen) regardless of how many frames
+ * are drawn or how many images are shown over time.
+ *
+ * <h2>Redraw suppression</h2>
+ * Even with a stable id, re-sending the payload on every frame is itself a leak source for
+ * the cell-based protocols: iTerm2 decodes and retains every inline image it receives, so a
+ * render loop that re-transmits the same picture every frame grows the terminal process
+ * without bound. {@link #needsEmit(ImageData, Rect)} returns {@code false} when the same
```
maxandersen: Stale reference — `needsEmit` doesn't exist; this should be `staleAreasToClear`. Same on line 91 (`The screen generation observed at the last needsEmit call`).

tamboui-image/src/main/java/dev/tamboui/image/protocol/NativeImageCache.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Decides whether the given image must be (re)transmitted and, if so, returns the previously
+     * shown footprints that this emission must clear from the screen first.
+     * <p>
+     * Returns {@code null} when the same image is already shown at the same footprint (skip).
+     * Otherwise records the new footprint and returns the list (possibly empty) of stale
+     * footprints whose pixels overlap the new one. Those must be wiped before drawing so that a
+     * shrinking footprint (e.g. FILL -> FIT) does not leave the larger previous image behind and
+     * repeated emissions at different positions do not stack up on screen.
```
maxandersen: There's an edge case here: if the image moves to a *non-overlapping* position and then back, the old entry stays in `shownByPosition` even though `Terminal.cleanupRawOutput()` already wiped it from screen. The return trip would incorrectly skip.

Since a protocol instance only renders one image at a time, tracking just the last emitted `(area, image)` pair instead of a full map would be simpler and avoid this. Alternatively, clearing entries for areas not in the current emission would work too.

Would be great to add a test for the non-overlapping move-and-return case alongside the fix.

tamboui-image/src/main/java/dev/tamboui/image/protocol/ITermProtocol.java
```diff
@@ -94,6 +110,18 @@ public void render(ImageData image, Rect area, Buffer buffer, OutputStream rawOu
         rawOutput.flush();
     }
 
+    /**
```
maxandersen: This is identical to `KittyProtocol.encodeBase64()` — could live as a static helper on `NativeImageCache` to avoid the duplication. Minor though, happy with a follow-up too.

maxandersen (changes requested):
Nice fix for a real and severe issue — the core approach is solid. A few things to address before merging.

maxandersen:
fyi, just tested with kitty too and it behaves slightly different too. still with overlapping images unfortunately.

cdhermann:
> but in term2 input stops working so something is definitely not right ;)

I experiences that behavior when I was using the Kitty protocol in iTerm2. I think iTerm2 missed parsing some control sequences from Kitty which makes it stuck. Using Kitty protocol I saw images in iTerm2 but I couldn't input data. iTerm2 works for me only with the iTerm2 protocol.

I'll give your load test demo a try

cdhermann:
Thanks for the review and the stress-test demo! Pushed updates:

- **Review points:** last-`(area,image)`-pair tracking (fixes the move-and-return skip) + test, `encodeBase64` deduped, javadoc fixed.
- **Overlap / "stuck":** Kitty image id is now derived from the area (was a per-instance counter, so a new protocol instance orphaned the old image) + previous footprint is cleared on move/resize.
- **Kitty terminals:** delete commands now also set `q=2` (the transmit already did) so the terminal's `OK` reply isn't read as input.
- **Freeze on big images:** transmitted resolution is capped to the display size.
- **macOS focus steal:** `tamboui-image` defaults to `java.awt.headless=true`.

Verified clean in Ghostty + iTerm2 (auto, manual, scaling, switching, resize; memory flat). Forcing Kitty in iTerm2 stays limited — iTerm2 doesn't seem to fully implement the Kitty protocol. Rationale per change in `docs/native-image-rendering.md`.

Sixel I haven't tested so far.

maxandersen:
What's the macfocus steal? Is that it grabs focus to the started dock app?

cdhermann:
> What's the macfocus steal? Is that it grabs focus to the started dock app?

Exactly. The terminal looses the focus. The started java app in the macOS dock receives the focus. Forcing AWT to act as headless disables that behavior

cdhermann:
@maxandersen, I found another issue when dealing with the merge conflict:

When cycling through the modes, after HIDDEN mode, there is a black rectangle in the image where the text was. I try do deal with that now.

cdhermann:
> When cycling through the modes, after HIDDEN mode, there is a black rectangle in the image where the text was. I try do deal with that now.

When I cycle SAME → CHANGING → HIDDEN → SAME in iTerm2, a dark hole stays in the image, exactly where the "hidden" text was before. In Ghostty it works fine.

I think the reason is this: in iTerm2 (and Sixel) the image lives inside the terminal cells. The image is sent during renderer.accept(), but the cell diff that removes the old "hidden" text runs after that in the same draw. So it overwrites the cells where the image already was, and we get a hole. Before, we sent the image every frame, so the next frame did draw over the hole again. Now we skip unchanged frames (that is the whole point of the memory fix), so the hole stays. Kitty is not affected, because its images are on a separate layer and cell writes do not touch them.

I think the correct fix is in the framework: send the raw output after the cell diff in Terminal.draw, so the image is always drawn on top of the cells. This would be correct for iTerm2 and maybe also Sixel, and makes no difference for Kitty. It is a change, but it is in the core render path, my assumption is.

For now, I only wrote it down as a known limitation in docs/native-image-rendering.md. I'm not sure how to proceed now. I'm not sure if I should touch the core render path.

maxandersen:
Interesting. What would be the changes you would do to core rendering path look like ? Maybe alternate pr ?

Also would be good to see what ratatui actually does in this case (if anything at all)

cdhermann:
> Interesting. What would be the changes you would do to core rendering path look like ? Maybe alternate pr ?

Currently, I am pretty busy but I try to provide an alternate PR as soon as I find some spare time.

cdhermann:
> Interesting. What would be the changes you would do to core rendering path look like ? Maybe alternate pr ?

I added the PR https://github.com/tamboui/tamboui/pull/379 for that.

For testing:

1. press 2 for the iTerm2 protocol or 3 for Sixel in the image-stress-test
2. c → CHANGING, c → HIDDEN, c → back to SAME.
3. Switch between hidden and image changes a few times.
4. Expectation: Upon returning, there should be no dark bar/hole—the image is now flushed after the cell diff and appears cleanly on top.

maxandersen:
So is https://github.com/tamboui/tamboui/pull/363 replacing this or combination?

PR #374, opened 2026-06-10
fix: exclude .settings directory from gradle demo module scanning

### LINKED ISSUE #373: .settings directory generated by VSCode Java Language Server causes Gradle build failure
After setting up the JDK and Gradle configurations, when opening this project for the first time in VSCode, the compilation was successful.

When you open or reload the Window for the second time, the following error will occur.

Image

MrLiuzy:
VSCode Java Language Server generates a .settings directory under the project root, which was incorrectly picked up by includeDemosFrom() and treated as a Gradle subproject, causing a naming convention error.

Fix #373

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -27,7 +27,7 @@ val modules = listOf(
 include(*modules.toTypedArray())
 
 fun includeDemosFrom(demosDir: File, projectPathPrefix: String) {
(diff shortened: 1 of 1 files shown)


melix (approved):
Darn IDEs polluting directories...

maxandersen (approved):

PR #375, opened 2026-06-13
feat(toolkit): add title alignment to Panel

jlc488:
## What
`Panel` always rendered its title left-aligned, even though the underlying
`Block`/`Title` widget already supports left/center/right alignment. This
exposes title alignment on the toolkit `Panel` element.

## Changes
- `titleAlignment(Alignment)` + `titleLeft()` / `titleCenter()` / `titleRight()`
- `bottomTitleAlignment(Alignment)` for the bottom title
- Alignment resolves as **explicit > CSS `text-align` > LEFT**, matching the
  element's other CSS-backed properties (the "explicit > CSS > default" pattern
  documented in AGENTS.md)

## Tests
`PanelTest` asserts the rendered title position for each alignment, the CSS
`text-align` path, and explicit-overrides-CSS.

## Docs
Added a snippet + section to `api-levels.adoc`.

## Notes for reviewers
The CSS path reuses the Panel's own `text-align` (via `CssStyleResolver.alignment()`).
Happy to switch to a dedicated child selector (e.g. `Panel-title`) or a new
property instead if you'd prefer to keep `text-align` free for other uses.

### DIFF
--- a/demos/flex-demo/src/main/java/dev/tamboui/demo/flex/FlexDemo.java
+++ b/demos/flex-demo/src/main/java/dev/tamboui/demo/flex/FlexDemo.java
@@ -9,6 +9,7 @@ package dev.tamboui.demo.flex;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import dev.tamboui.layout.Alignment;
 import dev.tamboui.layout.Flex;
 import dev.tamboui.style.Color;
 import dev.tamboui.toolkit.app.ToolkitRunner;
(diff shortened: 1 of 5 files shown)


maxandersen:
Nice improvement! The underlying `Title` widget already had alignment support and it was a real gap that `Panel` didn't expose it. Clean implementation, good tests.

I pushed a follow-up commit (7597aa4f) that adds two things:

**1. Pre-existing bug fix:** `titleOverflow` (ellipsis etc.) was never applied to the bottom title — only the top title got `.overflow(titleOverflow)`. Your PR carried this forward, so I added the one-liner plus a regression test.

**2. Flex demo update:** Updated Example 4 (Toolbar) to showcase all three alignments in context:
- Left-aligned title (default) on "Simple toolbar"
- `.titleCenter()` on "Status bar"
- `.titleRight()` + `.bottomTitle("ESC to close").bottomTitleAlignment(Alignment.CENTER)` on "Menu bar"

Shows off the new flexibility without needing a separate demo.

maxandersen:
Nice one - thanks @jlc488 !

@all-contributors add @jlc488 for code

PR #379, opened 2026-06-17
flush native image output after the cell diff in Terminal.java

cdhermann:
Native images were written while widgets rendered, before the cell diff.
The diff that clears the previous frame's text from a cell-based image's
cells (iTerm2, Sixel) then punched a hole in the image emitted during the
same draw. Terminal now buffers the frame's raw output and flushes it after
backend.draw(diff), so graphics composite on top of the freshly drawn cells.
cleanupRawOutput stays before the diff. Kitty (separate graphics layer) was
unaffected.

### DIFF
--- a/demos/aesh-ssh-http-demo/src/main/java/dev/tamboui/demo/aesh/AeshSshHttpDemo.java
+++ b/demos/aesh-ssh-http-demo/src/main/java/dev/tamboui/demo/aesh/AeshSshHttpDemo.java
@@ -1,7 +1,7 @@
 //DEPS dev.tamboui:tamboui-toolkit:LATEST
 //DEPS dev.tamboui:tamboui-aesh-backend:LATEST
-//DEPS org.aesh:terminal-ssh:3.1
-//DEPS org.aesh:terminal-http:3.1
+//DEPS org.aesh:terminal-ssh:3.12
+//DEPS org.aesh:terminal-http:3.12
 //DEPS org.apache.sshd:sshd-core:2.14.0
 //DEPS org.apache.sshd:sshd-netty:2.14.0
 //DEPS io.netty:netty-all:4.1.81.Final
@@ -98,8 +98,8 @@ public class AeshSshHttpDemo implements java.util.function.Consumer<Connection> 
     private void startSshServer() {
         try {
             NettySshTtyBootstrap bootstrap = new NettySshTtyBootstrap();
-            bootstrap.setPort(SSH_PORT);
-            bootstrap.setHost("localhost");
+            bootstrap.port(SSH_PORT);
+            bootstrap.host("localhost");
             bootstrap.start(this).get(10, TimeUnit.SECONDS);
             System.out.println("SSH server started on port " + SSH_PORT);
         } catch (Exception e) {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 jline = "3.25.1"
-aesh-terminal = "3.2"
+aesh-terminal = "3.12"
 junit = "5.14.3"
 assertj = "3.27.7"
 native-build-tools = "0.11.4"
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -67,6 +67,9 @@
     "image-demo": {
       "script-ref": "tamboui-image/demos/image-demo/src/main/java/dev/tamboui/demo/ImageDemo.java"
     },
+    "image-stress-test": {
+      "script-ref": "tamboui-image/demos/image-stress-test/src/main/java/dev/tamboui/demo/ImageStressTest.java"
+    },
     "markdown-demo": {
       "script-ref": "tamboui-markdown/demos/markdown-demo/src/main/java/dev/tamboui/demo/MarkdownDemo.java"
     },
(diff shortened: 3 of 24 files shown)


maxandersen:
**Moving this (and the linked #363) to the 0.6 milestone.**

Neither PR blocks the 0.5 / Camel-enablement goal — they're image-rendering correctness, not backend selection — so they shouldn't gate cutting 0.5. Parking them together for 0.6.

More importantly: before merging the flush-after-diff approach, we should look at how **ratatui-image** solves the exact same "cell-based image gets a hole punched by the diff" problem, because it uses a smaller, more surgical mechanism that also collapses the memory fix (#363) and the hole fix into one. TL;DR — instead of reordering the frame flush, it marks the image's cells as *"do not diff"* so the renderer can never overwrite them. Details and code below.

<details>
<summary>What ratatui-image does (and how it compares)</summary>

### ratatui's approach: `CellDiffOption::Skip`, not flush reordering

ratatui-image never touches flush ordering. When a **cell-based** protocol (iTerm2, Sixel) renders into the buffer, it does two things:

1. The **first cell** of the footprint gets the whole escape-sequence payload as its symbol, tagged `CellDiffOption::ForcedWidth(1)` so the diff always emits it as one unit.
2. **Every other covered cell** gets `CellDiffOption::Skip`:

```rust
// iterm2.rs / sixel.rs — identical pattern
if let Some(cell) = buf.cell_mut(render_area) {
    cell.set_symbol(data).set_diff_option(UNIT_WIDTH); // first cell = payload
}
// Skip entire area (except first cell)
for y in area.top()..area.bottom() {
    for x in area.left()..area.right() {
        if x == left && y == top { continue; }
        cell.set_diff_option(CellDiffOption::Skip);     // <- the key
    }
}
```

`CellDiffOption::Skip` tells the diff renderer **"never emit anything for these cells."** The diff therefore *cannot* punch a hole in the image — the covered cells are excluded from diffing entirely, rather than composited-over afterwards. The hole bug becomes impossible by construction, not fixed by timing.

### Comparison

|                    | This PR (#379)                                        | ratatui-image                                             |
|--------------------|-------------------------------------------------------|----------------------------------------------------------|
| Strategy           | Buffer raw output, **flush after** the cell diff      | **Exclude** image cells from the diff (`Skip`)           |
| Where it lives     | core render path (`Terminal.draw`)                    | the image widget + a per-cell buffer flag                |
| Kitty              | no-op (separate layer)                                | **Unicode placeholders** — Kitty cells go through the normal diff, no special-casing |
| Risk               | reorders every frame's output globally; must re-verify every protocol (incl. untested Sixel) | localized; the diff engine already honors the flag, protocol-agnostic |

### Implications for us

- **cdhermann's instinct that the core is involved was right**, but ratatui shows the cleaner seam is the **buffer/diff layer**, not flush ordering. Adding a per-cell "skip diff" option to `Cell`/`Buffer` and having cell-based protocols mark their footprint would let us drop per-frame retransmission *and* prevent the hole with one mechanism — and it's protocol-agnostic, which neutralises the "Sixel untested" risk (a skipped cell doesn't care which protocol drew it).
- **Kitty lesson (0.6+, not now):** ratatui doesn't treat Kitty as an ignored separate layer — it uses Unicode placeholder cells (`U+10EEEE` + diacritics) so Kitty images participate in normal cell layout/diffing and move/scroll with the grid. That's why forcing Kitty in iTerm2 misbehaves for us and why positioning is fragile. More robust long-term, but a bigger rewrite — out of scope.

### Suggested direction for 0.6

Prototype a `CellDiffOption::Skip` equivalent in TamboUI's `Cell`/`Buffer` + `Terminal` diff and have iTerm2/Sixel mark their footprint, instead of merging flush-after-diff as-is. Likely the smaller diff, and it removes the need to re-verify each protocol.

Source: `benjajaja/ratatui-image` — `src/protocol/{iterm2,sixel,kitty}.rs`.

</details>

cdhermann:
First, we are sorry for the long silence here, and that it took us so much time to come back to this. Right now most of our time is bound on another effort where we try to help, so we can only work on this in small steps, a little at a time. We did not want to leave your comment without an answer, and we really appreciate that you looked so closely into it.

Thanks for the detailed write-up on how ratatui-image handles this. The `CellDiffOption::Skip` idea is clearly the cleaner seam, and we want to try it.

On your earlier question (does #363 replace #379, or do they combine?): here is how we see it.

- **#363 (memory fix) as the base.** One thing we want to check with you: in TamboUI the image bytes go to the terminal through a separate raw stream, not through the cells. Because of that, the diff layer does not really know when to re-send an image, so we think the cache with the generation counter is still useful even with `Skip`. If we are missing something and `Skip` can cover that too, we are happy to learn.
- **#379 (the hole fix) as the place for `Skip`.** Instead of the flush-after-diff commit, we would add a `Cell.SKIP` to `Cell`/`Buffer`, have the diff never emit those cells, and let iTerm2 and Sixel mark their footprint with it. Then the diff can no longer punch a hole, and it works for any protocol. The flush reordering goes away.

So our idea is to keep #363 and #379 as two stacked PRs: #363 the memory cache, #379 the `Skip` mechanism on top.

We also want to be honest about Sixel: you are right that it is barely tested. On our machines auto-detection never picks Sixel (iTerm2 uses its own inline protocol, Ghostty uses Kitty), so we only ever saw Sixel by forcing the protocol in iTerm2. We never ran it on a native Sixel terminal like foot, xterm or mlterm. This is one more reason we like `Skip`: it is protocol-agnostic, so a skipped cell does not depend on which protocol drew it.

One part we still need to think through is `cleanupRawOutput`. Today it writes spaces over vacated image areas before the diff. With `Skip` the diff redraws those cells itself, but the pixel layer (Sixel, iTerm2) still has to be cleared, and terminals differ here. We will test this across iTerm2, Ghostty and, this time, a real Sixel terminal before we call it ready.

One question before we start: would you prefer we rework #379 in place, or open a fresh PR for the `Skip` approach and close #379? Both are fine for us.

Moving both to 0.6 makes sense, thank you for sorting that out. They are rendering correctness, not backend selection, so they should not block 0.5.

PR #383, opened 2026-06-18
fix: propagate Builder.bindings() to TuiRunner input reader

### LINKED ISSUE #382: ToolkitRunner.Builder.bindings() does not propagate to TerminalInputReader — custom bindings silently ignored
I kept hitting this building `jbang deps search` — every time I (or an AI agent working on the code) tried to use custom keybindings via `ToolkitRunner.builder().bindings(custom)`, they didn't work. We'd end up falling back to low-level `event.code() == KeyCode.F5` checks to get anything working, which defeats the purpose of the bindings system entirely. Tab specifically never worked no matter what we tried — unbinding `focusNext` had no effect. It took disassembling the framework bytecode to figure out why.

The pattern was always the same: write the clean version using `.bindings()` and `event.matches()`, watch it fail, then replace it with raw `KeyCode` checks. This happened multiple times across iterations before I traced it to the root cause.

## What happens

`ToolkitRunner.builder().bindings(custom).build()` only applies custom bindings to the render context. The `TuiRunner` (and its `TerminalInputReader`) is created from the Builder's `TuiConfig` field which retains **default** bindings. Every `KeyEvent` parsed by the input reader is stamped with default bindings, so all semantic methods (`isFocusNext()`, `isQuit()`, `isUp()`, `matches()`, `action()`, etc.) behave as if the user never customized bindings.

## Impact — both unbinding and adding are broken

| What the user does | Expected | Actual |
|---|---|---|
| `.unbind(FOCUS_NEXT)` — use Tab for custom action | `isFocusNext()` → `false`, Tab reaches handler | `isFocusNext()` → `true`, Tab silently swallowed by EventRouter |
| `.bind(F5, "myAction")` — custom action | `event.matches("myAction")` → `true` | `event.matches("myAction")` → `false` |
| `.bindings(BindingSets.vim())` — vim navigation | `j`/`k` trigger `isDown()`/`isUp()` | `j`/`k` typed as characters |

## Root cause in `ToolkitRunner.Builder.build()`

```java
public ToolkitRunner build() throws Exception {
    // BUG: config has default bindings, not the custom ones from .bindings()
    TuiRunner tuiRunner = TuiRunner.create(config);
    ToolkitRunner runner = new ToolkitRunner(...);
    
    // This only helps the render context, not the input reader
    runner.renderContext.setBindings(bindings);
    ...
}
```

The `TuiRunner` creates its `TerminalInputReader` with `config.bindings()` (defaults). The `TerminalInputReader` passes those bindings to `EventParser.readEvent()`, which stamps every `KeyEvent` via `KeyEvent.ofKey(code, bindings)`. The `EventRouter.routeKeyEvent()` then calls `event.isFocusNext()` (which checks the **event's** bindings) before forwarding to element handlers — so Tab is intercepted even though the user unbound it.

## Reproducer

```java
Bindings custom = BindingSets.standard().toBuilder()
    .unbind(Actions.FOCUS_NEXT)
    .bind(KeyTrigger.key(KeyCode.TAB), "searchCentral")
    .bind(KeyTrigger.key(KeyCode.F5), "searchCentral")
    .build();

// With 0.3.0: Tab never reaches the handler, F5 reaches it but
// event.matches("searchCentral") returns false
try (var runner = ToolkitRunner.builder().bindings(custom).build()) {
    runner.run(() ->
        panel("test").id("main").focusable()
            .onKeyEvent(event -> {
                if (event.matches("searchCentral")) {
                    // Never called for Tab (swallowed by EventRouter)
                    // Never called for F5 (matches() returns false)
                    doSearch();
                    return EventResult.HANDLED;
                }
                return EventResult.UNHANDLED;
            })
    );
}
```

maxandersen:
Fixes #382

`ToolkitRunner.Builder.bindings()` only set bindings on the render context but
passed the default `TuiConfig` (with default bindings) to `TuiRunner.create()`.
The `TerminalInputReader` stamped every `KeyEvent` with default bindings, so
custom bindings were silently ignored:

- **Unbinding** didn't work: `unbind(FOCUS_NEXT)` had no effect — Tab was still
  intercepted by the EventRouter for focus cycling
- **Adding** didn't work: `bind(F5, "myAction")` was invisible —
  `event.matches("myAction")` returned false
- **Alternate binding sets** didn't work: `bindings(BindingSets.vim())` was
  ignored — `j`/`k` were typed as characters instead of navigating

## Fix

- Add `TuiConfig.withBindings(Bindings)` — returns a copy with different bindings,
  preserving all other config settings
- In `Builder.build()`, call `config.withBindings(bindings)` before creating the
  `TuiRunner`, ensuring the `TerminalInputReader` stamps events with the correct
  bindings

## Tests (9 new)

**`TuiConfigTest`** (3):
- `withBindings` returns new config with different bindings
- `withBindings` preserves custom config settings
- `withBindings` with null falls back to defaults

**`ToolkitRunnerBuilderBindingsTest`** (6):
- Tab reaches element handler when focusNext is unbound
- Tab is intercepted when event has default bindings (demonstrates the bug)
- F5 always reaches element handler (baseline)
- Added custom binding is recognized via `event.matches()`
- Added binding reaches element handler via `event.matches()` (end-to-end through EventRouter)
- Custom bindings are self-consistent when applied to KeyEvents

### DIFF
--- a/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
+++ b/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
@@ -16,7 +16,6 @@ import dev.tamboui.toolkit.element.Element;
 import dev.tamboui.toolkit.element.RenderContext;
 import dev.tamboui.toolkit.element.Size;
 import dev.tamboui.toolkit.event.EventResult;
-import dev.tamboui.tui.TuiConfig;
 import dev.tamboui.tui.event.KeyEvent;
 
 import static dev.tamboui.toolkit.Toolkit.*;
@@ -42,12 +41,8 @@ public class ErrorHandlingDemo {
      * @throws Exception on unexpected error
      */
     public static void main(String[] args) throws Exception {
-        TuiConfig config = TuiConfig.builder()
-                .tickRate(Duration.ofMillis(100))
-                .build();
-
         try (ToolkitRunner runner = ToolkitRunner.builder()
-                .config(config)
+                .tickRate(Duration.ofMillis(100))
                 .faultTolerant(true)
                 .build()) {
 
--- a/demos/fakodex-demo/src/main/java/dev/tamboui/demo/coding/CodingAssistantDemo.java
+++ b/demos/fakodex-demo/src/main/java/dev/tamboui/demo/coding/CodingAssistantDemo.java
@@ -22,7 +22,6 @@ import dev.tamboui.toolkit.app.ToolkitRunner;
 import dev.tamboui.toolkit.element.Element;
 import dev.tamboui.toolkit.element.StyledElement;
 import dev.tamboui.toolkit.event.EventResult;
-import dev.tamboui.tui.TuiConfig;
 import dev.tamboui.tui.bindings.Actions;
 import dev.tamboui.tui.event.KeyEvent;
 import dev.tamboui.widgets.input.TextInputState;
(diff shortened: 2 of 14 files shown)


melix (approved):
Good catch, LGTM! My memory is failing me, but the inline runner may have the same issue

maxandersen:
ok, same bug in inlinetuirunner but in slightly different form - working on clean/aligned approach.

maxandersen:
When checking the bindings bug (https://github.com/tamboui/tamboui/issues/382) I also noticed the
builder/config APIs were inconsistent between Tui and InlineTui, so
fixing that here too so the fix could be the same on both sides.

give it a look @melix !

maxandersen:
merging as the new commits are just same fix done for inlinetui + align api approach.

PR #384, opened 2026-06-18
fix: propagate bindings to render context and align Builder APIs

maxandersen:
## Problem

When a user called `ToolkitRunner.builder().bindings(custom).build()`, custom bindings were set on the render context but **not** propagated to the `TuiConfig` passed to `TuiRunner.create()`. The `TerminalInputReader` stamped `KeyEvent`s with default bindings, so:

- `event.isFocusNext()` returned `true` for Tab even when the user explicitly unbound `focusNext`
- The `EventRouter` intercepted Tab for focus cycling and never forwarded it to element handlers
- Custom bindings added via `.bind()` were never recognized by `event.matches()`

## Fix

The `bindings()` method on both `ToolkitRunner.Builder` and `InlineToolkitRunner.Builder` now applies bindings directly to the config via `toBuilder()`, consistent with all other convenience methods (`mouseCapture()`, `tickRate()`, `noTick()`, `bracketedPaste()`, etc.). All builder methods use the same last-write-wins semantics — no special precedence rules.

Both runners' constructors and `create(config)` factory methods now propagate `config.bindings()` to the render context.

## Changes

- **`TuiConfig` / `InlineTuiConfig`**: Added `toBuilder()` for creating modified copies
- **`ToolkitRunner.Builder`**: `bindings()` applies to config immediately; new convenience methods (`mouseCapture`, `tickRate`, `noTick`, `bracketedPaste`)
- **`InlineToolkitRunner.Builder`**: Same treatment; added `config()` method; convenience methods delegate to `toBuilder()`
- **7 demos**: Simplified to use new builder convenience methods
- **Tests**: 650+ lines covering bindings propagation, EventRouter routing, and config copy semantics

### DIFF
--- a/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
+++ b/demos/error-handling-demo/src/main/java/dev/tamboui/demo/errorhandling/ErrorHandlingDemo.java
@@ -16,7 +16,6 @@ import dev.tamboui.toolkit.element.Element;
 import dev.tamboui.toolkit.element.RenderContext;
 import dev.tamboui.toolkit.element.Size;
 import dev.tamboui.toolkit.event.EventResult;
-import dev.tamboui.tui.TuiConfig;
 import dev.tamboui.tui.event.KeyEvent;
 
 import static dev.tamboui.toolkit.Toolkit.*;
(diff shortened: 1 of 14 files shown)


maxandersen:
The last commit simplifies the Builder pattern for consistency. Previously, `bindings()` used a separate field with a special "bindings always takes precedence" merge in `build()`, while all other convenience methods (`mouseCapture()`, `tickRate()`, etc.) applied immediately to the config via `toBuilder()`. This created an inconsistency:

```java
// mouseCapture is LOST — config replacement wipes it
builder.mouseCapture(true).config(newConfig).build()

// bindings SURVIVED — separate field bypassed config replacement
builder.bindings(custom).config(newConfig).build()
```

Now all builder methods behave the same way: apply immediately to the config, last write wins. The `withBindings()` methods on `TuiConfig`/`InlineTuiConfig` were only needed for the late-merge pattern and are removed — `toBuilder()` already provides that capability.

maxandersen:
replaced with #383

PR #385, opened 2026-06-18
refactor: simplify Builder bindings to use consistent last-write-wins semantics

maxandersen:
Remove the separate `bindings` field from `ToolkitRunner.Builder` and `InlineToolkitRunner.Builder`. The `bindings()` method now applies immediately to the config via `toBuilder()`, consistent with all other convenience methods (`mouseCapture`, `tickRate`, `noTick`, etc.).

This eliminates the `withBindings()` methods on `TuiConfig` and `InlineTuiConfig` which only existed to support the late-merge pattern in `build()`. The `toBuilder()` method already provides the same capability.

Follow-up to #383. Supersedes #384 (which can be closed).


maxandersen:
missed that withBindings() wasn't removed in #383/#384

PR #388, opened 2026-06-24
fix: render emoji presentation sequences (base + VS16) as 2-wide

ammachado:
## Problem

Text-default glyphs such as `U+2328` (⌨ keyboard), `U+23F9` (⏹ stop) and `U+23FA` (⏺ record) are 1-wide on their own but render as **2-wide emoji** when followed by Variation Selector-16 (`U+FE0F`), i.e. as an emoji presentation sequence (`⌨️`, `⏹️`, `⏺️`).

Two places treated these as width 1:

- `CharWidth.of(String)` summed `width(base=1) + width(VS16=0) = 1`.
- `Buffer.setString` placed the base glyph in a single cell and appended VS16 to it, never reserving a continuation cell.

Because the terminal actually advances 2 columns for the glyph while the buffer tracked only 1, everything after the emoji drifted one column out of sync. On redraw, the diff stopped one cell short and left **stale characters on screen**. This is visible, for example, when arrow-key scrolling over list popup items that begin with such an emoji: the row appears to shift left and leftover letters remain after the selection moves away.

## Fix

- `CharWidth.of(String)`: a 1-wide base glyph immediately followed by `U+FE0F` now counts as width **2**.
- `Buffer.setString`: when `U+FE0F` is appended to a base that was placed as 1-wide, promote it to occupy two cells by reserving a continuation cell.

Already-wide emoji that also carry VS16 (e.g. `U+1F5A5` 🖥️ desktop) are unaffected: their base is already width 2, so no extra promotion happens.

## Tests

- `CharWidthTest`: emoji presentation sequences (`⌨️`, `⏹️`, `⏺️`, and mixed `A⌨️B`) measure width 2; already-wide `🖥️` stays width 2.
- `BufferWideCharTest`: `⌨️` occupies two cells with a continuation; `⏺️ Rec` keeps following text correctly positioned; `🖥️` stays two cells; an emoji presentation sequence placed on the last column degrades to its 1-wide base (right-edge truncation, matching wide-char behavior).

The `of(String)` / `setString` Javadoc grapheme-handling lists were also updated to document the new emoji-presentation-sequence case.

Full `:tamboui-core:test` and `:tamboui-widgets:test` suites pass; `spotlessApply` is a no-op on the change.

## Follow-up

Review surfaced a pre-existing, unrelated no-op block in `Buffer.setString` (wide-char overwrite handling). Tracked separately in #389 to keep this PR focused.

---
_Submitted by Claude Code on behalf of Adriano Machado (@ammachado), who has reviewed and takes responsibility for this change._

### DIFF
--- a/demos/unicode-demo/src/main/java/dev/tamboui/demo/UnicodeDemo.java
+++ b/demos/unicode-demo/src/main/java/dev/tamboui/demo/UnicodeDemo.java
@@ -25,6 +25,9 @@ import dev.tamboui.text.Text;
 import dev.tamboui.widgets.block.Block;
 import dev.tamboui.widgets.block.BorderType;
 import dev.tamboui.widgets.block.Borders;
+import dev.tamboui.widgets.list.ListItem;
+import dev.tamboui.widgets.list.ListState;
+import dev.tamboui.widgets.list.ListWidget;
 import dev.tamboui.widgets.paragraph.Paragraph;
 
 /**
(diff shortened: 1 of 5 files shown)


davsclaus (approved):

davsclaus:
https://issues.apache.org/jira/browse/CAMEL-23818

maxandersen:
Verified the earlier review concerns are addressed in the current head (VS16 promotion gated on the Unicode emoji-variation-sequences table across `CharWidth.of`, `Buffer.setString`, and both truncation helpers; `truncateWithEllipsis` delegates to those; right-edge and stale-continuation cases tested).

Pushed one commit on top:
- tests pinning ZWJ sequences that embed variation bases (👩‍⚕️, ⛹️‍♀️) stay width 2 / one cell — the `baseCol == col - 1` guard prevents double-promotion, now locked in
- `unicode-demo` gained an interactive VS16 list (⏺️/⏹️/⌨️/… with j/k selection) reproducing the original stale-cells symptom: run it from main to see the bug, from this branch to see it fixed

The orphaned-continuation case when promoting over a wide-char *start* behaves the same as plain wide chars — correctly out of scope here, tracked in #389. LGTM from my side.

PR #390, opened 2026-06-25
fix: deliver mouse hover (MOVE) events

ammachado:
TamboUI never delivered `MouseEventKind.MOVE` events. Two independent root causes:

1. **Backend never enabled any-event tracking.** `JLineBackend` enabled DECSET modes 1000/1002/1015/1006 but not **1003**, so the terminal never reported bare mouse motion. Added an opt-in `mouseMotion` flag on `TuiConfig`, threaded through `TuiRunner` to a new `Backend.enableMouseCapture(boolean)` overload. JLine emits `?1003h` only when requested and always clears `?1003l` on teardown. Off by default, since motion fires an event for every cell the cursor crosses.
2. **Parser made `MOVE` unreachable.** `EventParser` classified any motion as `DRAG` before the no-button check, so the `MOVE` branch was dead code for real terminal input. It now reports motion-with-no-button as `MOVE` and motion-with-a-button-held as `DRAG`.

Both fixes are required: bug 1 makes the terminal send hover bytes, bug 2 makes the parser label them `MOVE`.

## Tests
- `EventParserTest`: SGR `MOVE` (button 35), `DRAG` (button 32, left held), and `PRESS` (button 0) parsing.
- `TuiConfigTest`: `mouseMotion` default-off, builder enable, and `toBuilder` round-trip.

## Downstream
Enables hover support in Apache Camel's `camel-tui`, which already keys its hover UI off `MouseEventKind.MOVE`: [CAMEL-23831](https://issues.apache.org/jira/projects/CAMEL/issues/CAMEL-23831). Once this lands in a release, Camel bumps `<tamboui-version>` and enables `mouseMotion(true)` alongside `mouseCapture(true)`.

## Note
`mouseMotion` currently affects only the JLine backend; the Panama and Aesh backends ignore the flag (they fall back to plain mouse capture via the no-arg method). A follow-up can extend them if desired.


davsclaus (approved):

maxandersen:
> the Panama and Aesh backends ignore the flag (they fall back to plain mouse capture via the no-arg method). A follow-up can extend them if desired.

I dont like knowingly release with "weaker" backends when/if it can be avoided. lets fix this before pushing in - unless you have good reason why we couldn't do it uniformly?

maxandersen:
fwiw even when i enable moion in tui-demo it never show the event arriviing so i can't with good concious  merge this yet.

maxandersen:
> fwiw even when i enable moion in tui-demo it never show the event arriviing so i can't with good concious merge this yet.

...and reason was that I was running java 25 so it siliently ignore request for motion.

we need it working across the backends.

PR #391, opened 2026-06-27
fix: avoid mistaking an absent backend for a broken one

ammachado:
## Summary

Backend discovery previously reported a present-but-broken backend as if no backend were on the classpath, sending users down the wrong path (`Add a backend dependency...`) when the real problem was a provider that failed to initialize (e.g. a JLine binary incompatibility). This PR makes discovery both more robust and more honest about failures.

- `SafeServiceLoader.load` now consults a de-duplicated, ordered set of candidate classloaders (thread-context, own, system) when no explicit loader is given, so providers bundled in fat-jars, containers, and app servers are found even when the thread-context classloader cannot see them. Providers and failures are de-duplicated across candidate loaders.
- Providers that fail to instantiate (`LinkageError` / `ServiceConfigurationError`) are now captured and each distinct failure is forwarded once, so a broken backend is reported with its real cause instead of the misleading "add a dependency" advice.
- An explicit classloader can be supplied for deterministic discovery via `BackendFactory.create(ClassLoader)`, `TuiConfig.backendClassLoader(...)`, and `TuiRunner.create(config, ClassLoader)`.
- `resolveProviders` only blames broken providers when there are no usable ones at all, so an unrelated failure no longer claims that "none could be initialized" when a working provider was actually discovered.

## Tests

- `SafeServiceLoaderTest` — failure capture, exclusive explicit loader, blind-TCCL fallback, cross-loader de-duplication.
- `BackendFactoryTest` — absent vs. broken provider messaging, user-selected-but-broken path, fallback discovery.
- `TuiConfigTest` / `TuiRunnerBackendClassLoaderTest` — `backendClassLoader` plumbing and precedence.

`./gradlew :tamboui-core:test` (BackendFactoryTest, SafeServiceLoaderTest) passes.


davsclaus (approved):

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -71,24 +71,44 @@ public static void initRecording() {
...
+     * @throws BackendException if no provider is found or all providers fail
      */
     public static Backend create() throws IOException {
+        return create((ClassLoader) null);
+    }
+
+    /**
+     * Creates a new backend instance, discovering providers using the given classloader.
+     * <p>
+     * Behaves like {@link #create()}, except that backend discovery uses {@code classLoader}
```
gnodet, line 85: **Style (low):** `new java.util.ArrayList<>()` uses a fully qualified name. AGENTS.md: _"You MUST use imports instead of fully qualified names in code."_ This follows the pre-existing pattern on line 129 (`resolveProviders`), but both instances should ideally use the import — `ArrayList` is already transitively available.

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -161,6 +194,24 @@ private static Backend tryProviders(List<BackendProvider> providers) {
...
+                .map(t -> "  " + t.getMessage())
+                .collect(Collectors.joining("\n"));
+        return new BackendException(
+                "Found " + loadFailures.size() + " BackendProvider(s) on the classpath but none "
+                        + "could be initialized (see cause).\nFailures:\n" + detail,
+                loadFailures.get(0)
+        );
+    }
+
     /**
```
gnodet, line 226: **Suggestion (low):** Consider attaching subsequent failures as suppressed exceptions so their full stack traces are preserved in logs/diagnostics:

```java
BackendException ex = new BackendException(
        "Found " + loadFailures.size() + " BackendProvider(s) on the classpath but none "
                + "could be initialized (see cause).\nFailures:\n" + detail,
        loadFailures.get(0)
);
for (int i = 1; i < loadFailures.size(); i++) {
    ex.addSuppressed(loadFailures.get(i));
}
return ex;
```

Currently, only `loadFailures.get(0)` is chained as the cause — the rest only appear as one-line summaries, losing their stack traces.

tamboui-core/src/main/java/dev/tamboui/terminal/BackendFactory.java
```diff
@@ -117,11 +139,18 @@ private static List<BackendProvider> resolveProviders(String providerSpec, List<
...
+                    // A requested provider can be missing because it is absent, or because it is
+                    // present on the classpath but failed to initialize. Only blame broken
+                    // providers when there are no usable ones at all; otherwise an unrelated
+                    // failure would wrongly claim that "none could be initialized" even though a
+                    // working provider was discovered (the requested name is simply absent).
+                    .orElseThrow(() -> (!loadFailures.isEmpty() && allProviders.isEmpty())
+                            ? brokenProvidersException(loadFailures)
+                            : new BackendException(
+                                    "No BackendProvider found on classpath for provider name" +
+                                            " '" + trimmedSpec + "'.\n" +
```
gnodet, line 157: **Note (low):** The heuristic `!loadFailures.isEmpty() && allProviders.isEmpty()` is reasonable — it avoids blaming unrelated broken providers when working ones exist. Worth noting there's a narrow edge case: if the user requests `jline` and `jline` is on the classpath but broken while `panama` works, `allProviders` is non-empty so the message says "not found" rather than "broken". Acceptable given `ServiceLoader`'s error model, but a brief inline comment documenting this trade-off would help future maintainers.

gnodet (commented):
## Review: `fix: avoid mistaking an absent backend for a broken one`

Nice work improving the backend discovery diagnostics — the core problem (a broken provider reported as absent) is genuinely confusing for users, and the fix is well-structured. The multi-classloader fallback in `SafeServiceLoader` is a meaningful robustness improvement for fat-jar and container deployments.

### Summary

| Category | Status |
|---|---|
| Correctness | ✅ Core logic is correct |
| Backward compat | ✅ Old APIs delegate to new overloads with `null` |
| Tests | ✅ Thorough — broken, absent, blind-TCCL, dedup |
| Scope | ✅ Focused on the diagnostic gap |
| Conventions | ⚠️ A few minor style violations (see findings) |
| Regressions | ✅ None expected |

### Confirmed Issues

**1. (Low) Fully qualified class names instead of imports — AGENTS.md violation**

AGENTS.md: _"You MUST use imports instead of fully qualified names in code."_

- `BackendFactory.java` new line: `new java.util.ArrayList<>()` (follows pre-existing pattern in the file, but the PR adds a new instance)
- `TuiConfigTest.java`: `new java.net.URLClassLoader(new java.net.URL[0], ...)`
- `TuiRunnerBackendClassLoaderTest.java`: `new java.net.URL[0]`

**2. (Low) Missing trailing newline in `SafeServiceLoader.java`**

The diff ends with `\ No newline at end of file`. Minor, but most style checkers flag this.

### Questions / Suggestions

**3. (Low) `brokenProvidersException` — subsequent failures lose stack traces**

Only the first failure is set as the `cause`. The others appear as one-line summaries in the detail string, losing their stack traces. Consider attaching them via `addSuppressed()` for better diagnostics when multiple providers break:

```java
BackendException ex = new BackendException("...", loadFailures.get(0));
for (int i = 1; i < loadFailures.size(); i++) {
    ex.addSuppressed(loadFailures.get(i));
}
return ex;
```

**4. (Low) `resolveProviders` heuristic edge case**

The condition `!loadFailures.isEmpty() && allProviders.isEmpty()` correctly avoids blaming unrelated broken providers when a working provider exists. However, there's a narrow edge case: if the user requests `tamboui.backend=jline`, and `jline` is on the classpath but broken (fails to instantiate) while `panama` works, the message would say _"No BackendProvider found… Add a dependency"_ — misleading because `jline` IS present, just broken. This is an acceptable limitation given `ServiceLoader` doesn't report which class failed, but a brief comment in the code noting this trade-off would help future maintainers.

> **Note:** This review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

ammachado:
@gnodet @davsclaus I've pushed f9af2008 addressing the review feedback:

- FQN → imports (`ArrayList` in `BackendFactory`, `URL`/`URLClassLoader` in the tests) per AGENTS.md
- `brokenProvidersException` now attaches subsequent failures via `addSuppressed()` so their stack traces are preserved
- Documented the `resolveProviders` "requested-but-broken while another works" trade-off inline
- Added the missing trailing newline in `SafeServiceLoader.java`

`:tamboui-core:test` and `:tamboui-tui:test` (BackendFactoryTest, SafeServiceLoaderTest, TuiConfigTest, TuiRunnerBackendClassLoaderTest) pass. Would appreciate another look when you have a moment.

davsclaus (approved):

gnodet (approved):
## Follow-up review after `16287bce` — all feedback addressed ✅

The new commit cleanly addresses all four findings from the previous review:

| # | Finding | Resolution |
|---|---|---|
| 1 | Fully qualified class names | ✅ Added `import java.util.ArrayList` in `BackendFactory.java`; added `import java.net.URL` / `URLClassLoader` in both test files. Also fixed the **pre-existing** FQN in `resolveProviders`. |
| 2 | Missing trailing newline in `SafeServiceLoader.java` | ✅ Fixed |
| 3 | `brokenProvidersException` — subsequent failures lose stack traces | ✅ Added `addSuppressed()` loop for failures at index 1+; updated javadoc to document the behavior |
| 4 | `resolveProviders` heuristic edge case | ✅ Added 4-line inline comment documenting the trade-off (ServiceLoader can't distinguish which class failed) |

Changes are minimal and mechanical — no new logic introduced. LGTM.

> **Note:** This review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

maxandersen:
Am I missing something or is there more changes in this than needed for the backend loading code improvements ?

ammachado:
Fixed. The branch was based on my fork main, so it included two unrelated commits (from PR #388) before the backend loading changes. I rebased it onto current upstream main and force-pushed with lease. The PR now shows only the two backend-discovery commits.

Local verification after the rebase:

```
./gradlew -q test
./gradlew -q build
./gradlew -q javadoc
```

maxandersen:
sorry - bad noise. rebasing this with #422 and a bit simplification. thanks for this and sorry for the delay!

maxandersen:
Rebased onto main (over the merged #422) preserving the original commit, and added two commits per the checklist above:

- `279fe134` — requested-but-broken providers now report their real cause with a fallback hint, and unknown-name errors disclose dropped providers; verified on a real Java 21 runtime with the Java-22-compiled panama backend
- `1923d86b` — trimmed the redundant `TuiRunner.create(config, ClassLoader)` overload; `TuiConfig.backendClassLoader()` is the single opt-in

All core/tui/toolkit tests and javadoc green locally. Follow-ups (JDK-matrix integration tests, docs, `backend.selected` report) tracked in #429.

PR #393, opened 2026-06-29
fix: clamp MarkdownView scroll to prevent blank viewport

### LINKED ISSUE #392: MarkdownView scroll clamp allows scrolling past content end
## Description

`MarkdownView` clamps the scroll value to `totalRows - 1`, which allows scrolling well past the point where content is visible. When `scroll >= totalRows - viewportHeight + 1`, the bottom of the content scrolls above the viewport, leaving blank space below.

## Expected Behavior

The scroll should be clamped to `Math.max(0, totalRows - contentArea.height())` so the furthest scroll position shows the last page of content aligned to the bottom of the viewport — the same behavior as text editors and browsers.

## Current Behavior

In `MarkdownView.java` line 243:
```java
int skip = Math.max(0, Math.min(scroll, totalRows - 1));
```

This allows `skip` values up to `totalRows - 1`, which means only the very last row of content is visible at the top of the viewport, with the rest being blank.

## Proposed Fix

Change the clamp to:
```java
int skip = Math.max(0, Math.min(scroll, Math.max(0, totalRows - contentArea.height())));
```

This ensures the content bottom never scrolls above the viewport bottom, while still allowing full scrolling through all content.

## Use Case

This is encountered when building scrollable chat/conversation UIs with `MarkdownView` where auto-scroll-to-bottom needs to pass a large scroll value. The current clamp makes it impossible to reliably scroll to the bottom without knowing the exact rendered line count in advance.

davsclaus:
## Summary

- Fix scroll clamp in `MarkdownView` to prevent scrolling past the content bottom
- Change clamp from `totalRows - 1` to `totalRows - contentArea.height()` so the last page of content stays visible
- Add tests verifying scroll clamp behavior including edge cases

Fixes #392

## Details

The `MarkdownView` scroll was clamped to `totalRows - 1`, which allows the content to scroll entirely above the viewport — leaving a blank screen. The fix clamps to `Math.max(0, totalRows - contentArea.height())` instead, so the furthest scroll position shows the last page of content aligned to the bottom of the viewport.

This is the same behavior as text editors and browsers — you can never scroll past the point where the bottom of content meets the bottom of the viewport.

## Test plan

- [x] Added unit test `scrollClampPreventsBlankScreen` with `BufferAssertions` verifying exact content rows shown
- [x] Added edge-case test `scrollClampContentShorterThanViewport` (content < viewport → no scroll)
- [x] Added edge-case test `scrollClampContentExactlyFitsViewport` (content == viewport → no scroll)
- [x] Fixed pre-existing tests (`rendersCodeBlockScrolledFromTop`, `rendersCodeBlockBottomBorderOnly`) that relied on the old buggy scroll behavior — adjusted viewport heights so scrolling is still possible
- [x] All 27 MarkdownViewTest tests pass

### DIFF
--- a/tamboui-markdown/src/test/java/dev/tamboui/markdown/MarkdownViewTest.java
+++ b/tamboui-markdown/src/test/java/dev/tamboui/markdown/MarkdownViewTest.java
@@ -329,4 +329,34 @@ class MarkdownViewTest {
             renderInto(view, 30, 6);
         }
     }
+
+    @Test
+    @DisplayName("scroll clamp prevents scrolling past content bottom")
+    void scrollClampPreventsBlankScreen() {
+        // Content renders to a few rows. A very large scroll value should show the
+        // last page of content, not a blank viewport.
+        String source = "line one\n\nline two\n\nline three";
+        int viewportHeight = 3;
+
+        // Render with a huge scroll — should show the bottom of the content
+        MarkdownView scrolled = MarkdownView.builder()
+            .source(source)
+            .scroll(Integer.MAX_VALUE)
+            .build();
+        Buffer scrolledBuf = renderInto(scrolled, 30, viewportHeight);
+
+        // At least one row in the viewport must contain non-space content
+        boolean hasContent = false;
+        for (int y = 0; y < viewportHeight && !hasContent; y++) {
+            for (int x = 0; x < 30; x++) {
+                if (!scrolledBuf.get(x, y).symbol().isBlank()) {
+                    hasContent = true;
+                    break;
+                }
+            }
+        }
+        assertThat(hasContent)
+            .as("viewport should show content, not blank space, when scroll is very large")
+            .isTrue();
+    }
 }
--- a/tamboui-markdown/src/test/java/dev/tamboui/markdown/MarkdownViewTest.java
+++ b/tamboui-markdown/src/test/java/dev/tamboui/markdown/MarkdownViewTest.java
@@ -7,6 +7,7 @@ package dev.tamboui.markdown;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
+import dev.tamboui.assertj.BufferAssertions;
 import dev.tamboui.buffer.Buffer;
 import dev.tamboui.layout.Rect;
 import dev.tamboui.style.Modifier;
@@ -258,12 +259,13 @@ class MarkdownViewTest {
     @DisplayName("renders bottom of a code block when top is scrolled off-screen")
     void rendersCodeBlockScrolledFromTop() {
         // 1-line code block = 3 rows (top border + content + bottom border).
+        // Viewport must be smaller than content so scrolling is possible.
         // Scroll by 1 so the top border is off-screen — should show content + bottom border.
         MarkdownView view = MarkdownView.builder()
             .source("```\ncode\n```")
             .scroll(1)
             .build();
-        Buffer buffer = renderInto(view, 20, 3);
+        Buffer buffer = renderInto(view, 20, 2);
 
         // Row 0: code content (top border scrolled away)
         assertThat(buffer.get(1, 0).symbol()).isEqualTo("c");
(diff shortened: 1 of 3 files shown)


tamboui-markdown/src/test/java/dev/tamboui/markdown/MarkdownViewTest.java
```diff
@@ -329,4 +329,34 @@ void survivesStreaming() {
...
+        for (int y = 0; y < viewportHeight && !hasContent; y++) {
+            for (int x = 0; x < 30; x++) {
+                if (!scrolledBuf.get(x, y).symbol().isBlank()) {
+                    hasContent = true;
+                    break;
+                }
+            }
+        }
+        assertThat(hasContent)
+            .as("viewport should show content, not blank space, when scroll is very large")
```
gnodet: **Suggestion (medium):** This assertion checks that *something* is visible, but not that the *correct* content is shown. It would pass even if the wrong rows were displayed.

A stronger assertion would verify the expected rows. With 5 content rows and a 3-row viewport, the clamp should be `max(0, 5-3) = 2`, showing rows 2–4 ("line two", blank, "line three").

Consider using `BufferAssertions` per project conventions (AGENTS.md: tests SHOULD use `BufferAssertions` test fixtures). For example:

```suggestion
        BufferAssert.assertThat(scrolledBuf).row(0).isEqualTo("line two");
```

(Exact API may differ — adapt to whatever `BufferAssertions` provides. The key point is asserting *what* content appears, not just *that* content appears.)

gnodet (commented):
## Review: `fix: clamp MarkdownView scroll to prevent blank viewport`

The production fix is **correct and well-scoped**. The formula `Math.max(0, Math.min(scroll, Math.max(0, totalRows - contentArea.height())))` properly handles all edge cases:

| Scenario | Result |
|---|---|
| `totalRows == 0` | Early return guards (safe) |
| `totalRows < viewport` | `skip = 0` ✅ |
| `totalRows == viewport` | `skip = 0` ✅ |
| `totalRows > viewport` | Max skip = `totalRows - height` ✅ |

No regression risk — the change strictly reduces the maximum `skip` value, and the existing `scrollsContent` test remains valid.

### Suggestion: edge-case tests

The test covers the "scroll way past end" case well. Two additional cases would strengthen confidence in the inner `Math.max(0, ...)` guard:
- Content **shorter** than viewport (`totalRows < height`) — verify `skip = 0`, all content visible
- Content **exactly fitting** viewport (`totalRows == height`) — verify `skip = 0`, all content visible

> **Note:** This review does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

tamboui-markdown/src/test/java/dev/tamboui/markdown/MarkdownViewTest.java
davsclaus: Thanks for the suggestion! Replaced the generic content-presence loop with `BufferAssertions.assertThat()` assertions that verify the exact rows shown ("line two" at row 0 and "line three" at row 2). Also added two edge-case tests for content shorter than and exactly fitting the viewport.

Additionally fixed two pre-existing tests (`rendersCodeBlockScrolledFromTop` and `rendersCodeBlockBottomBorderOnly`) that were broken by the scroll clamp fix — they used viewports equal to the content height, so the clamp correctly prevented any scrolling. Adjusted them to use smaller viewports.

maxandersen:
Pushed a small follow-up commit: the markdown-demo tracked an unbounded scroll counter, so after over-pressing `j` past the clamp, `k` appeared unresponsive until the counter drifted back below the max. The demo now derives the max from the existing `computeHeight` API and clamps on scroll-down (block chrome cancels out, so it's just `computeHeight(width) - height`). Added a test documenting the derivation.

Deliberately did not add a `maxScroll` convenience API — opened a separate issue to discuss whether interactive scrolling should instead get a self-clamping `MarkdownViewState` like `ListState`/`TextAreaState`.

PR #395, opened 2026-07-01
fix: avoid resetting terminal cursor blink timer on unchanged frames

gnodet:
## Summary

- Track previous cursor position in `Terminal.draw()` and skip redundant `backend.setCursorPosition()` calls when the position hasn't changed and no cells were redrawn
- This lets the terminal's native cursor blink animation run undisturbed between frames
- When cells are redrawn (`hadDiff`), the cursor is always repositioned since `backend.draw()` moves the physical cursor

## Context

Applications with a shell/input panel that call `Frame.setCursorPosition()` every frame were seeing a non-blinking cursor because the cursor position escape sequence was sent every render cycle, resetting the terminal's blink timer before it could complete a cycle.

### DIFF
--- a/tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
+++ b/tamboui-core/src/main/java/dev/tamboui/terminal/Terminal.java
@@ -17,6 +17,7 @@ import dev.tamboui.buffer.Buffer;
 import dev.tamboui.buffer.DiffResult;
 import dev.tamboui.error.RuntimeIOException;
 import dev.tamboui.jfr.TerminalDrawEvent;
+import dev.tamboui.layout.Position;
 import dev.tamboui.layout.Rect;
 import dev.tamboui.layout.Size;
 
@@ -39,6 +40,7 @@ public final class Terminal<B extends Backend> implements AutoCloseable {
     private Buffer currentBuffer;
     private Buffer previousBuffer;
     private boolean hiddenCursor;
+    private Position previousCursorPosition;
     private boolean previousFrameHadRawOutput;
     private List<Rect> previousRawOutputAreas = Collections.emptyList();
 
@@ -149,19 +151,24 @@ public final class Terminal<B extends Backend> implements AutoCloseable {
 
                 // Calculate diff and draw (zero-allocation DoD variant)
                 previousBuffer.diff(currentBuffer, diffResult);
-                if (!diffResult.isEmpty()) {
+                boolean hadDiff = !diffResult.isEmpty();
+                if (hadDiff) {
                     backend.draw(diffResult);
                 }
                 diffResult.clear();  // Clear after use to release Cell refs for GC
 
-                // Handle cursor
+                // Handle cursor — skip redundant setCursorPosition when nothing
+                // on screen changed, so the terminal's blink timer is not reset
                 if (frame.isCursorVisible()) {
                     frame.cursorPosition().ifPresent(pos -> {
                         try {
-                            backend.setCursorPosition(pos);
-                            if (hiddenCursor) {
-                                backend.showCursor();
-                                hiddenCursor = false;
+                            if (hiddenCursor || hadDiff || !pos.equals(previousCursorPosition)) {
+                                backend.setCursorPosition(pos);
+                                if (hiddenCursor) {
+                                    backend.showCursor();
+                                    hiddenCursor = false;
+                                }
+                                previousCursorPosition = pos;
                             }
                         } catch (IOException e) {
                             throw new RuntimeIOException(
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

davsclaus:
this will also help tamboui UI input fields and others where a cursor may be visible. Currently its a bit confusing as there are often no cursor and you type blindling and stare all over the screen to see where the input is being typed

gnodet:
Exactly — this was quite noticeable when testing the scrollbar mouse interaction work (#397). The cursor blink reset was happening every frame even when nothing changed, making input fields feel broken. With this fix, the cursor position escape sequence is only sent when cells actually change or the position moves, so the terminal's native blink animation runs undisturbed.

maxandersen (approved):
Nice catch. I wondered what caused that!

maxandersen:
hmm - can you point to a demo that behaves differently with this code?

maxandersen:
@davsclaus @gnodet i still struggle to see a diff in my testing. got a demo that shows the diff? (note: I grok the idea about the fix but just weirded out i'm not seeing a difference in behavior :)

maxandersen:
Rebased onto main (over #410's synchronized-update restructure) and **added a commit** extending the fix. In one sentence: `Terminal.draw()` now writes nothing at all to the terminal when a frame changes nothing — no cells changed, cursor unmoved, no raw output — instead of sending an empty synchronized-update wrapper and cursor reposition every time, which was resetting the terminal's cursor-blink timer (and wasting a flush syscall) on every idle redraw.

Why the extra commit: I couldn't find a way to reproduce the issue in any existing demo — they all hide the native cursor (`backend.hideCursor()`) and paint a static reversed-cell fake cursor instead, and most modern terminals (Ghostty, iTerm2, Terminal.app) default to a *steady* cursor anyway, so there was no blink to preserve. Only with the very specific demo below could I reproduce it — and it showed that in Ghostty the blink timer resets on **any** output, so skipping just the redundant `setCursorPosition` wasn't enough: the empty Mode 2026 BSU/ESU pair per idle frame (from #410) still suppressed blinking. Measured escapes over 3 idle seconds: 29 CUPs before → 1 after the original commit → **zero bytes** after the added commit.

Repro (`cursor_blink_probe.java`, run with jbang; `[i]`dle = unchanged frames at 10Hz, `[s]`ilent = no draw calls while idle, `[d]`irty = a cell changes each frame):

```java
//DEPS dev.tamboui:tamboui-widgets:0.5.0-SNAPSHOT
//DEPS dev.tamboui:tamboui-jline3-backend:0.5.0-SNAPSHOT

import dev.tamboui.layout.Position;
import dev.tamboui.terminal.Backend;
import dev.tamboui.terminal.BackendFactory;
import dev.tamboui.terminal.Terminal;
import dev.tamboui.widgets.paragraph.Paragraph;

public class cursor_blink_probe {
    public static void main(String[] args) throws Exception {
        try (Backend backend = BackendFactory.create()) {
            backend.enableRawMode();
            backend.enterAlternateScreen();
            backend.writeRaw("\u001b[1 q"); // blinking block: most terminals default to steady
            backend.flush();

            Terminal<Backend> terminal = new Terminal<>(backend);
            char[] mode = {'i'};
            long[] tick = {0};
            boolean[] needsDraw = {true};

            while (true) {
                if (mode[0] != 's' || needsDraw[0]) {
                    terminal.draw(frame -> {
                        String spinner = mode[0] == 'd' ? String.valueOf("|/-\\".charAt((int) (tick[0] % 4))) : " ";
                        Paragraph.builder()
                            .text("mode: " + mode[0] + "   [i]dle [s]ilent [d]irty [q]uit   " + spinner + "\n\n"
                                + "native cursor here > ")
                            .build()
                            .render(frame.area(), frame.buffer());
                        frame.setCursorPosition(new Position(21, 2));
                    });
                    needsDraw[0] = false;
                }
                tick[0]++;
                int c = backend.read(100);
                if (c == 'q' || c == 3) { break; }
                if (c == 'i' || c == 's' || c == 'd') { mode[0] = (char) c; needsDraw[0] = true; }
            }
            backend.writeRaw("\u001b[0 q");
            backend.leaveAlternateScreen();
            backend.disableRawMode();
        }
    }
}
```

Before the added commit: `[i]` solid, `[s]` blinks (proving any output resets blink). After: `[i]` and `[s]` both blink, `[d]` correctly solid.

@gnodet does this approach make sense to you? The skip-everything condition is `hadDiff || repositionCursor || hideCursorNow || rawOutput involved`, decided before opening BSU — I flipped the old `drawSendsSyncUpdateEvenWithNoDiff` test into `drawSkipsAllOutputWhenNothingChanged` and added moved/unchanged/hide cursor cases. A separate follow-up question this surfaced: TamboUI's input widgets hide the native cursor and paint static reversed cells, so users never get a blinking cursor at all (davsclaus's "typing blindly" point) — worth an issue?

maxandersen:
merging this to have in 0.5 and we can improve/adjust from there.

PR #397, opened 2026-07-01
Add mouse interaction support to Scrollbar widget

gnodet:
## Summary

Adds mouse interaction support to the `Scrollbar` widget, enabling click-to-page and drag-to-scroll functionality for both vertical and horizontal scrollbars.

### New files
- **`ScrollbarAction.java`** — Enum of mouse actions (`PRESS`, `DRAG`, `RELEASE`, `SCROLL_UP`, `SCROLL_DOWN`) that consumers map from their mouse events. Includes a Javadoc example showing `MouseEvent` → `ScrollbarAction` mapping.
- **`ScrollbarMouseTest.java`** — 29 unit tests across 7 groups covering all mouse interaction scenarios.

### Modified files
- **`Scrollbar.java`** — Added `handleMouseAction(int mouseX, int mouseY, ScrollbarAction action, Rect area, ScrollbarState state)` public method with:
  - Click on track before thumb → `pageUp()`
  - Click on track after thumb → `pageDown()`
  - Click on thumb → start drag
  - Drag → proportional thumb movement using inverse mapping from render formula
  - Release → end drag (consumed even outside bounds while dragging)
  - Scroll wheel → `prev()` / `next()`
  - Click on begin/end marker arrows → `prev()` / `next()`
  - Private `computeThumbGeometry()` helper — single source of truth for thumb size/position, shared by `render()`, `handlePress()`, and `handleDrag()`
- **`ScrollbarState.java`** — Added drag state tracking (`isDragging()`, package-private `startDrag()`/`endDrag()`/`dragOffset()`).

### Design decisions
- **No dependency on `tamboui-tui`**: Uses raw coordinates + `ScrollbarAction` enum instead of `MouseEvent` to avoid a circular dependency (`tui` → `widgets` → `tui`). Consumers map `MouseEvent` to `ScrollbarAction` at the call site.
- **Consistent thumb geometry**: `computeThumbGeometry()` is the single source of truth used by both rendering and mouse hit-testing, ensuring clicks always target the visually rendered thumb.
- **Drag outside bounds**: Drag events are consumed even when the mouse moves outside the scrollbar area, matching native scrollbar behavior.
- **RELEASE always ends drag**: RELEASE during an active drag is handled at every early-return point (empty area, zero content, small track), preventing a stuck drag after terminal resize.

## Test plan
- [x] 29 unit tests in `ScrollbarMouseTest` covering:
  - Click-to-page (up, down, multiple pages, near-end behavior, thumb-at-end drag)
  - Drag-to-scroll (proportional movement, drag outside bounds, full drag sequence, release without drag)
  - Scroll wheel (up, down)
  - Horizontal orientation (page left/right, horizontal drag)
  - Begin/end markers (click begin arrow, click end arrow, markers affect effective track length)
  - Vertical left orientation
  - Edge cases (empty area, zero content, content fits viewport, out of bounds, drag without press, release when not dragging, small track, position clamping)

### DIFF
--- a/tamboui-widgets/src/main/java/dev/tamboui/widgets/scrollbar/Scrollbar.java
+++ b/tamboui-widgets/src/main/java/dev/tamboui/widgets/scrollbar/Scrollbar.java
@@ -336,26 +336,9 @@ public final class Scrollbar implements StatefulWidget<ScrollbarState> {
         }
 
         // Calculate thumb position and size
-        int contentLength = state.contentLength();
-        int viewportLength = state.viewportContentLength() > 0
-            ? state.viewportContentLength()
-            : trackLength;
-
-        // Thumb size proportional to viewport/content ratio
-        int thumbSize = Math.max(1, (int) Math.ceil((double) viewportLength / contentLength * trackLength));
-        thumbSize = Math.min(thumbSize, trackLength);
-
-        // Thumb position
-        int scrollableRange = contentLength - viewportLength;
-        int thumbPosition;
-        if (scrollableRange <= 0) {
-            thumbPosition = 0;
-        } else {
-            double scrollFraction = (double) state.position() / scrollableRange;
-            scrollFraction = Math.max(0.0, Math.min(1.0, scrollFraction));
-            int thumbRange = trackLength - thumbSize;
-            thumbPosition = (int) Math.round(scrollFraction * thumbRange);
-        }
+        ThumbGeometry thumb = computeThumbGeometry(trackLength, state);
+        int thumbSize = thumb.size;
+        int thumbPosition = thumb.position;
 
         // Get effective styles
         Style effectiveStyle = style != null ? style : Style.EMPTY;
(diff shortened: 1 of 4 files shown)


gnodet:
Related PRs raised during this work:
- #398 — adds horizontal scroll wheel events (`SCROLL_LEFT`/`SCROLL_RIGHT`) to tamboui's mouse event system. Not blocking for this PR, but useful for future horizontal scrollbar support.

davsclaus (approved):

maxandersen:
Deferring this to **0.6** — not because of any problem with the code (it's solid: `computeThumbGeometry()` as the single source of truth for render + hit-test is exactly right, drag-offset handling is correct, teardown-on-release is robust, and the 29 tests are thorough), but because the scroll story is better shipped as one coherent round in 0.6 rather than piecemeal in 0.5.

Three things to do before merge:

### a) Demo must show the feature
Right now nothing exercises `handleMouseAction`, so the feature is invisible. Wire it into `scrollbar-demo` (or a list/table demo) so clicking the track pages, dragging the thumb scrolls, and the wheel works — with a `.tape` under `docs/video` if we want it on the site.

### b) Hook up horizontal scroll wheel (#399) and fix #428 in the same round
[#399](https://github.com/tamboui/tamboui/pull/399) (merged) added `SCROLL_LEFT`/`SCROLL_RIGHT`, but `ScrollbarAction` only has `SCROLL_UP`/`SCROLL_DOWN`, so a horizontal scrollbar can't be wheel-scrolled without the consumer remapping left→up. Add `SCROLL_LEFT`/`SCROLL_RIGHT` members and map them to `prev()`/`next()`. [#428](https://github.com/tamboui/tamboui/issues/428) (scrollbar-demo doesn't scroll horizontally) is almost certainly the same missing wiring — worth fixing together so the demo in (a) shows both axes.

### c) Simplify the `MouseEvent → ScrollbarAction` boilerplate
The current design keeps `Scrollbar` free of a `tamboui-tui` dependency (avoids a `tui → widgets → tui` cycle) by taking raw coords + a `ScrollbarAction` enum, which is the right call for the widget layer — but it forces every consumer to hand-write the same `switch (mouseEvent.kind())` mapping. Two-tier fix, no cycle needed:

- **Toolkit users (most people): zero boilerplate.** `tamboui-toolkit` already `api`-depends on **both** `tamboui-widgets` and `tamboui-tui`, and already has `ScrollbarElement`/`ScrollableElement` that handle mouse events internally. Do the `MouseEvent → ScrollbarAction` mapping inside those elements and call `handleMouseAction` for the user — so at the toolkit level scrolling "just works" with no mapping code at all.
- **Low-level `tamboui-tui` users: one-liner.** Add a small static mapper on the tui side (which can see both types), e.g. `ScrollbarMouse.actionFor(MouseEvent)` returning `Optional<ScrollbarAction>`, so callers write `ScrollbarMouse.actionFor(e).ifPresent(a -> scrollbar.handleMouseAction(e.x(), e.y(), a, area, state))` instead of a switch.

That keeps the clean no-cycle widget API while removing the copy-paste for everyone above it.

PR #406, opened 2026-08-02
fix(panama): dispatch SIGWINCH through JVM signal bridge

### LINKED ISSUE #405: Panama FFI upcall as SIGWINCH handler crashes GraalVM native-image
## Summary

Tamboui's `UnixTerminal.onResize()` installs a Panama FFI upcall as the SIGWINCH signal handler via `sigaction()`. This upcall runs Java code (acquires a `ReentrantLock`, writes to `resizePending`) from signal context, which crashes GraalVM SubstrateVM with:

\`\`\`
Fatal error: Must either be at a safepoint or in native mode
\`\`\`

The crash is easily reproducible by resizing the terminal window while a Tamboui TUI is running as a native image.

## Environment

- **Tamboui:** 0.4.0
- **GraalVM:** Oracle GraalVM 25.2.4+7.1
- **Substrate VM:** 25.0.4+7-LTS, serial gc, compressed references
- **native-image:** 25.0.4
- **OS:** Linux x86_64 (Fedora, kernel 7.1.5)
- **Terminal:** Ghostty

## Root Cause

The upcall installed in \`onResize()\` via \`LibC.createSignalHandler()\` + \`LibC.sigaction()\` is invoked by the kernel in signal context. Signal handlers can interrupt any thread at any point — including mid-Panama-FFI-transition where the thread is neither at a safepoint nor in native mode. The upcall then tries to transition back into Java, hitting SubstrateVM's safepoint assertion.

Even if the upcall body were trivial, signals can interrupt threads during Panama FFI downcalls elsewhere in the application, and the signal dispatch mechanism itself triggers the crash.

## Suggested Fix

Replace the \`sigaction()\`-based SIGWINCH handler with a polling approach:

1. Don't install a Panama upcall as a signal handler
2. Periodically check the terminal size via \`ioctl(TIOCGWINSZ)\` (e.g., every 200-300ms)
3. Set \`resizePending = true\` only when the size actually changes

Alternatively, use \`sun.misc.Signal.handle("WINCH", ...)\` which goes through the JVM's signal management rather than raw \`sigaction()\` — though this still has edge cases in SubstrateVM.

The polling approach is the safest option for native-image compatibility.

## Workaround

We work around this in [incus-spawn](https://github.com/Sanne/incus-spawn) by using reflection to pre-set the \`signalArena\` field to non-null before \`runner.run()\` calls \`onResize()\`, which causes the null-check to skip handler installation. We then set \`signal(SIGWINCH, SIG_IGN)\` and poll for size changes ourselves.

## Related

- Our issue: https://github.com/Sanne/incus-spawn/issues/212
- GraalVM upstream issue for the residual downcall crash: https://github.com/oracle/graal/issues/14146

maxandersen:
## Summary

- replace the raw Panama `sigaction()` upcall with a JVM-managed `SIGWINCH` bridge
- keep resize callbacks deferred to the terminal input loop
- add a native-image test that sends a real `SIGWINCH` and verifies handler restoration
- run the Panama native test on Linux and macOS CI

## Rationale

The existing handler lets the kernel enter Java through an FFM upcall while an arbitrary thread may be between Java and native states. Even though the callback only marks a resize as pending, entering Java and acquiring a `ReentrantLock` from signal context can trigger the SubstrateVM failure reported in #405.

The replacement follows the approach used by Aesh and JLine. It registers `sun.misc.Signal` reflectively, allowing the JVM to own the platform signal handler and dispatch the Java callback from a valid JVM context. The callback only sets a volatile flag; the input loop still invokes the application resize handler and reads the new size with `ioctl(TIOCGWINSZ)`.

This keeps resize handling event-driven. Polling would avoid signal APIs altogether, but it would add periodic native calls while idle and up to one polling interval of resize latency.

## Tradeoffs

`sun.misc.Signal` is an internal API, so the Panama module now declares `jdk.unsupported` and includes native-image reflection/proxy metadata. Signal handlers are also process-global. The bridge restores the previous handler when removed or closed, matching the existing lifecycle behavior.

The native regression test covers both JVM-managed delivery and restoration under GraalVM native-image. If a future JDK or GraalVM stops supporting this bridge, the native CI job should fail and polling remains a fallback.

## Testing

- `./gradlew -q test`
- `./gradlew -q build`
- `./gradlew -q javadoc`
- `./gradlew -q :tamboui-panama-backend:nativeTest`

Fixes #405

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -78,6 +78,9 @@ jobs:
       - uses: ./.github/actions/shared-build-setup
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Test Panama signal handling in native image
+        if: runner.os != 'Windows'
+        run: ./gradlew :tamboui-panama-backend:nativeTest
       - name: Build native executable
         run: ./gradlew demos:filemanager-demo:nativeCompile
       # - name: Run demo, send q after 2s (Unix) fails as no pty.
--- a/tamboui-panama-backend/build.gradle.kts
+++ b/tamboui-panama-backend/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     id("dev.tamboui.java-library")
+    id("org.graalvm.buildtools.native")
 }
 
 description = "Panama FFI backend for TamboUI TUI library"
@@ -12,6 +13,13 @@ tasks.withType<JavaCompile>().configureEach {
 
 tasks.withType<Test> {
     jvmArgs("--enable-native-access=ALL-UNNAMED")
+}
+
+graalvmNative {
+    binaries.named("test") {
+        buildArgs.addAll("--enable-native-access=ALL-UNNAMED", "-H:+SharedArenaSupport")
+    }
+    toolchainDetection.set(false)
 }
 
 dependencies {
--- a/tamboui-panama-backend/src/main/java/dev/tamboui/backend/panama/unix/SignalBridge.java
+++ b/tamboui-panama-backend/src/main/java/dev/tamboui/backend/panama/unix/SignalBridge.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.backend.panama.unix;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
+import dev.tamboui.terminal.BackendException;
+
+/**
+ * Bridges Unix signals through the JVM signal dispatcher.
+ */
+final class SignalBridge implements AutoCloseable {
+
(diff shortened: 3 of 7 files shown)


Sanne:
FWIW I couldn't test it yet as the use of these APIs requires additional module configurations and the exact errors are being swallowed by the service loading strategy in Tamboui.
I expect a Quarkus extension to be able to both make this smoother and more efficient, but you might want to think about that for the sake of non-quarkus users.

maxandersen:
@Sanne i think https://github.com/tamboui/tamboui/pull/391 or variant of it will help that case. 

Curious how you think quarkus could make this better as it would still need to be handled at runtime not buildtime?

Sanne:
Thanks @maxandersen , this seems effective. I'll be looking forward to a release containing this.

Sanne:
@maxandersen ? don't forget to merge this eventually :)

Sanne:
bump!

PR #410, opened 2026-08-07
feat: add synchronized output (Mode 2026) to prevent screen tearing

stalep:
Wrap Terminal.draw() with Mode 2026 BSU/ESU sequences so the terminal buffers all frame output and renders it atomically. This eliminates visible tearing on supported terminals (Kitty, Ghostty, WezTerm, Foot, Contour, iTerm2, Mintty) with no impact on unsupported ones which safely ignore the escape sequences.

Changes:
- Backend interface: add MODE_2026_BSU/ESU constants and beginSynchronizedUpdate()/endSynchronizedUpdate() default methods
- Terminal.draw(): wrap diff rendering + cursor handling in BSU/ESU with ESU in finally block to prevent stuck buffering state
- All 3 backends implement the methods (Aesh, JLine, Panama)
- SynchronizedOutputTest verifies correct call ordering

Also upgrade aesh-terminal from 3.14.1 to 3.16.4.

Replaces PR #282 (created against older codebase, now conflicting).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 jline = "3.25.1"
-aesh-terminal = "3.14.1"
+aesh-terminal = "3.16.4"
 junit = "5.14.3"
 assertj = "3.27.7"
 native-build-tools = "0.11.4"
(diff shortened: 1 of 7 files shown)


Sanne:
Could this relate to #409 ?

PR #412, opened 2026-08-08
fix: skip Mode 2027 query to eliminate 500ms startup penalty (#409)

stalep:
Enable Mode 2027 (grapheme cluster mode) unconditionally without querying via DECRQM first. Terminals that don't support Mode 2027 safely ignore the escape sequence — same as Mode 2026 (synchronized output).

The DECRQM query caused a 500ms timeout on:
- Terminals that don't respond (most terminals)
- AeshBackend where the response was consumed by the EventDecoder before reaching the inputQueue
- PanamaBackend when the terminal doesn't support Mode 2027

Also fix AeshBackend constructor race: move openNonBlocking() after stdinHandler setup to avoid losing input that arrives before the handler is registered.

### DIFF
--- a/tamboui-aesh-backend/src/main/java/dev/tamboui/backend/aesh/AeshBackend.java
+++ b/tamboui-aesh-backend/src/main/java/dev/tamboui/backend/aesh/AeshBackend.java
@@ -21,7 +21,6 @@ import org.aesh.terminal.utils.ANSI;
 import dev.tamboui.layout.Position;
 import dev.tamboui.layout.Size;
 import dev.tamboui.terminal.AbstractBackend;
-import dev.tamboui.terminal.Mode2027Status;
 import dev.tamboui.terminal.Mode2027Support;
 
 /**
@@ -81,7 +80,6 @@ public class AeshBackend extends AbstractBackend {
     private AeshBackend(Connection connection, boolean ownConnection) throws IOException {
         this.connection = Objects.requireNonNull(connection, "connection cannot be null");
         this.ownConnection = ownConnection;
-        this.connection.openNonBlocking();
         this.outputBuffer = new StringBuilder();
         this.inputQueue = new LinkedBlockingQueue<>();
         this.inAlternateScreen = false;
@@ -101,6 +99,9 @@ public class AeshBackend extends AbstractBackend {
                 resizeHandler.run();
             }
         });
+
+        // Start reader AFTER handlers are set up to avoid losing input (#409)
+        this.connection.openNonBlocking();
     }
 
     @Override
@@ -172,12 +173,11 @@ public class AeshBackend extends AbstractBackend {
     @Override
     public void enableRawMode() throws IOException {
         savedAttributes = connection.enterRawMode();
-        // Query and enable Mode 2027 (grapheme cluster mode) after entering raw mode
-        Mode2027Status status = Mode2027Support.query(this, 500);
-        if (status.isSupported()) {
-            Mode2027Support.enable(this);
-            mode2027Enabled = true;
-        }
+        // Enable Mode 2027 (grapheme cluster mode) unconditionally.
+        // Terminals that don't support it safely ignore the escape sequence.
+        // No DECRQM query — avoids 500ms timeout penalty (#409).
+        Mode2027Support.enable(this);
+        mode2027Enabled = true;
     }
 
     @Override
--- a/tamboui-jline3-backend/src/main/java/dev/tamboui/backend/jline3/JLineBackend.java
+++ b/tamboui-jline3-backend/src/main/java/dev/tamboui/backend/jline3/JLineBackend.java
@@ -18,7 +18,6 @@ import org.jline.utils.NonBlockingReader;
 import dev.tamboui.layout.Position;
 import dev.tamboui.layout.Size;
 import dev.tamboui.terminal.AbstractBackend;
-import dev.tamboui.terminal.Mode2027Status;
 import dev.tamboui.terminal.Mode2027Support;
 
 /**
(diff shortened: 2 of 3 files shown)


gnodet:
Great fix for the immediate pain — the 500ms penalty on non-responding terminals is a real problem. A couple of thoughts from how we handle this in JLine, in case it's useful for a future iteration.

**The unconditional-enable approach works**, but it means tamboui loses the ability to know whether Mode 2027 is actually active. That matters for width calculations: when Mode 2027 is on, a flag emoji like 🇫🇷 is 2 columns wide (one grapheme cluster); when it's off, the terminal treats it as two separate regional indicator codepoints (4 columns). If tamboui assumes Mode 2027 is active but the terminal silently ignored the escape, cursor positioning will be off by 2 columns for every flag/ZWJ emoji on screen.

**What JLine does instead — batch probe with a DA1 fence:**

Rather than probing Mode 2027 alone (and eating a timeout when the terminal doesn't answer), JLine sends a single batch query that probes *all* DEC private modes at once, terminated by a DA1 request (`CSI c`):

```
CSI ? u   CSI ? 2026 $ p   CSI ? 2027 $ p   CSI ? 2048 $ p   CSI c
```

DA1 is near-universally supported — every terminal that understands escape sequences responds to it. So it acts as a **fence**:

- DA1 response arrives **with** DECRPM answers before it → we know exactly which modes are supported
- DA1 response arrives **without** DECRPM → terminal doesn't support DECRQM; all modes marked "not supported" (but the terminal *does* speak escapes, so cursor-position fallbacks are safe)
- **No response at all** → dumb terminal; skip everything

The timeout covers the entire batch (200ms default, configurable), not 500ms per mode. One round-trip, all features detected.

**Three-state result enables smart fallbacks:**

JLine distinguishes `SUPPORTED` / `NOT_SUPPORTED` / `NO_RESPONSE`. When Mode 2027 comes back as `NOT_SUPPORTED` (meaning the terminal *did* respond to DA1, just not to DECRQM for 2027), JLine runs a cursor-position fallback: it writes test emoji (a flag and a ZWJ sequence), queries cursor position via DSR/CPR, and checks whether the cursor advanced 2 columns (grouped as one cluster) or 4 (ungrouped). This detects native grapheme cluster support on terminals that predate Mode 2027 — which is surprisingly common.

The `NO_RESPONSE` case is the key insight: if the terminal didn't answer DA1 at all, the cursor probe is skipped entirely because DSR/CPR would also block forever. That's what eliminates the timeout penalty.

**How this could look in tamboui:**

1. Combine the Mode 2027 DECRQM with a DA1 sentinel in a single write
2. Read with a ~200ms timeout — if DA1 comes back, parse whatever preceded it
3. If Mode 2027 is supported → enable it (as today)
4. If DA1 responded but Mode 2027 didn't → optionally run a cursor-position probe to detect native emoji grouping
5. If nothing responded → skip (no penalty)

The relevant JLine code is in [`AbstractTerminal.probeModes()`](https://github.com/jline/jline3/blob/master/terminal/src/main/java/org/jline/terminal/impl/AbstractTerminal.java) if you want to look at the implementation.

No urgency — the unconditional enable is a perfectly fine pragmatic fix. Just sharing in case tamboui wants accurate width computation down the road. 🙂

maxandersen:
Rebased onto current main — notably over #410 (Mode 2026), which touched the same backend files; clean rebase, all backend/core tests and spotless green locally.

Reviewed: the unconditional enable is safe for TamboUI today — `mode2027Enabled` only drives the symmetric disable on cleanup (equally ignorable by non-supporting terminals); no width computation consults it, so the accuracy concern is future-facing. The AeshBackend constructor race fix (reader started after handlers register) is a solid catch on its own.

@gnodet's DA1-fence batch probe is now captured as #437 for a future iteration, per his "no urgency" framing — `Mode2027Support.query`/`Mode2027Status` stay in core as its building blocks. From my side this is ready to merge.

maxandersen:
Measured the fix before/after under identical conditions: a real PTY with nothing answering DECRQM (via `script -q /dev/null …`), which is what the majority of terminals effectively do — timing taken around `enableRawMode()`, where the query lives:

| | `enableRawMode()` | on the wire |
|---|---|---|
| main (DECRQM query) | **505.8 ms** | `CSI ?2027$p`, then blocks the full timeout |
| this PR (unconditional) | **0.5 ms** | `CSI ?2027h` fire-and-forget (+ `?2027l` on cleanup) |

The full 500 ms penalty is gone, matching #409's report. Two notes from the measurement:

- The headless/pipe path never had the delay — JLine's dumb terminal returns EOF immediately, so the query "succeeded fast" there. The penalty only bites on a real TTY that stays silent, which is why it was invisible in CI and on terminals like Ghostty/Kitty that answer DECRQM quickly, and painful everywhere else (e.g. Apple Terminal).
- `mode2027Enabled` only drives the symmetric disable-on-cleanup — no width computation consults it — so the unconditional enable has no correctness downside today. The capability-aware future is tracked in #437.

Repro (`startup_delay.java`, run with `script -q /dev/null jbang startup_delay.java`):

```java
//DEPS dev.tamboui:tamboui-jline3-backend:0.5.0-SNAPSHOT

import dev.tamboui.terminal.Backend;
import dev.tamboui.terminal.BackendFactory;

public class startup_delay {
    public static void main(String[] args) throws Exception {
        long t0 = System.nanoTime();
        Backend backend = BackendFactory.create();
        long created = System.nanoTime();
        backend.enableRawMode();
        long rawMode = System.nanoTime();
        backend.disableRawMode();
        backend.close();

        System.err.printf("create backend : %6.1f ms%n", (created - t0) / 1_000_000.0);
        System.err.printf("enableRawMode  : %6.1f ms   <-- Mode 2027 query lives here%n",
                (rawMode - created) / 1_000_000.0);
    }
}
```

maxandersen:
fwiw @gnodet and @stalep its things like https://github.com/tamboui/tamboui/issues/437 and this - would be great if we could have a small console library both aesh and jline and then tamboui uses :)

PR #418, opened 2026-08-16
fix: apply recording to explicitly configured backends

ammachado:
## Problem

Recording is applied in exactly one place, `BackendFactory.create()`, which loads the `RecordingConfig` (installing the `System.out` capture and the shutdown hook that writes the cast) and wraps the backend in `RecordingBackend` (which loads the tape via `InteractionPlayer`).

But `TuiRunner.create()` only calls that factory when `TuiConfig` has no backend set:

```java
Backend backend = config.backend() != null ? config.backend() : BackendFactory.create();
```

So any application that builds its own backend and passes it to `TuiConfig.Builder#backend(Backend)` silently loses recording: no tape playback, no `.cast` file, and a clean exit that is indistinguishable from success.

Applications have good reasons to supply their own backend. Apache Camel's TUI does it because it ships both the JLine and Aesh backends (the latter for its browser mode), and ServiceLoader auto-discovery can pick the wrong one for a local session. That cost them a silently broken `--record` flag for several months, which is what prompted this change.

## Change

- Expose the wrapping as `BackendFactory.applyRecording(Backend)`, and reuse it inside `create()`.
- `TuiRunner.create()` now calls it for explicitly configured backends, fixing the bug at the source for every caller rather than requiring each one to remember.
- `applyRecording` is **idempotent**. An application that already wrapped its own backend (to work around this on an older version) will not end up with two `RecordingBackend`s. Stacking would give the outer wrapper its own `InteractionPlayer`, which would consume the tape while the inner recorder saw no input.
- It also checks `RecordingConfig.isEnabled()` before `load()`, since `load()` returns its cached config without re-reading the system properties. This makes the method reflect the current property state.

## Tests

New `BackendFactoryRecordingTest` covers the disabled path, the wrapped path (asserting the backend reports the *configured* cast size rather than the delegate's, which proves the config reached the wrapper), and idempotency.

The enabled assertions live in a single test method on purpose: `RecordingConfig` caches its config process-wide once loaded, so a second test loading different values would get the first test's config back and fail depending on execution order.

`:tamboui-core:test` and `:tamboui-tui:test` pass: 758 tests, forced rerun.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of @ammachado

### DIFF
--- a/tamboui-core/src/test/java/dev/tamboui/internal/record/BackendFactoryRecordingTest.java
+++ b/tamboui-core/src/test/java/dev/tamboui/internal/record/BackendFactoryRecordingTest.java
@@ -18,7 +18,7 @@ import dev.tamboui.terminal.TestBackend;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
- * Tests for {@link BackendFactory#applyRecording(Backend)}.
+ * Tests for {@link BackendFactory#recordIfEnabled(Backend)}.
  * <p>
  * Applications that build their own backend and pass it to {@code TuiConfig.Builder#backend(Backend)} bypass
  * {@code BackendFactory.create()}, which used to be the only place recording was applied. Such applications got a clean
@@ -52,7 +52,7 @@ class BackendFactoryRecordingTest {
     void returnsTheBackendUnchangedWhenRecordingIsDisabled() {
         Backend backend = new TestBackend(80, 24);
 
-        assertThat(BackendFactory.applyRecording(backend)).isSameAs(backend);
+        assertThat(BackendFactory.recordIfEnabled(backend)).isSameAs(backend);
     }
 
     @Test
@@ -61,7 +61,7 @@ class BackendFactoryRecordingTest {
         System.setProperty("tamboui.record.width", "120");
         System.setProperty("tamboui.record.height", "30");
 
-        Backend wrapped = BackendFactory.applyRecording(new TestBackend(80, 24));
+        Backend wrapped = BackendFactory.recordIfEnabled(new TestBackend(80, 24));
 
         // The recording backend reports the configured cast size rather than the delegate's size,
         // which proves the config reached the wrapper instead of merely something being returned.
@@ -71,6 +71,6 @@ class BackendFactoryRecordingTest {
         // A caller may wrap its own backend and then hand it to TuiRunner, which wraps again. Stacking
         // would give the outer wrapper an interaction player of its own, so the tape would be consumed
         // by the wrong layer and the inner recorder would never see any input.
-        assertThat(BackendFactory.applyRecording(wrapped)).isSameAs(wrapped);
+        assertThat(BackendFactory.recordIfEnabled(wrapped)).isSameAs(wrapped);
     }
 }
--- a/tamboui-core/src/test/java/dev/tamboui/internal/record/BackendFactoryRecordingTest.java
+++ b/tamboui-core/src/test/java/dev/tamboui/internal/record/BackendFactoryRecordingTest.java
@@ -2,7 +2,7 @@
  * Copyright TamboUI Contributors
  * SPDX-License-Identifier: MIT
  */
-package dev.tamboui.terminal;
+package dev.tamboui.internal.record;
 
 import java.nio.file.Path;
 
@@ -10,8 +10,10 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
-import dev.tamboui.internal.record.AnsiTerminalCapture;
 import dev.tamboui.layout.Size;
+import dev.tamboui.terminal.Backend;
+import dev.tamboui.terminal.BackendFactory;
+import dev.tamboui.terminal.TestBackend;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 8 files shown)


tamboui-core/src/test/java/dev/tamboui/internal/record/BackendFactoryRecordingTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @AfterEach
+    void tearDown() {
+        // load() installs a System.out capture; restore the real stream for the remaining tests
+        if (AnsiTerminalCapture.isInstalled()) {
+            AnsiTerminalCapture.uninstall();
+        }
+        System.clearProperty("tamboui.record");
+        System.clearProperty("tamboui.record.width");
+        System.clearProperty("tamboui.record.height");
```
davsclaus, line 46: Non-blocking: `tearDown` clears the system properties and uninstalls `AnsiTerminalCapture`, but `RecordingConfig.activeConfig` stays populated for the life of the JVM — `clearActive()` is package-private in `dev.tamboui.internal.record` and unreachable from this test's `dev.tamboui.terminal` package.

Today this is harmless: no other test loads recording, and the `isEnabled()` property guard stops a later `create()`/`applyRecording()` from wrapping once the property is cleared. The trap is for the *next* recording test sharing the JVM — `load()` would hand it this stale, temp-dir-backed config. The in-class comment already notes this within the class; just flagging that it also leaks across classes. Worth considering exposing a test-only reset (or moving these tests into `dev.tamboui.internal.record` so they can call `clearActive()`) if more recording tests get added.

davsclaus (approved):
## Review: apply recording to explicitly configured backends

Well-targeted, well-documented fix. The root-cause framing is correct — recording was only applied inside `BackendFactory.create()`, so any caller supplying its own backend via `TuiConfig.Builder#backend(Backend)` silently lost recording. Routing explicit backends through `applyRecording()` fixes it at the source. The idempotency guard, the `isEnabled()`-before-`load()` ordering, and the tests are all sound, and the Javadoc is genuinely helpful.

I traced the other entry points: `InlineTuiRunner` always builds its backend via `BackendFactory.create()` (already recording-wrapped), and `TuiRunner.create(TuiConfig)` is the only path accepting a user-supplied backend — so this change closes the gap completely. No convention violations (title/branch/license headers match repo style).

### Non-blocking observations

1. **Process-wide `RecordingConfig` cache is never reset by the test** (inline comment below). Harmless today, but a latent cross-class isolation hazard for the next recording test.
2. **Behavioral nuance in `create()` (intentional, worth confirming).** Previously `create()` wrapped whenever `load()` returned non-null — including a *cached* config after the property had been cleared. Now it wraps only when `isEnabled()` reports the property is currently set. Arguably more correct (the PR body explains the reasoning), but it is a subtle behavior change for the `create()` path too, not only the new explicit-backend path.
3. **Minor:** after `isEnabled()` returns true, `load()` can only return `null` in a racy clear-between-calls window, so the `if (recordingConfig != null)` guard is effectively always-true defensive code. Fine as-is.

### Not verified
I did not run the Gradle test tasks locally. The PR states 758 tests pass; the new test logic checks out on inspection.

tamboui-core/src/test/java/dev/tamboui/internal/record/BackendFactoryRecordingTest.java
ammachado: Good catch, fixed in 5f430f61 by taking the second option: the test moved to `dev.tamboui.internal.record` (alongside `InteractionPlayerTest`), so `tearDown` can call `clearActive()` directly. No production visibility was widened, and the temp-dir-backed config no longer outlives the test.

Side effect: with the cache reset per test, the "these have to share one method or execution order decides the outcome" constraint is gone. The two enabled assertions stay in one method now because they cover one scenario, not because they have to.

ammachado:
Thanks for the review. All four points addressed in 5f430f61.

**1. Process-wide `RecordingConfig` cache** — replied inline. Test moved to `dev.tamboui.internal.record` so `tearDown` can call `clearActive()`; no production visibility widened.

**2. Behavioral nuance in `create()`** — intentional, and you're right that it deserved to be stated rather than inferred. The `applyRecording` Javadoc now says so explicitly:

> Enablement is read from the `tamboui.record` system property on every call. This also changes `create()`, which previously wrapped whenever the config loaded: because the config is cached process-wide, `create()` kept wrapping after the property had been cleared. Both paths now stop wrapping as soon as the property goes away.

**3. Effectively always-true guard** — kept the null check (the race window is real, and `new RecordingBackend(backend, null)` would NPE later rather than at the call site), but collapsed it into the return so it no longer reads as a branch with a meaningful else:

```java
RecordingConfig recordingConfig = RecordingConfig.load();
return recordingConfig != null ? new RecordingBackend(backend, recordingConfig) : backend;
```

**4. Test verification** — ran locally on macOS (aarch64, Gradle 9.2.1): `./gradlew test` is green. **3009 tests, 0 failures, 0 errors, 5 skipped.** The 5 skipped are pre-existing platform gates, unrelated to this PR:

- `WindowsTerminalPeekTest` (3) — Windows-only
- `CharsetDetectionTest` (2) — `utf8IsDefaultForModernSystems`, `charsetIsDetectedFromEnvironment`

The 758 figure in the PR description was a single module's count, not the whole build; 3009 is the full suite.

maxandersen:
Rebased onto main (over #422/#391, resolving the `BackendFactory.create`/`TuiRunner.create` conflicts) preserving the original commits, and added three commits on top — flagging them since they extend the PR's scope, shout if you'd rather see them split out:

- `86da0d27` — recording can now be enabled with only `tamboui.record.config=script.tape`; the cast output is derived from the tape name (`script.cast`). Naming the input tape stays the explicit act; no implicit file lookup is introduced.
- `aff6ade9` — new `recording.adoc`: the `tamboui.record*` properties, supported tape commands, and the headless drive-and-record recipe were previously undocumented.
- `070361ce` — `applyRecording` renamed to `recordIfEnabled` (conditional + idempotent, the name now says so; never shipped, so free).

Verified end-to-end with a headless jbang probe: an app passing its own backend via `TuiConfig.backend(...)` is driven by a VHS tape and produces a valid `.cast` — the exact Camel `--record` case. Multi-instance recording semantics documented as future work in #433.

maxandersen:
thanks for this!  this is awesome for agents (and humans :)

PR #419, opened 2026-08-24
Add text wrap support to TextArea

ammachado:
## Summary
- Add word/character wrapping to `TextArea`/`TextAreaState`, reusing the existing `Overflow` vocabulary (`CLIP`/`WRAP_WORD`/`WRAP_CHARACTER`) that `Paragraph` already exposes via `text-overflow`.
- `TextAreaState` gains a display-row model (`computeDisplayRows`) and wrap-aware overloads of `ensureCursorVisible`/`moveCursorUp`/`moveCursorDown`, so Up/Down move by visual row once wrapping is enabled. `CLIP`-mode behavior (the default) is unchanged.
- `TextAreaElement` exposes `overflow(Overflow)` plus the `wrapWord()`/`wrapCharacter()`/`clip()` shortcuts (and honors the `text-overflow` CSS property), wired through to the widget, to Up/Down key handling, and to its preferred size.

Related to [CAMEL-24395](https://issues.apache.org/jira/browse/CAMEL-24395) (camel-tui unifying its view/edit rendering onto a single `TextArea`-based path) and the gaps captured in `docs/textarea-readonly-and-styling-gaps.md`.

## Test plan
- [x] `./gradlew :tamboui-widgets:test` — 839 tests pass
- [x] `./gradlew :tamboui-toolkit:test` — 913 tests pass
- [x] `./gradlew :tamboui-widgets:check :tamboui-toolkit:check` — spotless clean
- [x] `./gradlew :docs:compileJava` — snippets still compile against the unchanged no-arg/2-arg overloads

### DIFF
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
@@ -324,6 +324,20 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
         // Add border height
         int borderHeight = (title != null || borderType != null) ? 2 : 0;
         int height = lineCount + borderHeight;
+
+        // Wrapped lines take more than one row each; ask the widget, which knows the width its
+        // border and line-number gutter leave for text.
+        Overflow effectiveOverflow = resolveOverflow(context);
+        if (availableWidth > 0 && state != null
+                && (effectiveOverflow == Overflow.WRAP_WORD || effectiveOverflow == Overflow.WRAP_CHARACTER)) {
+            boolean isFocused = context != null && elementId != null && context.isFocused(elementId);
+            height = TextArea.builder()
+                    .showLineNumbers(showLineNumbers)
+                    .overflow(effectiveOverflow)
+                    .block(buildBlock(context, isFocused))
+                    .build()
+                    .preferredHeight(availableWidth, state);
+        }
         return Size.of(width, height);
     }
 
@@ -467,27 +481,8 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
             .placeholderStyle(effectivePlaceholderStyle)
             .showLineNumbers(showLineNumbers)
             .lineNumberStyle(effectiveLineNumberStyle)
-            .overflow(effectiveOverflow);
-
-        Color effectiveBorderColor = isFocused && focusedBorderColor != null
-                ? focusedBorderColor
-                : borderColor;
-
-        if (title != null || borderType != null || effectiveBorderColor != null) {
-            Block.Builder blockBuilder = Block.builder()
-                    .borders(Borders.ALL)
-                    .styleResolver(styleResolver(context));
-            if (title != null) {
-                blockBuilder.title(Title.from(title));
-            }
-            if (borderType != null) {
-                blockBuilder.borderType(borderType);
-            }
-            if (effectiveBorderColor != null) {
-                blockBuilder.borderColor(effectiveBorderColor);
-            }
-            builder.block(blockBuilder.build());
-        }
+            .overflow(effectiveOverflow)
+            .block(buildBlock(context, isFocused));
 
         TextArea widget = builder.build();
 
@@ -498,6 +493,30 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
         } else {
             frame.renderStatefulWidget(widget, area, state);
         }
+    }
+
+    /** Returns the border block to draw around the text, or null when there is none. */
+    private Block buildBlock(RenderContext context, boolean isFocused) {
+        Color effectiveBorderColor = isFocused && focusedBorderColor != null
+                ? focusedBorderColor
+                : borderColor;
+
+        if (title == null && borderType == null && effectiveBorderColor == null) {
+            return null;
+        }
+        Block.Builder blockBuilder = Block.builder()
+                .borders(Borders.ALL)
+                .styleResolver(context != null ? styleResolver(context) : null);
+        if (title != null) {
+            blockBuilder.title(Title.from(title));
+        }
+        if (borderType != null) {
+            blockBuilder.borderType(borderType);
+        }
+        if (effectiveBorderColor != null) {
+            blockBuilder.borderColor(effectiveBorderColor);
+        }
+        return blockBuilder.build();
     }
 
     /**
--- a/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
+++ b/tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
@@ -10,6 +10,7 @@ import java.util.Map;
 
 import dev.tamboui.layout.Rect;
 import dev.tamboui.style.Color;
+import dev.tamboui.style.Overflow;
 import dev.tamboui.style.Style;
 import dev.tamboui.terminal.Frame;
 import dev.tamboui.toolkit.element.RenderContext;
@@ -72,6 +73,8 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
     private boolean showCursor = true;
     private boolean showLineNumbers = false;
     private Style lineNumberStyle;
+    private Overflow wrap;
+    private int lastVisibleWidth;
     private TextChangeListener changeListener;
 
     /** Creates a new text area element with a default state. */
@@ -198,6 +201,48 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
     }
 
     /**
+     * Sets the wrap mode.
+     * <p>
+     * {@link Overflow#CLIP} (the default) preserves horizontal-scroll behavior. {@code
+     * WRAP_WORD}/{@code WRAP_CHARACTER} wrap long lines across multiple screen rows instead,
+     * and Up/Down move the cursor by visual row rather than logical line.
+     *
+     * @param overflow the wrap mode
+     * @return this builder
+     */
+    public TextAreaElement overflow(Overflow overflow) {
+        this.wrap = overflow;
+        return this;
+    }
+
+    /**
+     * Disables wrapping (horizontal scroll instead). This is the default.
+     *
+     * @return this builder
+     */
+    public TextAreaElement clip() {
+        return overflow(Overflow.CLIP);
+    }
+
+    /**
+     * Wraps long lines at word boundaries.
+     *
+     * @return this builder
+     */
+    public TextAreaElement wrapWord() {
+        return overflow(Overflow.WRAP_WORD);
+    }
+
+    /**
+     * Wraps long lines at character boundaries.
+     *
+     * @return this builder
+     */
+    public TextAreaElement wrapCharacter() {
+        return overflow(Overflow.WRAP_CHARACTER);
+    }
+
+    /**
      * Sets the title for the border.
      *
      * @param title the border title
@@ -314,7 +359,7 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
         if (!focused) {
             return EventResult.UNHANDLED;
         }
-        boolean handled = handleTextAreaKey(state, event);
+        boolean handled = handleTextAreaKey(state, event, lastVisibleWidth, wrap);
         if (handled && changeListener != null) {
             changeListener.onTextChange(state.text());
         }
@@ -324,7 +369,7 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
     /**
      * Handles common key events for text area input.
      */
-    private static boolean handleTextAreaKey(TextAreaState state, KeyEvent event) {
+    private static boolean handleTextAreaKey(TextAreaState state, KeyEvent event, int visibleWidth, Overflow wrap) {
         switch (event.code()) {
             case BACKSPACE:
                 state.deleteBackward();
@@ -339,10 +384,10 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
                 state.moveCursorRight();
                 return true;
             case UP:
-                state.moveCursorUp();
+                state.moveCursorUp(visibleWidth, wrap);
                 return true;
             case DOWN:
-                state.moveCursorDown();
+                state.moveCursorDown(visibleWidth, wrap);
                 return true;
             case HOME:
                 state.moveCursorToLineStart();
@@ -391,12 +436,14 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
             .placeholder(placeholder)
             .placeholderStyle(effectivePlaceholderStyle)
             .showLineNumbers(showLineNumbers)
-            .lineNumberStyle(effectiveLineNumberStyle);
+            .lineNumberStyle(effectiveLineNumberStyle)
+            .wrap(wrap);
 
         Color effectiveBorderColor = isFocused && focusedBorderColor != null
                 ? focusedBorderColor
                 : borderColor;
 
+        Block block = null;
         if (title != null || borderType != null || effectiveBorderColor != null) {
             Block.Builder blockBuilder = Block.builder()
                     .borders(Borders.ALL)
@@ -410,8 +457,11 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
             if (effectiveBorderColor != null) {
                 blockBuilder.borderColor(effectiveBorderColor);
             }
-            builder.block(blockBuilder.build());
+            block = blockBuilder.build();
+            builder.block(block);
         }
+
+        lastVisibleWidth = computeVisibleWidth(area, block);
 
         TextArea widget = builder.build();
 
@@ -420,6 +470,26 @@ public final class TextAreaElement extends StyledElement<TextAreaElement> {
         } else {
             frame.renderStatefulWidget(widget, area, state);
         }
+    }
+
+    /**
+     * Computes the text-content width available inside {@code area}, after the block border
+     * (if any) and the line-number gutter (if enabled) — mirroring {@code TextArea}'s own
+     * layout math, since that width isn't otherwise exposed back to the element.
+     */
+    private int computeVisibleWidth(Rect area, Block block) {
+        Rect inputArea = block != null ? block.inner(area) : area;
+        if (inputArea.isEmpty()) {
+            return 0;
+        }
+        int gutterWidth = 0;
+        if (showLineNumbers) {
+            int lineDigits = String.valueOf(state.lineCount()).length();
+            gutterWidth = Math.max(2, lineDigits) + 2;
+        }
+        return gutterWidth > 0 && inputArea.width() > gutterWidth
+            ? inputArea.width() - gutterWidth
+            : inputArea.width();
     }
 
     /**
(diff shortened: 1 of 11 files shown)


davsclaus (approved):

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextAreaState.java
```diff
@@ -245,15 +309,221 @@ public void moveCursorToEnd() {
...
+            String line = lines.get(row).toString();
+            if (noWrap) {
+                result.add(new DisplayRow(row, 0, line.length()));
+                continue;
+            }
+            if (line.isEmpty()) {
+                result.add(new DisplayRow(row, 0, 0));
+                continue;
+            }
+            List<int[]> segments = wrap == Overflow.WRAP_WORD
```
davsclaus: `Overflow` has six values, but anything that isn't `CLIP`/`null`/`WRAP_WORD` falls through to `wrapLineByCharacter` here. So `ELLIPSIS`, `ELLIPSIS_START`, and `ELLIPSIS_MIDDLE` — all reachable via `TextArea.Builder.wrap(Overflow)`, `TextAreaElement.overflow(Overflow)`, or the `text-overflow` CSS property — would silently render as character wrapping.

Could we make this predictable? Options: treat unsupported values as `CLIP`, reject them in the builder, or document that only `CLIP`/`WRAP_WORD`/`WRAP_CHARACTER` are supported. This is my most substantive comment; everything else is minor.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
```diff
@@ -422,6 +472,26 @@ protected void renderContent(Frame frame, Rect area, RenderContext context) {
         }
     }
 
+    /**
+     * Computes the text-content width available inside {@code area}, after the block border
+     * (if any) and the line-number gutter (if enabled) — mirroring {@code TextArea}'s own
+     * layout math, since that width isn't otherwise exposed back to the element.
+     */
+    private int computeVisibleWidth(Rect area, Block block) {
```
davsclaus: `computeVisibleWidth` re-derives the content width by copying the widget's gutter formula (`Math.max(2, lineDigits) + 2` + border inset). I verified it matches `TextArea`'s math exactly today, and I appreciate the honest comment flagging it — but if the widget's gutter layout ever changes, Up/Down (which rely on the element's `lastVisibleWidth`) would wrap differently than what's actually rendered, with no test to catch the drift.

Worth considering a single source of truth — e.g. stash the computed content width on the state during render, or a small shared helper the widget and element both call. Non-blocking.

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextArea.java
```diff
@@ -374,6 +437,21 @@ public Builder lineNumberStyle(Style style) {
...
+         * Sets the wrap mode.
+         * <p>
+         * {@link Overflow#CLIP} (the default) preserves today's horizontal-scroll behavior.
+         * {@code WRAP_WORD}/{@code WRAP_CHARACTER} wrap long lines across multiple screen rows
+         * instead of scrolling horizontally.
+         *
+         * @param wrap the wrap mode
+         * @return this builder
+         */
+        public Builder wrap(Overflow wrap) {
```
davsclaus: Minor API-consistency question: `Paragraph.Builder` exposes `overflow(Overflow)`, whereas here the widget builder is `wrap(Overflow)` while `TextAreaElement` exposes both `overflow(...)` and the `wrapWord()`/`wrapCharacter()`/`clip()` shortcuts. Was the divergence intentional ("wrap" reads better for an editor), or would aligning the widget builder to `overflow()` be preferable for a consistent public API?

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextArea.java
```diff
@@ -245,22 +295,34 @@ public void renderWithCursor(Rect area, Buffer buffer, TextAreaState state, Fram
...
+                int from = Math.min(scrollCol, cursorLine.length());
+                int to = Math.min(cursorCol, cursorLine.length());
+                relativeCol = CharWidth.of(cursorLine.substring(from, to));
+            }
         } else {
-            String cursorLine = state.getLine(cursorRow);
-            int from = Math.min(scrollCol, cursorLine.length());
-            int to = Math.min(cursorCol, cursorLine.length());
-            relativeCol = CharWidth.of(cursorLine.substring(from, to));
+            List<TextAreaState.DisplayRow> rows = state.computeDisplayRows(textArea.width(), wrap);
```
davsclaus: Very minor UX edge (just confirming it's acceptable): with `WRAP_WORD`, the break whitespace belongs to no `DisplayRow` — e.g. `[0,7)` then `[8,13)` for "one two three" at width 7. A cursor sitting exactly on that consumed space (col 7, reachable via Left/Right) maps via `findDisplayRowIndex` to the first row with `relativeCol == width`, which fails the `relativeCol < textArea.width()` visibility check below, so the cursor briefly disappears. Low severity — noting in case you'd rather snap it to the start of the next visual row.

davsclaus (commented):
Thanks for this — it's a clean, well-tested addition. I built the branch in a worktree and the full `tamboui-widgets` + `tamboui-toolkit` suites pass with spotless clean. Nice touches: `CLIP` stays the default and its path is untouched, backward compatibility is preserved via the delegating overloads (`moveCursorUp()` / `moveCursorDown()` / `ensureCursorVisible(int,int)`), the wrap vocabulary reuses the existing `Overflow` enum (consistent with `Paragraph`), and the tests validate real behavior (CJK width, empty lines, cross-logical-line cursor movement) rather than just exercising code. Java 8 source compatibility holds too.

No blocking issues — a few suggestions and questions, left inline plus two here that aren't tied to a specific changed line:

**Docs (per `CONTRIBUTING.md`):** the guide asks to update `docs/src/docs/asciidoc/` when public API/behavior changes. The `[[textarea]]` section in `widgets.adoc` doesn't yet mention wrap modes, the new `wrap`/`overflow` builder methods, or `text-overflow` CSS support — a short note mirroring the Paragraph "Wrap modes" block would keep docs in sync. (Aside: the pre-existing Paragraph doc at `widgets.adoc:163-167` references a `Wrap` enum that no longer exists — stale and not caused by this PR, but adjacent if you feel like fixing it.)

**Performance (question, fine to defer):** in wrap mode a single focused frame calls `computeDisplayRows` ~3 times (`ensureCursorVisible` + `renderWrapped` + `renderWithCursor`), plus once per Up/Down keystroke, each re-wrapping the whole document. Given the stated driver (camel-tui unifying view/edit onto `TextArea`, potentially large content), is caching the wrapped layout — invalidated on text/width change — worth a follow-up?

_Scope: this is a conventions + correctness pass and does not replace CodeRabbit, Sourcery, or SonarCloud._

_Reviewed by Claude Code on behalf of @davsclaus. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextAreaState.java
ammachado: Good catch, fixed in 89e85b9d.

I went with "treat unsupported values as `CLIP`" rather than rejecting them in the builder: `text-overflow` is a shared CSS property, so a stylesheet-wide `text-overflow: ellipsis` would throw on a `TextArea` that never opted in. Falling back is the quieter, more predictable behavior.

The decision now lives in one place, `TextAreaState.isWrapping(Overflow)`, which `computeDisplayRows`, `moveCursorUp`/`moveCursorDown`, `ensureCursorVisible`, and `TextArea.Builder.resolveOverflow()` all route through, so there is no longer a fall-through path. The rationale (a text area lets the caret reach text that truncation would hide) is documented on `isWrapping`, on both builders, and in `widgets.adoc`.

Covered by `computeDisplayRowsTruncatingModesDoNotWrap`, `moveCursorDownTruncatingModeMovesByLogicalLine`, `renderTreatsEllipsisAsClip`, and `truncatingOverflowFallsBackToClip`.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
ammachado: Agreed, and fixed in 89e85b9d by your first suggestion: `TextArea.render()` now records the content width it actually used on the state, and `TextAreaElement` reads it back via `TextAreaState.lastRenderedWidth()`. `computeVisibleWidth` is gone, along with the element's `lastVisibleWidth` field.

While in there I also removed the second copy of the same formula inside `TextArea` itself: `render()` and `renderWithCursor()` each had their own inline gutter math, and both now call `gutterWidth(state)` / `textAreaRect(inputArea, gutterWidth)`.

For the drift you were worried about, `TextAreaTest.renderPublishesContentWidth` renders with both a border and a gutter, then re-wraps at `lastRenderedWidth()` and asserts the resulting segments match the text actually drawn in the buffer, so a change to the gutter layout that doesn't reach the published width fails the test. `TextAreaElementTest.arrowKeysUseTheRenderedContentWidth` covers the Up/Down side.

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextArea.java
ammachado: Not intentional, just inconsistency on my part. Aligned to `overflow(Overflow)` in 89e85b9d.

`wrap()` was introduced in this PR and has never shipped, so I renamed it outright rather than adding an alias, and the internal fields and parameters are named `overflow` throughout to match. `TextAreaElement` keeps the `wrapWord()` / `wrapCharacter()` / `clip()` shortcuts on top of `overflow(...)`, which reads well at the DSL layer without forking the vocabulary.

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextArea.java
ammachado: Snapped it, in 89e85b9d, but only for display.

There is a new `TextAreaState.findCursorDisplayRowIndex`, used by `renderWithCursor` and `ensureCursorVisible`: a cursor at an offset past the current row's `endCol` is drawn at, and scrolled to, the start of the following visual row of the same logical line.

I deliberately left `findDisplayRowIndex` itself alone. Snapping there too breaks the Up/Down round-trip: `moveCursorUp` clamps into `[0,7]` and lands on offset 7 as the end of the `"one two"` row, and `moveCursorDown` then has to resolve that offset back to row 0 so it can return to `"three"`. If movement snapped, `Down` would resolve it to row 1 and find nothing below. So the two questions are answered separately: `findDisplayRowIndex` is "which row owns this offset", `findCursorDisplayRowIndex` is "where is the caret painted". Both are documented with that split.

`renderWithCursorOnConsumedWordWrapBreak` and `ensureCursorVisibleSnapsCursorOnConsumedWrapBreak` cover it.

ammachado:
Thanks for the thorough pass. All six points are addressed in 89e85b9d, including the two from the review body that weren't tied to a line. `./gradlew build` (tests + javadoc + spotless) is green.

**Docs.** The `[[textarea]]` section in `widgets.adoc` now has a "Wrap modes" block covering `Overflow.CLIP` / `WRAP_CHARACTER` / `WRAP_WORD`, the builder methods on both layers, the fact that Up/Down move by visual row while wrapping, the line-number-on-first-row-only behavior, and the `text-overflow` CSS property. There's a `NOTE` spelling out the truncating-mode fallback, since that's exactly the surprise your inline comment flagged.

I also took the adjacent fix: the Paragraph block referenced a `Wrap` enum that no longer exists, and now lists the real `Overflow` values against `.overflow(...)`.

**Performance.** Not deferred, it turned out to be small. `TextAreaState` caches the result of `computeDisplayRows`, keyed on a text-version counter (bumped by `insert`, `deleteBackward`, `deleteForward`, `clear`, `setText`) plus the width and mode. A focused frame now wraps the document once instead of three times, and Up/Down keystrokes reuse the layout the render just built rather than re-wrapping per keypress. The cached list is handed out unmodifiable.

Given the camel-tui driver you asked about, that seemed worth doing now rather than leaving a per-keystroke full-document re-wrap in the path for large content. `computeDisplayRowsInvalidatedByTextChange` and `computeDisplayRowsInvalidatedByWidthOrMode` pin the invalidation.

maxandersen:
Rebased onto current main (over the merged #388, whose `CharWidth` changes this PR's wrap logic consumes — clean rebase, all widgets/toolkit tests and javadoc green) and added one commit: `textarea-demo` now shows the three overflow modes side by side (main editor = clip with an overlong line, Notes = editable word wrap, read-only display = character wrap), so the visual-row Up/Down behavior is easy to exercise.

One related finding, deliberately **not** in this PR: TextArea's per-code-point width accounting misses VS16 emoji presentation sequences (drift vs. the post-#388 buffer). That's pre-existing, filed as #434.

All four earlier review points are confirmed addressed in `89e85b9d` (now `02e9ac99`); the existing approval predates that commit, so this needs one fresh maintainer pass.

maxandersen:
One more commit (TDD): `TextAreaElement` now honors the `text-overflow` CSS property, closing an inconsistency — `TextElement` already resolved it, so a stylesheet-wide `text-overflow` rule styled paragraphs but silently skipped text areas. Resolution order is programmatic `overflow(...)` > CSS > `CLIP` (mirrors `TextElement.resolveOverflow`); Up/Down key handling uses the value resolved at the last render, same pattern as `lastRenderedWidth`. Ellipsis values still fall back to clip via the existing widget-level guard. Three new tests incl. precedence and key-handling paths; javadoc CSS section updated.

tamboui-toolkit/src/main/java/dev/tamboui/toolkit/elements/TextAreaElement.java
```diff
@@ -197,6 +199,50 @@ public TextAreaElement lineNumberStyle(Style style) {
...
+    }
+
+    /**
+     * Wraps long lines at word boundaries.
+     *
+     * @return this builder
+     */
+    public TextAreaElement wrapWord() {
+        return overflow(Overflow.WRAP_WORD);
+    }
```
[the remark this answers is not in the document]
ammachado: Fixed in 39b93784.

`TextAreaElement.preferredSize` is now wrap-aware. With `WRAP_WORD`/`WRAP_CHARACTER` (set programmatically or resolved from the `text-overflow` CSS property), the height comes from a new `TextArea.preferredHeight(width, state)`. It counts `computeDisplayRows` at the width the border and line-number gutter leave for text, using the same helpers `render()` uses, so sizing can't drift from what is drawn. `Column` sizes an unconstrained child by its preferred height, so wrapped rows are no longer clipped.

I deliberately did not add a `Constraint.min` default the way `TextElement` does. Sizing a text area to its content keeps wrap mode consistent with the `CLIP` default, and an explicit constraint still wins.

Covered by `columnSizesWrappedTextArea`, `preferredSizeCountsWrappedRows`, `preferredSizeUsesCssResolvedOverflow`, and `TextAreaTest.preferredHeightFitsWrappedText`.

_Claude Code on behalf of Adriano Machado (@ammachado)_

_This was generated by an AI agent and may contain inaccuracies. Please verify before relying on it._

tamboui-widgets/src/main/java/dev/tamboui/widgets/input/TextAreaState.java
```diff
@@ -245,15 +344,262 @@ public void moveCursorToEnd() {
...
+    private static List<int[]> wrapLineByCharacter(String line, int maxWidth) {
+        List<int[]> segments = new ArrayList<>();
+        int start = 0;
+        int width = 0;
+        int i = 0;
+        int len = line.length();
+        while (i < len) {
+            int codePoint = line.codePointAt(i);
+            int codePointWidth = CharWidth.of(codePoint);
+            int charCount = Character.charCount(codePoint);
```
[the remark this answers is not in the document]
ammachado: Fixed in 39b93784, including lines 482 and 490 and the matching word-wrap comment on line 494 from the review summary.

Both wrap modes now step by grapheme cluster, using `GraphemeClusters.clusterEnd` (which Left/Right and Backspace already use), and measure each cluster as a string with `CharWidth.of(String)`. A ZWJ sequence or flag is never split across rows, and it is as wide during wrapping as it is when rendered.

Covered by `computeDisplayRowsWrapCharacterKeepsZwjSequenceWhole` (`a👨‍🦲b` at width 3 wraps to `a👨‍🦲` / `b`) and `computeDisplayRowsWrapWordKeepsZwjSequenceWhole`.

The VS16 width drift filed as #434 is separate and not addressed here.

_Claude Code on behalf of Adriano Machado (@ammachado)_

_This was generated by an AI agent and may contain inaccuracies. Please verify before relying on it._

ammachado:
Addressed the seven comments that Copilot's review listed as suppressed (they have no threads, so I'm answering them here). The fixes are in 39b93784, on top of @maxandersen's edcc0c27. `./gradlew build` (tests, javadoc, spotless) is green.

1. **`text-overflow` never reaching the widget (`TextAreaElement.java:443`).** Already fixed by edcc0c27: the element resolves programmatic > CSS > `CLIP`, and key handling uses the value resolved at the last render. 39b93784 reuses that resolution for sizing.
2. **The promised `TextAreaElement.wrap()`.** The PR description was stale. The widget builder method was renamed to `overflow(Overflow)` in the previous review round, and the element never had `wrap()`. I updated the description.
3. **Stray `[len, len]` tail in `wrapLineByCharacter` (`TextAreaState.java:482`).** The tail is now appended only when text is left over. Note that your exact example (a single 2-column glyph in a 1-column area) still gets a second, empty row, now on purpose: it is the caret row from item 6, so the caret after the glyph has a cell. There is no third row (`computeDisplayRowsOverWideGlyph`).
4. **ZWJ sequences in word wrap (`TextAreaState.java:494`).** Fixed with the grapheme-cluster change; see the inline thread on line 462.
5. **Up/Down losing the visual column (`TextAreaState.java:322`).** Up/Down now start from the row the caret is drawn on and keep its screen column (width- and grapheme-aware). `CLIP` is unchanged. Your example now lands at offset 9, `t|hree`. This changes a design choice from the previous review round: movement and drawing now share one lookup, `findCursorDisplayRowIndex`. A caret at a word-wrap break is drawn right after the row's text when the row has room, and moves to the start of the next row only when the row is full. That keeps Up from a long row into a short one at `ab|`, not `a|b`.
6. **Caret disappearing after trailing whitespace (`TextAreaState.java:510`).** A line whose last row fills the width, or whose trailing whitespace a word break consumed, now ends with an empty display row. `one two ` at width 7 draws the caret at the start of row 2, where typing continues. The trade-off is that a line exactly as wide as the area shows one blank row after it. The alternative, adding the row only for the line under the caret, would shift the layout as the caret moves between lines.
7. **`scrollDown` clamping to logical lines (`TextAreaState.java:577`).** Added a `scrollDown(amount, visibleRows, visibleCols, overflow)` overload that clamps to the display-row count when wrapping and delegates to the existing overload otherwise, like the other wrap-aware overloads. The 2-argument overload is unchanged.

_Claude Code on behalf of Adriano Machado (@ammachado)_

_This was generated by an AI agent and may contain inaccuracies. Please verify before relying on it._

maxandersen:
Thanks @ammachado — verified `39b93784` locally (widgets/toolkit/css tests green): the grapheme-cluster wrapping measuring each cluster via `CharWidth.of(String)` also means VS16 emoji presentation sequences now measure 2-wide in the wrap paths, which covers the wrap-side of #434 (noted there; the CLIP-mode horizontal scroll accounting remains).

Pushed one more demo commit on top: `css-demo` now has a bottom text-area pane whose overflow mode comes entirely from the theme (`dark.tcss` = `wrap-word`, `light.tcss` = `clip`) — pressing `t` re-wraps/clips the same element with zero code, demonstrating the CSS path end to end. Demo story: `textarea-demo` = programmatic API, `css-demo` = stylesheet-driven.

PR #421, opened 2026-08-25
Fix #420: expose Table viewport height API for accurate scrollbar sizing

### LINKED ISSUE #420: Expose Table viewport height API for accurate scrollbar sizing
## Problem

Consumers of the `Table` widget that render their own scrollbar (e.g. `renderTableWithScrollbar` in [pilot](https://github.com/maveniverse/pilot)) need to know how many data rows are actually visible in the viewport. Currently there's no public API for this, forcing consumers to approximate with `area.height() - 1` — which is wrong when the table has a block border but no header, or a header but no block, etc.

The actual viewport height depends on:
- Block borders (top + bottom) — via `Block.verticalChrome()`
- Header row height — via `header.totalHeight()`
- Footer row height — via `footer.totalHeight()`

All of these are already computed internally in `Table.render()` but not exposed.

## Proposed API

Add a method on `Table` that computes the number of visible data rows for a given area:

```java
public int viewportHeight(Rect area) {
    int height = area.height();
    if (block != null) {
        height -= block.verticalChrome();
    }
    if (header != null) {
        height -= header.totalHeight();
    }
    if (footer != null) {
        height -= footer.totalHeight();
    }
    return Math.max(0, height);
}
```

Also consider making `Block.verticalChrome()` and `Block.horizontalChrome()` public (they already exist as package-private at [Block.java:206](https://github.com/tamboui/tamboui/blob/main/tamboui-widgets/src/main/java/dev/tamboui/widgets/block/Block.java#L206)).

## Context

See [pilot PR #93 discussion](https://github.com/maveniverse/pilot/pull/93#discussion_r3195341777) for the background on why this is needed.

gnodet:
## Summary

- Add `Table.viewportHeight(Rect)` method that computes the number of data rows visible in a given area, accounting for block chrome (borders, titles, padding), header row, and footer row heights
- This enables consumers like [pilot's `renderTableWithScrollbar`](https://github.com/maveniverse/pilot/pull/93#discussion_r3195341777) to size scrollbars accurately instead of approximating with `area.height() - 1`
- Note: `Block.verticalChrome()` and `Block.horizontalChrome()` are already public — no visibility change was needed

## Test plan

- [x] Added 7 new test cases covering all viewport height scenarios:
  - No chrome (returns full area height)
  - Block chrome subtracted
  - Header height subtracted
  - Footer height subtracted
  - All three combined (block + header + footer)
  - Never-negative guarantee (tiny area)
  - Header with bottom margin
- [x] All 831 existing tests pass
- [x] Spotless formatting check passes

Closes #420

### DIFF
--- a/tamboui-widgets/demos/table-demo/src/main/java/dev/tamboui/demo/TableDemo.java
+++ b/tamboui-widgets/demos/table-demo/src/main/java/dev/tamboui/demo/TableDemo.java
@@ -29,6 +29,8 @@ import dev.tamboui.widgets.block.BorderType;
 import dev.tamboui.widgets.block.Borders;
 import dev.tamboui.widgets.block.Title;
 import dev.tamboui.widgets.paragraph.Paragraph;
+import dev.tamboui.widgets.scrollbar.Scrollbar;
+import dev.tamboui.widgets.scrollbar.ScrollbarState;
 import dev.tamboui.widgets.table.Cell;
 import dev.tamboui.widgets.table.Row;
 import dev.tamboui.widgets.table.Table;
@@ -56,6 +58,7 @@ public class TableDemo {
 
     private boolean running = true;
     private final TableState tableState = new TableState();
+    private final ScrollbarState scrollState = new ScrollbarState(DATA.size());
 
     private TableDemo() {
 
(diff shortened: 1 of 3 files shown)


gnodet (commented):
Reviewed the implementation — looks correct:

- `viewportHeight(Rect)` mirrors what `Table.render()` computes internally for `visibleHeight`
- `block.verticalChrome()` matches `block.inner(area).height()` subtraction (borders + titles + padding) ✓
- `header.totalHeight()` and `footer.totalHeight()` account for row height + margin ✓
- `Math.max(0, ...)` prevents negative results for tiny areas ✓
- Tests cover all relevant combinations (no chrome, block only, header only, footer only, all three, negative, header margin)

This will let pilot replace the `area.height() - 1` approximation in `renderTableWithScrollbar` with `table.viewportHeight(area)`.

davsclaus (approved):

maxandersen:
Pushed a small refactor on top: `render()` duplicated the viewport math (`tableArea.height()` minus header/footer) in the scroll-to-selected block. It now calls `viewportHeight(area)` instead, so the calculation has a single owner and the scrollbar-facing API can never drift from what render actually uses.

Safe because render returns early when `tableArea.isEmpty()`, so `block.inner(area)` didn't clamp — meaning `tableArea.height() == area.height() - block.verticalChrome()`, exactly what `viewportHeight` computes. All widget tests (incl. the 7 new ones) pass.

maxandersen:
Also pushed a demo commit: `table-demo` now renders a vertical scrollbar overlaid on the table's right border, sized via `table.viewportHeight(area)` and synced from `tableState.offset()` after render — so the new API has an in-repo consumer demonstrating the intended usage (same pattern as pilot's scrollbar sizing). The scrollbar only appears when rows overflow the viewport (e.g. an 80x24 terminal).

PR #422, opened 2026-08-25
fix(core): treat comma-separated backend spec as preference order with fallback

gnodet:
## Summary

- Fix `BackendFactory.resolveProviders()` to skip unavailable providers instead of throwing immediately
- When `tamboui.backend` is set to `"panama,jline3"` but the panama backend can't load (e.g. its classes are compiled for Java 22 but the runtime is Java 21), the method now falls through to `jline3` instead of throwing `BackendException`
- Only throws if **none** of the specified providers were found — the error message now shows all requested names and the actually available providers

## Problem

`SafeServiceLoader` correctly catches `UnsupportedClassVersionError` (via `LinkageError`) when a backend JAR targets a newer Java version, but `resolveProviders()` then fails because it uses `orElseThrow()` on each entry individually. The comma-separated list was intended as a preference order with fallback (as documented in the Javadoc: *"tries each in order until one succeeds"*), but the implementation treated every entry as mandatory.

## Fix

Replace `orElseThrow()` with `ifPresent(resolved::add)` so unavailable providers are silently skipped, matching the documented behavior. Throw only when the final resolved list is empty.

## Test plan

- [x] Added `BackendFactoryTest` with 3 cases: fallback when first provider missing, order preserved when all available, throws when none available
- [ ] Build and run tests with Java 25+ (build requires Java 25+, not available in CI environment)
- [ ] Verify pilot CLI starts on Java 21 with `tamboui.backend=panama,jline3`

### DIFF
--- a/tamboui-core/src/test/java/dev/tamboui/terminal/BackendFactoryTest.java
+++ b/tamboui-core/src/test/java/dev/tamboui/terminal/BackendFactoryTest.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright TamboUI Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package dev.tamboui.terminal;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Tests for {@link BackendFactory}'s provider resolution logic.
+ * <p>
+ * Uses reflection to test the private {@code resolveProviders} method directly,
+ * verifying that a comma-separated provider specification acts as a preference
+ * order with fallback rather than a mandatory list.
+ */
+class BackendFactoryTest {
+
+    @Test
+    @DisplayName("resolveProviders skips unavailable provider and falls back to next")
+    void resolveProviders_skipsUnavailableProvider() throws Exception {
+        var method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
+        method.setAccessible(true);
+
+        // Only "beta" is available; "alpha" is not (simulating a provider that
+        // failed to load, e.g. because it targets a newer Java version)
+        List<BackendProvider> available = new ArrayList<>();
+        available.add(new StubBackendProvider("beta"));
+
+        @SuppressWarnings("unchecked")
+        List<BackendProvider> resolved = (List<BackendProvider>) method.invoke(null, "alpha,beta", available);
+
+        assertThat(resolved).hasSize(1);
+        assertThat(resolved.get(0).name()).isEqualTo("beta");
+    }
+
+    @Test
+    @DisplayName("resolveProviders preserves order when all providers are available")
+    void resolveProviders_preservesOrder() throws Exception {
+        var method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
+        method.setAccessible(true);
+
+        List<BackendProvider> available = new ArrayList<>();
+        available.add(new StubBackendProvider("beta"));
+        available.add(new StubBackendProvider("alpha"));
+
+        @SuppressWarnings("unchecked")
+        List<BackendProvider> resolved = (List<BackendProvider>) method.invoke(null, "alpha,beta", available);
+
+        assertThat(resolved).hasSize(2);
+        assertThat(resolved.get(0).name()).isEqualTo("alpha");
+        assertThat(resolved.get(1).name()).isEqualTo("beta");
+    }
+
+    @Test
+    @DisplayName("resolveProviders throws when no specified provider is available")
+    void resolveProviders_throwsWhenNoneAvailable() throws Exception {
+        var method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
+        method.setAccessible(true);
+
+        List<BackendProvider> available = new ArrayList<>();
+        available.add(new StubBackendProvider("gamma"));
+
+        assertThatThrownBy(() -> method.invoke(null, "alpha,beta", available))
+                .hasCauseInstanceOf(BackendException.class)
+                .cause()
+                .hasMessageContaining("alpha,beta");
+    }
+
+    /**
+     * Minimal BackendProvider stub for testing provider resolution.
+     */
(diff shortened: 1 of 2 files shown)


tamboui-core/src/test/java/dev/tamboui/terminal/BackendFactoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+ * Uses reflection to test the private {@code resolveProviders} method directly,
+ * verifying that a comma-separated provider specification acts as a preference
+ * order with fallback rather than a mandatory list.
+ */
+class BackendFactoryTest {
+
+    @Test
+    @DisplayName("resolveProviders skips unavailable provider and falls back to next")
+    void resolveProviders_skipsUnavailableProvider() throws Exception {
+        var method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
```
davsclaus: `var` is a Java 10+ feature; this module compiles with `--release 8` and `-Werror`, so this fails CI's `Full Gradle Build` job with "warnings found and -Werror specified". Please use the explicit type and add `import java.lang.reflect.Method;` at the top of the file.

```suggestion
        Method method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
```

tamboui-core/src/test/java/dev/tamboui/terminal/BackendFactoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+        List<BackendProvider> resolved = (List<BackendProvider>) method.invoke(null, "alpha,beta", available);
+
+        assertThat(resolved).hasSize(1);
+        assertThat(resolved.get(0).name()).isEqualTo("beta");
+    }
+
+    @Test
+    @DisplayName("resolveProviders preserves order when all providers are available")
+    void resolveProviders_preservesOrder() throws Exception {
+        var method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
```
davsclaus: Same issue as line 29 — `var` isn't valid under `--release 8`.

```suggestion
        Method method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
```

tamboui-core/src/test/java/dev/tamboui/terminal/BackendFactoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+        assertThat(resolved).hasSize(2);
+        assertThat(resolved.get(0).name()).isEqualTo("alpha");
+        assertThat(resolved.get(1).name()).isEqualTo("beta");
+    }
+
+    @Test
+    @DisplayName("resolveProviders throws when no specified provider is available")
+    void resolveProviders_throwsWhenNoneAvailable() throws Exception {
+        var method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
```
davsclaus: Same issue as line 29 — `var` isn't valid under `--release 8`.

```suggestion
        Method method = BackendFactory.class.getDeclaredMethod("resolveProviders", String.class, List.class);
```

davsclaus (changes requested):
Thanks for tracking this down — the root-cause analysis (SafeServiceLoader correctly skipping the unavailable panama provider, but `resolveProviders()` treating every entry in the list as mandatory) is spot on, and the fix itself matches the documented "tries each in order until one succeeds" contract in `resolveProviders()`'s Javadoc. Reusing the existing `formatAvailableProviders()` helper for the improved error message is a nice touch too.

One blocking issue and one non-blocking suggestion:

**Blocking — CI is currently red.** The `Full Gradle Build (excluding distributions)` check fails: `tamboui-core` compiles with `sourceCompatibility`/`targetCompatibility` = Java 8 and `-Werror` (`buildSrc/src/main/kotlin/dev.tamboui.java-base.gradle.kts`). `var` is a Java 10+ feature, so `javac --release 8` emits a restricted-type-name warning on each use in `BackendFactoryTest.java`, and `-Werror` turns that into a compile failure. See inline comments on the three occurrences.

**Non-blocking — test approach.** This introduces the first use of reflection (`setAccessible(true)`) to invoke a private method in this codebase's tests, as far as I could find. It works, but it's brittle (a rename/signature change breaks the test silently at runtime instead of compile time) and diverges from the existing convention of testing through the public API. Worth considering making `resolveProviders` package-private for testability instead, though I won't block on this since it's a style preference rather than a correctness issue.

This review only covers project conventions/build correctness — it doesn't replace a full static-analysis pass or a functional review of the panama/jline3 fallback behavior itself (which the author's own test plan notes as not yet verified on Java 21).

davsclaus (approved):

maxandersen:
Pushed one more commit (TDD: failing test first). The class JavaDoc documents provider selection by fully qualified class name, but `resolveProviders` only matched `BackendProvider#name()` — the failing test's error message even listed the exact FQCN it refused to match as "available". Since this PR already rewrites that filter, fixed it here: the filter now matches `name()` **or** the provider's FQCN. Also replaced a fully-qualified `java.util.ArrayList` with an import per project conventions.

This resolves the FQCN doc-mismatch concern raised on #391 at its source, so #391's rebase only needs to keep the behavior.

maxandersen:
Executed the remaining test-plan items with jbang probes against locally published 0.5.0-SNAPSHOT artifacts from this branch (all three backends on the classpath):

| Scenario | JVM | Result |
|---|---|---|
| `tamboui.backend=panama,jline3` | 21 | ✅ falls back to **jline3** (the pilot/Camel case — was `BackendException` before this PR) |
| `tamboui.backend=panama,jline3` | 25 | ✅ **panama** tried first (falls to jline3 only without a TTY) |
| `tamboui.backend=panama` | 21 | ✅ clear error: requested spec + available providers |
| `tamboui.backend=<FQCN of JLineBackendProvider>` | 21 | ✅ works (fixed in follow-up commit) |
| `tamboui.backend=bogus,alsobogus` | 21/25 | ✅ error names full spec + available providers (panama correctly listed only on 25) |
| no property / `TAMBOUI_BACKEND` env | 21 | ✅ auto-discovery + env var both work |

Two observations for follow-ups (not this PR):
1. When a requested provider is present-but-incompatible (panama on Java 21), the error still says "add a backend dependency" — misleading, since the jar is there. That's #391's failure-reporting territory; noted there.
2. Auto-discovery order with no property is raw classpath order (no JDK/suitability ranking) — separate design question.

From my side this is ready to merge; CI green including the follow-up commits.

maxandersen (approved):

PR #425, opened 2026-09-07
[markdown] enable to wire a syntax highlighter to markdown view

rmannibucau:


maxandersen:
hmm - i would love to have syntax highlighting in markdown but that shouldn't be limited to markdown.

we have #173  and others that enables syntax highlighting for other components and maybe time is to define a syntax highligting interface which can be replaced/reused in these scenarios?

maxandersen:
to be clear - the PR seems to have all the right parts; just saying the syntax highlighting part shouldn't be tied/hidden inside markdown package.

rmannibucau:
Agree, I'd just add that even if highlightjs design with regex is far to bé perfect it is insanely simple so worth having by default for extensibility and keep core light IMHO (explaining my choices)

maxandersen:
@rmannibucau yes, I'm fine having that lightweight hilitejs inspired default regex syntax highlighter included.

Not having that was actually one of the reasons I got stuck figuring out how to best integrate the other highlight options.>

rmannibucau:
think widget bundle is the right location, core is ok from dependency standpoint but looks too low level for this design, will update the PR

maxandersen:
This looks like a good direction, but a few issues — I rebased the PR on main and wired up filemanager-demo's file preview to actually have a demo for it. In doing so (with an LLM helping track it down) I spotted that the js and java grammars (python/sql too) can trigger `StackOverflowError` when an unclosed `"` or escape-heavy run is present:

```java
RegexSyntaxHighlighter.defaults().highlight(
    "\"" + repeat("a", 20000), "java", Style.EMPTY, SyntaxTheme.DEFAULTS); // -> StackOverflowError
```

For most content that probably won't happen, but when streaming markdown (LLM output) an unclosed string is the *normal* mid-stream state — and highlighting is on by default — so I think we should fix it, or at least avoid the error. I added tests that show the limit and will fail on CI until addressed.

Likely fix direction: rewrite the string/comment rules in non-backtracking form (possessive / char-class-first), and/or add a per-line length cap (very long lines fall back to `PLAIN`) — the cap would also avoid the multi-second highlight of a single 20k-char line I measured along the way.

wdyt?

rmannibucau:
gave a try if you can have a look

maxandersen:
Had a look — the fix works, confirmed at both layers: I isolated the possessive quantifier against a 200k-char unclosed string (no line cap in play) and it match-fails cleanly in ~10ms where the greedy version overflows, and your sub-cap 9k tests pin exactly that. The `maxLineLength` guard also takes care of the multi-second single-line pathologies I'd measured earlier. Nice — this is more minimal than the mode-based rewrite I had been sketching, so I dropped that.

Pushed three commits on top:

1. **Per-language smoke tests for all built-in grammars** (previously only java had correctness coverage). They immediately caught two real bugs, fixed in the second commit: SQL keywords were case-sensitive (`SELECT` didn't highlight) and CSS numbers didn't match at all when carrying units (`4px` — the shared `NUMBERS` pattern's trailing `\b` can't sit between a digit and a letter).
2. **Semantic class tagging**: each span's style now carries its `TokenType` as a typed style extension plus `Tags.of("syntax-keyword")` etc. — allows to use CSS for theme/style. It also sets this PR up as the base layer for #130/#173 to become optional `SyntaxHighlighter` providers later — for those, we should probably document `SyntaxTheme` as advisory since such engines bring their own themes.
3. **A `properties` grammar** (with a new line-anchored rule variant so URI-heavy values like `kafka:topic?x=y` don't get key-colored) — that completes coverage of the languages camel-tui's hand-rolled highlighter supports, setting up its replacement (@davsclaus and @gnodet - wdyt?)

One forward-looking note: the `SyntaxHighlighter` interface may later grow an *optional* line-oriented capability (`highlightLine(line, entryState) -> (Line, exitState)`) so caching editors can re-highlight only changed lines — additive, no breaking change to what's here.

With that I think this is in good shape — @rmannibucau anything you want to adjust, or shall we get it a final review pass?

rmannibucau:
@maxandersen im happy with current state so if you are let's go :)

davsclaus:
Sounds good max.

maxandersen:
gave gpt 5 sol a go and it found a few things - parsers stuck on empty input and recreated grammars on every screen render (wont work well ;) - working through fixing those.

maxandersen:
did a review pass on this and added a set of follow-up commits:

- added opt-in syntax highlighting for `TextArea` / `TextAreaElement`, including the shorter `.highlighter("java")` and `.highlighter("java", theme)` APIs
- made syntax token styling follow the usual explicit theme > element CSS > generic CSS > defaults precedence for both markdown and text areas
- stopped recompiling all built-in regex grammars during markdown measure/render; the immutable grammars are now shared and lazily initialized for native-image friendliness
- protected the tokenizer from zero-width custom regex rules and reject empty literal/multiline delimiters before they can stall the render thread
- fixed HTML token semantics so tag names, attributes and text are distinct
- fenced info strings such as `java linenums` now pass `java` to the highlighter while keeping the full block title
- function calls now tokenize as `FUNCTION("run")` + `PUNCTUATION("(")`, rather than `FUNCTION("run(")`
- validate that `maxLineLength` is positive
- cleaned up the public docs/javadocs and documented CSS syntax selectors and actual style precedence

all of this was done test-first with regression coverage for each issue. locally `./gradlew -q test build javadoc` passes.

maxandersen:
fwiw I think there are some optimizations needed for full blown usage - but this is a good start and its optin for textarea so lets see what shows up

PR #452, opened 2026-09-15
perf: 2.1x faster frame pipeline (four independent optimizations)

franz1981:
Four independent optimizations of the frame pipeline, one per commit. Each commit passes `./gradlew build javadoc` on its own.

On the sparkline dashboard through the real Panama (FFM) backend at 384×108 (a 4K terminal), a frame cost **770.9 µs** and allocated **981 KB**: `renderWidget` 56.5%, `Buffer.diff` 31.5%, `draw` 3.6%.

**What changed**

1. `Style` mirrors its two `EnumSet<Modifier>` fields as `int` bitmasks, and `Buffer` shares one `String` per glyph in U+2500–U+28FF. Comparing equal-but-distinct styles no longer walks `RegularEnumSet`, `String.equals` exits on identity, and `setString` stops building a `char[]`+`String`+`byte[]` per non-ASCII code point — that chain alone was 55% of all allocation.
2. `DiffResult` stored per-cell `(x, y, Cell)` in three parallel arrays, but the only consumer read all three in lockstep. It now stores `(start, length)` pairs for runs of adjacent changed cells on a row plus a reference to the rendered buffer, and `draw` positions the cursor once per run.
3. `PanamaBackend` now overrides `writeRaw(CharSequence)` (the default called `toString()` on the cursor `StringBuilder` once per run) and `ByteArrayBuilder.appendUtf8` encodes UTF-8 in place instead of `String.getBytes` per glyph.
4. `Buffer.setStyle` returns early when the patch is empty. Widgets fill their background this way every frame — `Block` over its whole area — and with an empty style the pass is provably a no-op, yet paid a bounds check, index, patch, equals and reference store per cell. Six of the eight widgets that call it default to `Style.EMPTY`.

**Result**

| | before | after |
|---|---|---|
| frame, real FFM backend, 384×108 | 770.9 µs (1298 FPS) | **359.8 µs (2780 FPS)** — 2.14× |
| `SparklineDemoBenchmark` 160×50 | 4239 ± 80 ops/s | 7191 ± 198 ops/s |
| allocation / frame, 160×50 | 422,459 ± 4424 B | 188,285 ± 206 B — −55% |
| `BackendDrawBenchmark` full redraw | 41.1 ± 1.7 µs | 23.7 ± 0.2 µs |

`Buffer.diff` falls to 19.9% of the frame and `Cell.equals` to 14.5%. What remains is structural: `renderWidget` now owns 66%, and one `Cell` object per cell written is 92% of the remaining allocation.

**How it was measured** — existing benchmarks only: [`SparklineProfiler`](https://github.com/tamboui/tamboui/blob/main/tamboui-benchmarks/src/jmh/java/dev/tamboui/benchmarks/SparklineProfiler.java) real-terminal mode with `-Dtamboui.backend=panama` on a pty sized 384×108 (5 × 10 s runs, spread <1%), [`SparklineDemoBenchmark`](https://github.com/tamboui/tamboui/blob/main/tamboui-benchmarks/src/jmh/java/dev/tamboui/benchmarks/SparklineDemoBenchmark.java) (3 forks × 8 iterations, `-prof gc`) and [`BackendDrawBenchmark`](https://github.com/tamboui/tamboui/blob/main/tamboui-benchmarks/src/jmh/java/dev/tamboui/benchmarks/BackendDrawBenchmark.java), plus async-profiler for the phase split. Ryzen 9 7950X, boost off, `taskset -c 0-3`, JDK 25, `-Xms1g -Xmx1g` (chosen by measurement: 64m cost 3% throughput and 5× the GC CPU). GC logs checked every run: all collections young, none mixed or full, Old regions flat.

One caveat: [`FullPipelineBenchmark`](https://github.com/tamboui/tamboui/blob/main/tamboui-benchmarks/src/jmh/java/dev/tamboui/benchmarks/FullPipelineBenchmark.java) at 320×100/50% improves from ~154 µs to somewhere between 81 and 112 µs depending on the invocation, with only ±4 µs within a run. The gain is real but I can't pin it down, and I don't know why that benchmark is unstable between forks.

**API change** — `DiffResult.add(int, int, Cell)` is removed and `capacity()` counts runs. No caller remained in the repo. `getX`/`getY`/`getCell` still work but are now `O(runCount)`; backends should iterate `runCount()`/`runStart()`/`runLength()`/`cellAt()`. Worth a release note, as the build has no binary-compatibility check.

**Correctness** — `AbstractBackendDrawTest` asserts the run-based draw is byte-identical to the old per-cell algorithm over 15 frames (wide CJK, ZWJ emoji, wide↔narrow rewrites, hyperlinks, area change, zero-width area); `ByteArrayBuilderTest` checks the encoder against `String.getBytes(UTF_8)`. Both were verified to fail against mutated implementations.

Not included: `JLine3Backend` and `AeshBackend` need the same `writeRaw(CharSequence)` override, but I had no way to measure them.


franz1981:
Added a stress test with the Doom demo from #216 — a much heavier workload than a dashboard, and as Max noted there it already helped surface perf glitches.

Setup: 384×108, real Panama/FFM backend (verified by class loading — `JLineBackend` never loads), `--render block --protocol half` so the render mode can't vary, 3000 frames along a fixed deterministic camera path, `-Xms1g -Xmx1g`, JVM pinned to 4 cores with the pty reader kept off them. 5 runs per side, median (range):

| | before | after |
|---|---|---|
| FPS | 175.1 (167.9–176.7) | **349.8 (321.0–381.3)** — 2.0× |
| frame p50 | 5.41 ms | **2.75 ms** |
| frame p99 | 10.11 ms | **6.51 ms** |
| allocation / frame | 19.0 MB | **7.9 MB** — −58% |
| young GCs, same 3000 frames | 94 | **40** |
| peak RSS | 742 MB | 739 MB — unchanged |

Frame latency improves at every percentile (HdrHistogram over every frame), so there is no tail regression.

Where the 11 MB/frame went — async-profiler alloc, cross-checked against the GC logs within 2%:

- `Objects.hash(...)` in `Style.computeHashCode`: varargs `Object[]` **3469 → 4 KB/frame**, and the `EnumSet` iterator it walks **5382 → 24 KB/frame**. Doom builds a `Style` per shaded cell, so that alone was ~8.8 MB/frame of hashing garbage.
- `EnumSet.copyOf` in `Style`'s constructor: 5343 → 2780 KB/frame — still the next lever.
- `Color$Rgb` unchanged at 2052 KB/frame (Doom's own shading, untouched).

RSS is unchanged, and worth saying explicitly: G1 commits to absorb the allocation *rate*, so allocating less per frame does not shrink RSS on its own.

tamboui-core/src/main/java/dev/tamboui/style/Style.java
```diff
@@ -26,6 +26,9 @@ public final class Style {
     private final EnumSet<Modifier> subModifiers;
```
maxandersen, line 26: do we need to still keep these around?

maxandersen:
These are all clean and fairly isolated so lets put them in!

I think we should consider just holding onto a bitmask rather than carry around enumsets - we can always recreate them if needed for nicer/simpler apis.

